PR #2, opened 2022-02-17
YALB-257: Add local development scripts

dens0ne:
### Description of work
- Provides `package.json` file with scripts that now live in `scripts/node` to help development process

### Functional testing steps:
- [x] `git checkout` this branch
- [x] confirm you can `npm run build` and it will get your local env started (watch out for error when you have no DB to be imported)

## Notes
Theme related docs and scripts are not accurate/functional. We can refactor those later when we have this narrowed down.

### DIFF
--- a/.github/BUILD_TOOLS.md
+++ b/.github/BUILD_TOOLS.md
@@ -1,0 +1,36 @@
+# Build tools
+
+**Important**
+You may need to authorize your machine with your token if you are not able to pull assets.
+
+```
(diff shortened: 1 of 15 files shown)


nJim (approved):
Work is approved. The commands that need to work now are functional. There are a few that are stubbed out for the future. We should consider if the 'reference' database thing is right for this project.

There are a few things to change in the Readme file. I can take this on as I have some documentation tickets on my plate. I would rather see this work merged, so we can move forward. Tweaks can come later.

PR #15, opened 2022-04-26
YALB-521: Wire Pop-out Image

ModulesUnraveled:
## Related to the following PRs:
- https://github.com/yalesites-org/component-library-twig/pull/80
- https://github.com/yalesites-org/atomic/pull/7

### Description of work
- Creates a`theming.services.yml` file that disables basically every cache so that theming is easier (must be opted in via `settings.local.php` - so it won't affect developers not actively working on the theme)
- Places the `alert_block` ... somehow that got missed before?
- Creates a "Full Width" responsive image style and applies it to the default view mode for the image media type
- Creates a "Image - Float" responsive image style and applies it to the "Image - Float" view mode for the image media type
- Configures the "Pop-out Image" to render its image using the "Image - Float" view mode

### Functional testing steps:
- [x] Create a page, and add a Pop-out Image component
- [x] Upload and image, and fill in some text. Save the page
- [x] Verify the image uses the following "sizes" attribute: `(min-width:992px) 395px, (min-width:838px) 790px, 96vw`

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/block.block.alert_block.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/block.block.alert_block.yml
@@ -1,0 +1,20 @@
+uuid: ebeb6fd8-d226-41aa-a5cd-f6c9c1f4fbf7
+langcode: en
+status: true
+dependencies:
+  module:
+    - ys_alert
+  theme:
+    - atomic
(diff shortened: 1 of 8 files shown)


nJim (approved):
Got up and running on my local. Popup image component looks great and follows the specified breakpoints. Ship it.

PR #16, opened 2022-04-27
YALB-294: Create CTA Banner paragraph

ModulesUnraveled:
### Relates to the following PR:
- https://github.com/yalesites-org/atomic/pull/8

### Description of work
- Creates the "CTA Banner" paragraph type
- Creates the "Hero" tab and field on the standard page content type

### Functional testing steps:
- [ ] Log in to the site https://pr-16-yalesites-platform.pantheonsite.io/user/login
- [ ] [Create a page](https://pr-16-yalesites-platform.pantheonsite.io/node/add). Fill in the title and add a "pop-out image" component.
- [ ] Save the page, and verify:
	- [ ] The page title is shown at the top
	- [ ] The pop-out image component displays appropriately
- [ ] Edit the page, and navigate to the "Hero" tab.
- [ ] Fill in the Banner fields and save the page
- [ ] Verify the page title you saw a moment ago is gone, and the banner is shown instead (with the page title as the banner heading)


nJim (approved):
Tested on my local and all works well. Thank you so much for running with these site building tasks. Ship it. (pending your note about a name change)

I still have some potential concerns about the approach, mainly that we are creating specific rules in the data model to match the business rules around the design. WordPress has something called a 'feature image' and this item can be used in wildly different ways from theme to theme. Similarly we will need all future themes to consider this hero. Nothing inherently bad, but I want to talk about it in a future iteration so I'm documenting my thoughts here.

ahitson:
The link on the banner needs top padding when stacked.
This is one of the components that would benefit from a name change.

ahitson:
For pop out image, the image needs a caption field

ModulesUnraveled:
@ahitson The text will have the standard paragraph spacing on the bottom (between it and the link) so once we get the text formatting stuff configured, it'll look the same as it does in the component library.

I did add the pop-out-image caption field, so please re-review that, and verify it's there, and displays as expected.

ahitson:
[Discussed removing the extra padding on the caption text.]

miketullo95 (approved):
both the pop-out image and the CTA banner look great!

PR #21, opened 2022-04-30
YALB-377:  Adds linting

mickey4k:
### Description of work
- Adds linting packages.
- Adds documentation for running linting.
- Runs linting to correct existing files.

### Functional testing steps:
- [ ] Follow documentation to install and run linting - https://github.com/yalesites-org/yalesites-project/blob/YALB-377--adds-linting/docs/CONTRIBUTING.md#linting (copied and pasted below)

#### Linting
- [ ] To run linting, you'll need to create a Github Personal Access token to be able to pull down the Yale package.
- [ ] Visit the [Github instructions for creating a Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token).
- [ ] Give it "Write packages" permissions. This will also check the repo permissions.
- [ ] Export the token from your ~/.bash_profile or ~/.bashrc file by adding: 
    `export YALESITES_BUILD_TOKEN=[token]`
- [ ] Source the file with `source ~/.bash_profile` or `source ~/.bashrc`.

To run linting, you'll need to install packages in the project root and in web/core:
- [ ] cd into web/core with cd web/core and run `yarn`
- [ ] cd back to the root with cd ../../ and run `npm install`

Available npm commands for testing. All commands below can be run with `npm run [command]`. For example, `npm run lint:js`.
- [ ] prettier - runs prettier code beautifier tool on docs, .github, custom modules, and custom themes directories.
- [ ] lint:js - uses eslint to validate the JS in custom modules and themes.
- [ ] lint:style - uses eslint to validate the CSS in custom modules and themes.
- [ ] lint - runs both lint:js and lint:styles.
- [ ] test - runs prettier, lint:js and lint:styles.

## Notes:
- Static tests will fail as ys_mail needs some revision for dependency injection. But you can see that they're being run.


.npmrc
```diff
@@ -0,0 +1,2 @@
+@yalesites-org:registry=https://npm.pkg.github.com
+//npm.pkg.github.com/:_authToken=${YALESITES_BUILD_TOKEN}
```
nJim, line 2: I don't think this line is needed. While you need a token to update/build packages, I believe public packages are open to everyone. While your instructions work, getting an authToken add unnecessary complexity to contributing to this project. I think this line can be removed.

composer.json
```diff
@@ -1,117 +1,111 @@
...
+  "config": {
+    "preferred-install": "dist",
+    "sort-packages": true,
+    "platform": {
+      "php": "7.4"
     },
-    "scripts": {
-        "unit-test": "echo 'No unit test step defined.'",
-        "lint": "find web\/modules\/custom web\/themes\/custom -name '*.php' -exec php -l {} \\;",
-        "code-sniff": [
```
nJim, line 110: I'm fine with removing this from the composer.json. Getting everything into consistant set of scripts will feel great. We need to also make changes to `.ci/test/static/run` to call the new scripts.

nJim:
@mickey4k Sorry for the premature feedback. I see now that the title of the PR is WIP. I will hold tight on my review.

.npmrc
ModulesUnraveled: I think it is required. GitHub packages (even public ones) need this in order to be installed. So, while the repo is public, in order to install the package version, you have to have an auth Token... It's annoying, but that's what I found

ModulesUnraveled: For reference, here's what I've found: https://github.community/t/download-from-github-package-registry-without-authentication/14407/2

docs/CONTRIBUTING.md
```diff
@@ -104,3 +105,29 @@ Development standards will be rigorously enforced on this project. Drupal mainta
 - [Twig coding standards](https://www.drupal.org/docs/develop/coding-standards/twig-coding-standards)
 - [JavaScript coding stand](https://www.drupal.org/docs/develop/standards/javascript)
 - [CSS Architecture](https://www.drupal.org/docs/develop/standards/css/css-architecture-for-drupal-8)
+
+### Linting
+
+To run the linting, first you'll need to gain access to the [yalesites-org/eslint-config-and-other-formatting](https://github.com/yalesites-org/eslint-config-and-other-formatting) repo if you don't already have access.
```
ModulesUnraveled, line 111: This isn't actually necessary. It's a public repo, and it's released as a public package. So the only requirement is that they have a personal access token saved. That's a requirement of GitHub for installing packages, both private and public.

docs/CONTRIBUTING.md
```diff
@@ -104,3 +105,29 @@ Development standards will be rigorously enforced on this project. Drupal mainta
...
+
+- Visit the [Github instructions for creating a Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token).
+- Give it "Write packages" permissions. This will also check the repo permissions.
+- Export the token from your ~/.bash_profile or ~/.bashrc file by adding:
+  `export YALESITES_BUILD_TOKEN=[token]`
+- Source the file with `source ~/.bash_profile` or `source ~/.bashrc`.
+
+To run linting, you'll need to install packages in the project root and in web/core:
+
+- cd into web/core with `cd web/core` and run `yarn`
```
ModulesUnraveled, line 122: It's yarn required by Drupal? We generally try to use either yarn or npm on a project. Seeing them both here the up a flag for me

docs/CONTRIBUTING.md
mickey4k: Removed.  Thanks.

composer.json
mickey4k: I ended up having to move everything into composer and bash files.  Turns out shell.exec doesn't return error codes to there was no way to trigger a static_tests fail.

docs/CONTRIBUTING.md
mickey4k: https://www.drupal.org/about/core/policies/core-change-policies/frontend-developer-tools-for-drupal-core

docs/CONTRIBUTING.md
```diff
@@ -104,3 +105,28 @@ Development standards will be rigorously enforced on this project. Drupal mainta
...
+
+### Linting
+
+To run the linting, you'll need to create a Github Personal Access token to be able to pull down the Yale package.
+
+- Visit the [Github instructions for creating a Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token).
+- Give it "Write packages" permissions. This will also check the repo permissions.
+- Export the token from your ~/.bash_profile or ~/.bashrc file by adding:
+  `export YALESITES_BUILD_TOKEN=[token]`
+- Source the file with `source ~/.bash_profile` or `source ~/.bashrc`.
```
ModulesUnraveled, line 117: I wonder if we want to put this all here, or link to the Prerequisites section of the [Linting and Formatting repo](https://github.com/yalesites-org/eslint-config-and-other-formatting#installation)? That one has a link to help people who use other shells, and I consider it to be "canonical." If we ever changed the requirements, I'd like to only have to do it in one place...

ModulesUnraveled: I actually think this should probably be in documentation somewhere more general under a heading like "Setting up your machine to install YaleSites packages" since it also applies to the theme, and anything else that we eventually package as a GitHub package.

.ci/.npmrc-example
```diff
@@ -0,0 +1,4 @@
+scripts-prepend-node-path=true
+always-auth=true
+@yalesites-org:registry=https://npm.pkg.github.com/
```
nJim, line 3: I understand why we need a .npmrc file in general, but why do we keep a separate one in the `.ci` directory? Should this just exist at the project root that way it is useful for local development too? Or is it best practice to not track this file?

.circleci/config.yml
```diff
@@ -93,6 +93,18 @@ jobs:
           name: Setup environment variables
           command: /build-tools-ci/scripts/set-environment
 
+      # Create theme npmrc.
+      - run:
+          name: Create .npmrc.
+          command: mv ./.ci/.npmrc-example ./.npmrc
+
+      # Authencate npmrc.
+      - run:
```
nJim, line 102: I don't think this needs to be done. I believe we are only using public packages so so there is no reason to authenticate. We will need an .npmrc file so that npm knows to look at github, but I think the whole authentication part is not actually adding value for this project.

docs/CONTRIBUTING.md
nJim: I don't believe we need any of this. Since we are pulling public packages, these steps should be unnecessary.

nJim (changes requested):
Looks great. I have only one real change to request. I don't think the github package authentication is necessary. This may also change the instructions and ci stuff a little.

I share Brian's concern about needing yarn. I'm wondering if we can replace the core linting with out own in the future. This is not something we have to take on at this time -- just something to chew on.

docs/CONTRIBUTING.md
ModulesUnraveled: They would be unnecessary if we were hosting on npm. To install GitHub packages, however - even public ones, requires a personal access token

nJim:
Thank you for the clarification Brian. Since the authtoken is required, then I'm marking this as complete and merging it in. Thank you for the awesome attention to detail on this one Mickey. There as a lot of moving pieces. I think we should revisit some of these scripts as a group and look for more ways to automate the process, but this is an amazing first step.

nJim (approved):

PR #22, opened 2022-04-30
YALB-525: Two Column Layout Builder Layout

mickey4k:
### Description of work
- Defines a new ys_layouts module to contain custom layouts for the platform.
- Creates a new layout for a two column layout and provides a template for this layout.
- Enables the ys_layout module. Also enables contextual links. Not important for this PR, be we will leave it in there for now unless there are any objections.

### Functional testing steps:
- [x] [Login to the multidev](https://pr-22-yalesites-platform.pantheonsite.io/user) (password in JIRA ticket.)
- [x] Navigate to the [Layout Paragraph Sections](https://pr-22-yalesites-platform.pantheonsite.io/admin/config/content/layout_paragraphs/sections) settings page.
- [x] Click the 'Use as a Layout Section' for one of the paragraphs (any of them.) 
- [x] Verify that 'YaleSites Layouts' appears as an optgroup in this list of layouts and a 'Two column' layout exists in this group.

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/templates/layouts/two-column/ys-layouts--yds-twocol.html.twig
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/templates/layouts/two-column/ys-layouts--yds-twocol.html.twig
@@ -1,21 +1,11 @@
-{#
-# In this file, we're checking `if directory` in order to determine if the
-# active environment is a Drupal site. We need to do this because Drupal's
-# drag-and-drop interface needs specific attributes on the wrapper elements
-# that would not make sense, or even cause errors in environments like
-# Storybook.
-#}
-
-{% set two_column__base_class = 'yds-two-column' %}
-
-<div {{ attributes.addClass(two_column__base_class) }} data-component-width='max' data-embedded-components>
-    <div {{ attributes.addClass("#{two_column__base_class}__inner") }}>
-        <div {{ region_attributes.primary.addClass("#{two_column__base_class}__primary") }}>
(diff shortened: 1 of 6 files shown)


web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/templates/layouts/two-column/ys-layouts--yds-twocol.html.twig
```diff
@@ -0,0 +1,24 @@
...
+{% set two_column__base_class = 'yds-two-column' %}
+
+<div {{ attributes.addClass(two_column__base_class) }} data-component-width='max' data-embedded-components>
+    <div {{ attributes.addClass("#{two_column__base_class}__inner") }}>
+        <div {{ region_attributes.primary.addClass("#{two_column__base_class}__primary") }}>
+            {% block two_column__primary %}
+                {{ content.primary }}
+            {% endblock %}
+        </div>
+        <div {{ region_attributes.secondary.addClass("#{two_column__base_class}__secondary") }}>
```
ModulesUnraveled, line 18: I think in this file, we can delete the data-attributes and probably just write the classes manually, rather than using the variable set at the top. (Drupal core doesn't do the variable thing. We do it because we use the BEM function to generate classes, but since this file doesn't necessarily know about our component library etc, this one can be "Drupal standard")

ModulesUnraveled (approved):
I think this is the right starting point. I may discover more needs once I start implementing it more, but it seems to do everything I need it to for now. Thanks! I'm going to merge it so that I can keep working on the Layout Paragraphs stuff, but please review my last two commits for sanity. 👍

PR #24, opened 2022-05-02
YALB-516: Create Admin Theme

ModulesUnraveled:
[YALB-516: Create Admin Theme](https://yaleits.atlassian.net/browse/YALB-516)

### Description of work
- Creates an admin theme that we can add to/update later

### Functional testing steps:
- [ ] Verify the theme is installed and set as the admin theme... That's about it.


web/profiles/custom/yalesites_profile/config/sync/block.block.alertblock.yml
```diff
@@ -1,21 +0,0 @@
-uuid: a0ac4316-3e39-4a46-a8d1-edb87feb0d52
```
nJim, line 1: Should this have been deleted? Looks like we lost a block along the way.

web/profiles/custom/yalesites_profile/config/sync/core.extension.yml
```diff
@@ -82,5 +82,5 @@ theme:
   atomic: 0
   claro: 0
   gin: 0
-  bartik: 0
```
nJim, line 85: I could have sworn this was just added as a dependency for the gin theme, but I can not find a reason in the code to backup my assumption. So I suppose it's okay that bartik is going away. Not sure why it was added in the first place.

web/profiles/custom/yalesites_profile/themes/custom/ys_admin_theme/ys_admin_theme.info.yml
```diff
@@ -0,0 +1,36 @@
...
+  highlighted: Highlighted
+  help: Help
+  content: Content
+  page_top: "Page top"
+  page_bottom: "Page bottom"
+  sidebar_first: "First sidebar"
+regions_hidden:
+  - sidebar_first
+
+components:
```
nJim, line 25: Is this accidental copy paste?

web/profiles/custom/yalesites_profile/config/sync/core.extension.yml
ModulesUnraveled: Gin uses Claro as a base theme, so I don't think Bartik is needed...

web/profiles/custom/yalesites_profile/themes/custom/ys_admin_theme/ys_admin_theme.info.yml
ModulesUnraveled: No, the admin theme needs to know about these components because of the way layout builder works... I'll verify while/after I get some things ironed out in the authoring experience stuff, but if we use layout builder on the backend (which I think we will) the admin theme needs to know where our layouts are... which might actually end up pointing to your module instead of this theme in the end...

web/profiles/custom/yalesites_profile/config/sync/block.block.alertblock.yml
ModulesUnraveled: Not sure... I just did a config import, and I have it, but its machine name is `alert_block` (with an underscore) so maybe we had two before?

nJim (commented):
Look appropriate. This is a good starting place for upcoming overrides.

PR #30, opened 2022-05-11
YALB-519 - adds Editoria11y module

mickey4k:
### Description of work
- Adds and configs Editoria11y module.

### Functional testing steps:
- [x] Add some content that has obvious accessibility issues such as a skipped heading levels or a table with no header cells or a link with a url as the link text. 
- [x] Save the page.  See the editoria11y widget in the lower right.  Click on it to update it's feedback.


nJim (approved):
Looks good. Ship it 🚀

PR #33, opened 2022-05-16
YALB-451: BE: Roles and Permissions

vinmassaro:
### Description of work
- Adds site_admin and platform_admin user roles
- Adds default permissions for each role

### Functional testing steps:
- [ ] Visit `admin/people/roles` and see "Platform administrator" and "Site administrator" roles
- [ ] Visit `admin/people/permissions` and see new permissions set for both roles


nJim (approved):
Tested on local and this is exactly what we need to get started. Ship it.

PR #44, opened 2022-05-24
YALB-536: configs layout_paragraphs modal

mickey4k:
### Description of work
- Fixes modal resizing by configures modal settings

### Functional testing steps:
- [x] Add a Text With Image paragraph and edit from the frontend.  The modal should automatically resize.


nJim (approved):
So glad to see this is a setting we can change. Brian and I thought we were going to need tom custom work. This fixes a very annoying problem. Good job. Ship it!

PR #45, opened 2022-05-25
YALB-574 - Install demo content

mickey4k:
### Description of work
- Adds demo content
- Updates setup script to build from config instead of database

### Functional testing steps:
- [ ] Run `npm run setup`
- [ ] Make sure the 7 demo nodes are present

### Notes:
Remaining todo:
Standard Page Two  - There were no fields on the Section paragraph.  The Section paragraph will need to be built out first.  Where can I find the information to do that?


web/profiles/custom/yalesites_profile/content/news.content.yml
```diff
@@ -0,0 +1,16 @@
...
+  field_content:
+    - entity: 'paragraph'
+      type: 'text'
+      status: 1
+      field_text:
+        - format: 'basic_html'
+          value: |
+            <p>This will be a page with news articles listed or if views aren't ready it will be a category with articles as the subpages.</p>
+            <p>&nbsp;</p>
+            <p><a data-entity-substitution="canonical" data-entity-type="node" data-entity-uuid="97a702b9-efa2-4814-b038-566725e84cba" href="/node/7">Longform article&nbsp;</a></p>
```
nJim, line 14: I think we can get rid of these data attributes as they were generated by Drupal in a previous sandbox.

nJim (changes requested):
Great work Mickey. I appreciate the attention to detail. Only one small change requested to remove data attributes. I also left a comment in JIRA about ignoring the 'section' work for now as we want to prioritize this against other work.

I also added the 'dont merge' tag to this branch. This change adds about 50MB of images. I'm going to see if we can host these externally or add git LFS to this project to avoid bloat i the repo. I can take that on.

PR #57, opened 2022-06-23
fix(YALB-606): adds heading field to cta banner.

codeb-a:
### Description of work
- Adds heading field to cta banner
- updates the cta banner paragraph field so the page <title> is no longer used as the CTA banner heading.

### Functional testing steps:
- [x] log in
- [x] Navigate to content/page/banner
- [x] add cta banner to banner


nJim (approved):
Everything looks great and works as expected. Thank you for this change. I will need this field in another ticket that I am working on. Ship it.

PR #58, opened 2022-06-24
ci: enable Github Actions

vinmassaro:
### Description of work
- Enables Github actions workflows.
- This is an initial implementation to move from CircleCI and it may be buggy.
- Static tests have been commented out because they are failing in Github Actions, so they will need to be investigated later.


nJim (approved):
Cant say that I understand this line-by-line, but the work is very encouraging. I think there is a lot to tweak and cleanup, but that will come with time. There is no risk to us enabling this work and turning off CircleCI to help move this project forward. Great work Vincent!

PR #64, opened 2022-07-01
YALB-585: Allow banner in page content

nJim:
### Description of work
- Enables the CTA banner as a valid paragraph in the page content
- Adds help text to the banner field.

### Functional testing steps:
- [ ] Create a page and verify that a 'CTA Banner' is valid option for the content field.


ModulesUnraveled (changes requested):
Everything works as expected. However, the Banner heading field is not required, but the component template expects it, so if it's not provided in Drupal, there's an empty heading tag.

We should either:
1. Make the heading field required (I think this is the right option)
2. Add conditionals to the component template so we don't render empty tags

nJim:
Fixed issue identified by Brian by making the heading field required and merging the work.

PR #65, opened 2022-07-07
YALB-610: Wire Pager

ModulesUnraveled:
### Description of work
- Updates the news and events views to support passing the heading from the paragraph down to the view

### Functional testing steps:
- [ ] Import config, and then follow the steps in the [Atomic PR](https://github.com/yalesites-org/atomic/pull/28)


nJim (approved):
This approach works for me. We should keep an eye on these views arguments and see if they grow too much. But for right now it's a lightweight solution. Approved.

PR #70, opened 2022-07-20
YALB-443: Add YS Core module with footer social links UI

marcb4k:
### Description of work
- Adds a new module, ```ys_core```
-  UI form at ```/admin/yalesites/footer``` to save social links into config
- Adds a block for social links to be printed to the page
- Template override for theme is ```social-links-block.html.twig```

### Functional testing steps:
- [ ] Enable the YS Core module as usual
- [ ] Visit ```/admin/yalesites/footer``` to view the form
- [ ] Enter in some URL's for social networks
- [ ] Verify that URL's are the only type of data that can be entered
- [ ] Save the changes
- [ ] Visit the block layout page and add the Social Links Block to an active region
- [ ] Save the block
- [ ] Verify on the front-end that you see a social links block with links
- [ ] Note that only links that have been entered will appear


web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesSocialLinksBlock.php
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Drupal\ys_core\Plugin\Block;
+
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Block\BlockBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+
+/**
+ * Provides a 'Hello' Block.
```
nJim, line 11: Oops. Boilerplate is showing.

web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesSocialLinksBlock.php
```diff
@@ -0,0 +1,72 @@
...
+    $socialWeibo = $config->get("ys_core.social_weibo_link");
+
+    return [
+      '#theme' => 'ys_social_links',
+      '#social_links' => [
+        'facebook' => $socialFacebook,
+        'instagram' => $socialInstagram,
+        'twitter' => $socialTwitter,
+        'youtube' => $socialYouTube,
+        'weibo' => $socialWeibo,
```
nJim, line 67: Since this is a 'social links' block, I'm not sure there is any value added by wrapping the links within an array. Instead, I recommend making these individually named variables to make a stricter contract between the theme hook and the template. 
```
        '#theme' => 'ys_social_links',
        '#facebook' => $socialFacebook,
        '#instagram' => $socialInstagram,
        '#twitter' => $socialTwitter,
        '#youtube' => $socialYouTube,
        '#weibo' => $socialWeibo,
        ```

File: web/profiles/custom/yalesites_profile/modules/custom/ys_core/templates/ys-social-links.html.twig [hunk 3]
```diff
@@ -0,0 +1,15 @@
+{% if social_links.facebook %}
```
nJim, line 1: See comment above. Since this is already a social links component, the wrapper array only adds ambiguity to the shape of the variable. This can simply be `instagram` as a variable.
https://github.com/yalesites-org/yalesites-project/pull/70/files#r927278582

web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Add template file for social links.
+ *
+ * Implements hook_theme().
+ */
+function ys_core_theme($existing, $type, $theme, $path): array {
+  return [
+    'ys_social_links' => [
+      'variables' => [
+        'social_links' => [],
```
nJim, line 17: See above comment. Since this is already a social links component, the wrapper array does not provide value. Instead it adds ambiguity to the shape of the data. https://github.com/yalesites-org/yalesites-project/pull/70/files#r927278582 Suggest changing this to:
```
  return [
    'ys_social_links' => [
      'variables' => [
        'facebook' => NULL,
        'instagram' => NULL,
        ...etc...
      ],
    ],
  ];
```

web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.permissions.yml
```diff
@@ -0,0 +1,2 @@
+administer ys_core_footer_settings:
+  title: 'Administer YaleSites Footer Settings'
```
nJim, line 2: Let's remove this file as per https://github.com/yalesites-org/yalesites-project/pull/70/files#r927282639

nJim (changes requested):
Excellent work. Everything is well organized and follows modern Drupal best practices. I left some nit-picky changes in the PR. 

One big thing is missing and that is updated configuration that is exported to the profile. Once enabled on your localdev, please export your configuration. This should result in one new file (ys_core.settings.yml) and one updated file (core.extensions.yml) to reflect the changes.

web/profiles/custom/yalesites_profile/modules/custom/ys_core/config/install/ys_core.settings.yml
```diff
@@ -0,0 +1,6 @@
+ys_core:
+  social_facebook_link: ''
+  social_instagram_link: ''
+  social_twitter_link: ''
+  social_youtube_link: ''
+  social_weibo_link: ''
```
nJim, line 6: By default, configuration will be wiped out when changes are pushed to the downstream site. That is, the links stored in the database will be repalced on config-import each time we deploy changes. Please at this configuration file to the list of configs to ignore. This is only a partial solution and we will need additional modules in the future to support these per-site configs. admin/config/development/configuration/ignore

nJim (approved):
Looks great. Thank you for all the changes. Everything is in working order. Ship it!

PR #71, opened 2022-07-21
YALB-637: DevOps: add support for building out the frontend

vinmassaro:
### Description of work
- Adds a build step to the CI workflow to check if PRs in this repo have a matching branch name in the atomic or component-library-twig repos. If so, build those branch versions so that we can see changes from all three repos in the same multidev environment.

### Functional testing steps:
- [ ] Given branch name `my-branch`:
  - [ ] Create branch `my-branch` with changes in https://github.com/yalesites-org/atomic
  - [ ] Create branch `my-branch` with changes in https://github.com/yalesites-org/component-library-twig
  - [ ] Create branch `my-branch` with or without changes in this repository 
  - [ ] In the generated multidev environment, verify that changes made in all repos can be seen


.github/workflows/build_deploy_and_test.yml
```diff
@@ -107,19 +107,51 @@ jobs:
...
+          fetch-depth: 0
+      - name: create .npmrc
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        run: |
+          echo @yalesites-org:registry=https://npm.pkg.github.com >> .npmrc
+          echo //npm.pkg.github.com/:_authToken="$GITHUB_TOKEN" >> .npmrc
+      - name: build frontend components
+        env:
+          BRANCH: ${{ github.head_ref || github.ref_name }}
```
ModulesUnraveled, line 142:
```suggestion
          BRANCH: ${{ github.head_ref || github.ref_name }}
          YALESITES_BUILD_TOKEN: ${{ secrets.YALESITES_BUILD_TOKEN }}
```

Does adding this here fix the authentication issue?

ModulesUnraveled: We may not need to do lines 134-139 at all, if so. (We can keep the existing `.npmrc` file, and just pass the token as an environment variable like we do locally)

vinmassaro: After I rebased from develop, `.npmrc` exists in the repo now, so this suggestion worked. Thanks!

vinmassaro:
@nJim and @ModulesUnraveled: this PR is ready for review.

There is a matching branch named `gh-ci-test` in the atomic and component-library-twig repos. Since this doesn't really begin working until this PR is merged, review of the code changes is appreciated. I left in a step to start a tmate session if the `build_frontend` step fails, so we can SSH into the runner and see the current state of things if we begin to see failures or unexpected behavior after these changes are merged. Thanks!

.ci/build/build_frontend
```diff
@@ -0,0 +1,58 @@
...
+cd "$GITHUB_WORKSPACE"/web/profiles/custom/yalesites_profile
+
+# Require the branch version if it exists.
+if [ -n "$(git ls-remote --heads https://github.com/yalesites-org/atomic.git "$BRANCH")" ]; then
+  composer require --prefer-dist --no-update yalesites-org/atomic:dev-"$BRANCH"
+fi
+
+cd "$GITHUB_WORKSPACE"
+rm -rf "${GITHUB_WORKSPACE:?}"/"$ATOMIC_PATH"
+composer update --prefer-dist
```
nJim, line 22: I believe all of this can be done in the if conditional block? If the branch does not exist, then all of the files should be in the right place.

nJim (approved):

nJim:
Merging work so we can see the workflow in action. There were some comment about possible tweaks to the bash script. No work was required, but as this grows we may want to find ways to break it apart and organize it. This looks great. Thanks Vincent!

PR #72, opened 2022-07-25
YALB-614: Follow menu trail

marcb4k:
### Description of work
- Adds menu breadcrumb module
- Breadcrumbs will follow menu structure, not path structure

### Functional testing steps:
- [x] Add the breadcrumb block to a region
- [x] Add some menu items and make some of those items subs of other menu items
- [x] Verify that the breadcrumb for these pages matches the menu structure

### DIFF
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -37,6 +37,7 @@
         "drupal/mailsystem": "^4.3",
         "drupal/maxlength": "^2.0@RC",
         "drupal/menu_admin_per_menu": "^1.4",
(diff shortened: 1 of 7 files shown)


nJim (changes requested):
Everything works great. Based on Brians feedback, let's add the breadcrumbs block to the content region. Somewhere near the top of the page. It's a judgement call. We may remove this in the future and use twig_tweak instead, but for now let's place the block on the page so Brandon and wire the component.

marcb4k:
OK, added the breadcrumb block to the content region.

nJim (approved):
Solid. Approved. 🚢 it.

PR #74, opened 2022-07-28
YALB-656: Adds CKEditor plugins to handle email, phone, and new window

marcb4k:
### Description of work
- This adds several CKEditor plugins to handle easily handle adding email and phone links and also adds the ability to open links in a new window.
- One of the plugins is a custom plugin forked from the original [CKEditor Link plugin](https://ckeditor.com/cke4/addon/link) and is stored in a [public repo attached to yalesites-org](https://github.com/yalesites-org/ckeditor-advanced-link-plugin). This gives us the ability to change the title text of the button and change the icon to distinguish the plugin from the default Drupal link plugin.
- Note: Currently there is a known bug with the modal within modal where this second link button will not allow you to enter text inside the modal.

### Functional testing steps:
- [x] Add a new page
- [x] Add a paragraph of type "Text"
- [x] Verify there are 2 link buttons on the toolbar - the first is the Drupal link that uses Linkit, the second is the new "Email/Phone/Regular" link to allow for email and phone links
- [x] Once we get the bug fixed (as noted above), to change the type of link, select the "Link Type" select list and choose either "Email" or "Phone"
- [x] To open links in a new window for the first link button, expand the "Advanced" accordion and turn on the "Open in new window/tab"
- [x] To open links in a new window for the new link button, click on the "Target" tab at the top of the modal and select "New Window (_blank).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,6 +10,30 @@
     "drupal": {
       "type": "composer",
       "url": "https://packages.drupal.org/8"
+    },
+    "ckeditor-plugin/link": {
+      "type": "package",
+      "package": {
+        "name": "ckeditor-plugin/link",
+        "version": "4.19.3",
+        "type": "drupal-library",
+        "dist": {
+          "url": "https://github.com/yalesites-org/ckeditor-advanced-link-plugin/archive/refs/heads/main.zip",
+          "type": "zip"
+        }
+      }
+    },
+    "ckeditor-plugin/fakeobjects": {
+      "type": "package",
+      "package": {
+        "name": "ckeditor-plugin/fakeobjects",
+        "version": "4.19.1",
+        "type": "drupal-library",
+        "dist": {
+          "url": "https://download.ckeditor.com/fakeobjects/releases/fakeobjects_4.19.1.zip",
+          "type": "zip"
+        }
+      }
     }
   },
   "require": {
@@ -55,6 +79,9 @@
         "[project-root]/.gitattributes": false
       }
     },
+    "installer-types": [
+      "ckeditor-plugin"
+    ],
     "installer-paths": {
       "web/core": [
         "type:drupal-core"
@@ -85,6 +112,9 @@
       ],
       "web/private/scripts/quicksilver/{$name}/": [
         "type:quicksilver-script"
+      ],
+      "web/libraries/ckeditor/plugins/{$name}": [
+        "type:ckeditor-plugin"
       ]
     },
     "composer-exit-on-patch-failure": true,
--- a/web/profiles/custom/yalesites_profile/config/sync/filter.format.basic_html.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/filter.format.basic_html.yml
@@ -16,7 +16,7 @@ filters:
     status: true
     weight: -10
     settings:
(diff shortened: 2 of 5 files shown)


web/profiles/custom/yalesites_profile/config/sync/filter.format.basic_html.yml
```diff
@@ -16,7 +16,7 @@ filters:
     status: true
     weight: -10
     settings:
-      allowed_html: '<em> <strong> <cite> <blockquote cite> <code> <ul type> <ol start type> <li> <dl> <dt> <dd> <h2 id> <h3 id> <h4 id> <h5 id> <h6 id> <p> <br> <span> <a href hreflang data-entity-substitution data-entity-type data-entity-uuid title> <table> <caption> <tbody> <thead> <tfoot> <th> <td> <tr> <sup> <sub>'
+      allowed_html: '<em> <strong> <cite> <blockquote cite> <code> <ul type> <ol type start> <li> <dl> <dt> <dd> <h2 id> <h3 id> <h4 id> <h5 id> <h6 id> <p> <br> <span> <table> <caption> <tbody> <thead> <tfoot> <th> <td> <tr> <sup> <sub> <a href hreflang data-entity-substitution data-entity-type data-entity-uuid title target !href>'
```
nJim, line 19: @marcb4k Andy idea what `!href` is?

composer.json
```diff
@@ -10,6 +10,30 @@
...
       "url": "https://packages.drupal.org/8"
+    },
+    "ckeditor-plugin/link": {
+      "type": "package",
+      "package": {
+        "name": "ckeditor-plugin/link",
+        "version": "4.19.3",
+        "type": "drupal-library",
+        "dist": {
+          "url": "https://github.com/yalesites-org/ckeditor-advanced-link-plugin/archive/refs/heads/main.zip",
```
nJim, line 21: Great approach and I love what you did with the icon. My only concern is that this repo you are pointing to is not actually a fork. It is a copy with no relationship to the original project. So if there are updates, we can not pull them in. HOWEVER, this will get replaced with CKEditor5 in the near future, so I don't think that I am concerned.

nJim (changes requested):
Everything looks great and is working as expected. There are problems with the model and some other rough edges that we can address in future work. Thank you for digging around to get a handle on all of these plugins and to figure out what is possible. I have one set of possible changes where we should explore the CKEditor 5 even though it is in experimental mode. I will reach out on slack with details.

web/profiles/custom/yalesites_profile/config/sync/filter.format.basic_html.yml
marcb4k: Hum, I didn't add/change that from what it was but that is odd. I can double-check that one and fix/remove it. It obviously should allow href!

composer.json
marcb4k: Yah, I thought about that but to fork it would have caused the entire CKEditor 4 repo to come in, unless you know of a way to just fork a specific directory?

nJim:
This work is on hold for the foreseeable future. I will close this PR without merging. We will keep this branch alive for a little while until we decide the next steps. It may be that the next version of Drupal give us more direction and we can avoid custom work.

PR #75, opened 2022-08-03
YALB-635: Add custom breadcrumb block to change link item key from text to title

marcb4k:
### Description of work
- Adds a custom breadcrumb block called "YaleSites Breadcrumbs"
- The ```items``` array now contains links that have a ```title``` key instead of the ```text``` key

### Functional testing steps:
- [ ] Enable YS Core module (if not already enabled)
- [ ] Add the YaleSites Breadcrumbs block to a region
- [ ] Use twig template from ```ys_core/templates/ys-breadcrumb-block.html.twig``` in your theme
- [ ] Verify that ```item.title``` exists and works as expected

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
@@ -13,7 +13,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  *
  * @Block(
  *   id = "ys_breadcrumb_block",
- *   admin_label = @Translation("Breadcrumbs"),
+ *   admin_label = @Translation("YaleSites Breadcrumbs"),
  *   category = @Translation("YaleSites Core"),
  * )
  */
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace Drupal\ys_core\Plugin\Block;
+
+use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Breadcrumb\BreadcrumbBuilderInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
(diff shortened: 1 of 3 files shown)


web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /**
+   * Turns off block cache.
+   *
+   *   Returns 0 to disable block cache.
+   */
+  public function getCacheMaxAge() {
+    return 0;
+  }
```
nJim, line 96: I'm not a caching expert so I get a little lost in the details, but I know this relates to a longstanding issue where the max-age of a block is not bubbled up to page level cache controls. This means that updates may not get to anonymous users or at least that the caches are not invalidated when they should be. https://www.drupal.org/project/drupal/issues/2352009

I believe a better approach is for us to cache this block per route. If something changes on a route (like when a node is updated) then the cached version of that block is updated. So we can replace the `getCacheMaxAge` method with:

```
  use Drupal\Core\Cache\Cache;
  
  
  /**
   * {@inheritdoc}
   */
  public function getCacheContexts() {
    return Cache::mergeContexts(parent::getCacheContexts(), ['route']);
  }
```

Please take a look at this possible solution and let me know if you think it will work.

web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * The current route match.
+   *
+   * @var \Drupal\Core\Routing\RouteMatchInterface
+   */
+  protected $routeMatch;
+
+  /**
+   * Constructs a new SystemBreadcrumbBlock object.
```
nJim, line 37: Small typo. I'm guessing `SystemBreadcrumbBlock` was copied from some boilerplate.

nJim (changes requested):
Looks great. I like the approach and I see us doing similar work for other parts of the YaleSite's platform. That is, creating our own middleware to control what data is passed to the theme layer.

I've left only one big comment on the caching piece. Take a peek and let me know if this looks like a good solution.

ModulesUnraveled (approved):
It looks like we're getting what we need on the frontend. So, whenever you get the caching figured out, I say shipit.

marcb4k:
@nJim - Good catch and thank you for pointing me in the right direction. In addition to your change, I also added an extra cache tags check so if the node changes (path) or menu changes (title) then the cache will be rebuilt.

nJim (approved):
Works great. :shipit:

PR #78, opened 2022-08-10
YALB-635: Add is_active to breadcrumb items to denote active page

marcb4k:
### Description of work
- Adds ```is_active``` to each breadcrumb item to denote if it is the active page

### Functional testing steps:
- [ ] Nest some main menu items under each other
- [ ] If using the default module template, it is now using the ```is_active``` to not show a link for that breadcumb item on the current page
- [ ] If using another twig template, inside the for loop, you can use ```item.is_active``` that will denote the current page

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
@@ -75,7 +75,6 @@ class YaleSitesBreadcrumbBlock extends BlockBase implements ContainerFactoryPlug
     $links = [];
 
     foreach ($breadcrumbs as $breadcrumb) {
-
       array_push($links, [
         'title' => $breadcrumb->getText(),
         'url' => $breadcrumb->getUrl()->toString(),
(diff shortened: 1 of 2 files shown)


web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
```diff
@@ -75,9 +75,11 @@ public function build() {
     $links = [];
 
     foreach ($breadcrumbs as $breadcrumb) {
+
```
ModulesUnraveled, line 78: Is adding this extra blank line standard formatting?

web/profiles/custom/yalesites_profile/modules/custom/ys_core/templates/ys-breadcrumb-block.html.twig
```diff
@@ -3,12 +3,13 @@
...
  #   - title: Link title
  #   - url: Link url
+ #   - is_active: (bool) is current page
  #}
 
 <ul>
   {% for item in items %}
     <li>
-      {% if item.url %}
+      {% if not item.is_active %}
```
ModulesUnraveled, line 12: There are indentation issues here. And I'm wondering if we should make this a little more obvious by removing the `not`. So:

```
{% if item.is_active %}
  {{ item.title }}
{% else %}
  <a href="{{ item.url }}">{{ item.title }}</a>
{% endif %}
```

ModulesUnraveled (changes requested):
This totally works! Thank you!

Just one formatting issue, one question about making it more clear, and another about php formatting standards.

marcb4k:
@ModulesUnraveled - OK, fixed. Ready to go?

ModulesUnraveled (approved):
Looks great! :shipit:

PR #79, opened 2022-08-11
YALB-665: Add responsive image style for content width

marcb4k:
### Description of work
- Add responsive image style for content width image
- Change title of Full Width image style to Image - Max Width to match Storybook
- Change default image style for image paragraph to use new content width image

### Functional testing steps:
- [ ] Import config (if not already done)
- [ ] Add an image paragraph to a page
- [ ] Inspect image and make sure the secret is using the new 3x2 image responsive image style

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/responsive_image.styles.full_width.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/responsive_image.styles.full_width.yml
@@ -18,7 +18,7 @@ dependencies:
     - image.style.max_width_800
     - image.style.max_width_960
 id: full_width
-label: 'Full Width'
+label: 'Image - Max Width'
 image_style_mappings:
   -
     image_mapping_type: sizes
--- a/web/profiles/custom/yalesites_profile/config/sync/responsive_image.styles.image_content_width.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/responsive_image.styles.image_content_width.yml
@@ -1,0 +1,45 @@
+uuid: b960773d-0a57-42b5-9597-79fefe948b93
+langcode: en
+status: true
+dependencies:
+  config:
+    - image.style.3_2_1120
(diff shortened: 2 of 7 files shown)


web/profiles/custom/yalesites_profile/config/sync/responsive_image.styles.image_content_width.yml
```diff
@@ -0,0 +1,45 @@
...
+    - image.style.3_2_1760
+    - image.style.3_2_1920
+    - image.style.3_2_2080
+    - image.style.3_2_2240
+    - image.style.3_2_2400
+    - image.style.3_2_320
+    - image.style.3_2_480
+    - image.style.3_2_640
+    - image.style.3_2_800
+    - image.style.3_2_960
```
ModulesUnraveled, line 19: Images that are just "content" shouldn't be cropped. So this should use the `max_width` image styles instead of the 3_2 ones. See /admin/config/media/responsive-image-style/full_width

web/profiles/custom/yalesites_profile/config/sync/responsive_image.styles.full_width.yml
```diff
@@ -18,7 +18,7 @@ dependencies:
     - image.style.max_width_800
     - image.style.max_width_960
 id: full_width
-label: 'Full Width'
+label: 'Image - Max Width'
```
ModulesUnraveled, line 21: I think it's probably a good idea to rename this to match the term used on the frontend. Can you also rename the machine name though? (Since we're still in early alpha, it's okay to introduce breaking changes) I'm not sure where else this is used at the moment though, so that might require some digging

ModulesUnraveled (changes requested):
One change request, and one comment/suggestion

ModulesUnraveled (approved):
Everything looks good to me! Thanks for doing that refactor!

PR #82, opened 2022-08-15
YALB-667 - Prepend home to breadcrumb list

marcb4k:
### Description of work
- Prepend home link to the breadcrumb list

### Functional testing steps:
- [ ] Visit a page that has more than 2 links deep in a menu (so the breadcrumbs are visible)
- [ ] Verify that you can see a home link and that it goes to the home page


nJim (approved):
Perfect solution. Code is reviewed and will get the job done.

PR #85, opened 2022-08-24
YALB-91: Custom cards create paragraph

marcb4k:
### Description of work
- Adds a new paragraph type, custom cards to hold custom card paragraph types
- Adds a new paragraph type, custom card, that is only used inside the custom cards paragraph type

### Functional testing steps:
- [ ] Import config, if not already done (```drush cim```)
- [ ] Verify that there are 2 new custom card paragraph types on ```admin/structure/paragraphs_type```
- [ ] Verify adding custom cards to a page


web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.paragraph.custom_cards.default.yml
```diff
@@ -0,0 +1,58 @@
...
+    - paragraphs_ee
+    - paragraphs_features
+    - text
+id: paragraph.custom_cards.default
+targetEntityType: paragraph
+bundle: custom_cards
+mode: default
+content:
+  field_cards:
+    type: paragraphs
```
nJim, line 20: I'm not sure if it's this value or not, but something in this file should change. You are using "Paragraphs (stable)" but other components are using "Paragraphs Legacy" (see accordion and callout which both have child items.) Please change this to use the legacy version for consistency. I know, it's a nit picky thing that may change in the future.

nJim (changes requested):
Looks great. One change requested.

nJim (approved):
Solid. Thank you. Ship it. Since this is just configuration, Yale can review it when we get to the wiring stage. This will reduce some of the testing load we are throwing on them.

PR #87, opened 2022-08-25
YALB-105: Quick links paragraph

marcb4k:
### Description of work
- Adds a new paragraph type - Quick links

### Functional testing steps:
- [x] If not already done, import the config ```drush cim```
- [x] Verify that there is a new paragraph type, Quick Links, under ```/admin/structure/paragraphs_type```
- [x] Add a new page, and add Quick Links to the content area. Quick links can have as many links as needed.
- [x] After saving the page, verify that the content shows up on the node view.


nJim (approved):
Everything is appearing where it needs to be. I saw your note about the image not being required per the designs and updated the data model to match. Good catch. We will wait to review this work with Yale once we are in the wiring stage. This will limit the amount of testing they will have to conduct. Ship it.

PR #88, opened 2022-08-27
YALB-41: Promoted news/events items

marcb4k:
### Description of work
- Changes wording of core "sticky" field from "Sticky at top of lists" to "Pin to the beginning of list" for node add forms.
- Note: This does not affect the back-end wording for use in views, etc. Those areas will still say "Sticky at top of lists".
- Adds checkbox for "Pin to beginning of list" to events edit page. The News edit page already had this field enabled.
- Changes views sort order for the four views (two each for news and events) to sort first by "sticky" and then by the previously enabled date field.

### Functional testing steps:
- [ ] Add a new news or event node.
- [ ] On the right sidebar, expand "Promotion options" and verify that there is a checkbox for "Pin to the beginning of list"
- [ ] Visit any one of the four view admin pages (```/admin/structure/views/view/event_cards```, ```/admin/structure/views/view/event_list```, ```/admin/structure/views/view/news_cards```, ```/admin/structure/views/view/news_list``` ) and verify that the sort order is set for "Sticky at top of lists" is first.
- [ ] Add some news/events nodes and add the "Pin to beginning of list" for some of the nodes.
- [ ] Revisit / update preview for the view admin page(s) visited above and notice the change in the sort order with "pinned" items at the top of the list.

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.node.event.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.node.event.default.yml
@@ -42,7 +42,6 @@ third_party_settings:
     group_content:
       children:
         - title
-        - field_subtitle
         - field_content
       label: Content
       region: content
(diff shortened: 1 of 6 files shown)


nJim (approved):
Code looks good. I tested the PR and it's exactly what we needed. We may change the label for this button in the future, but that's an easy change. All the hard parts are done. Ship it!

PR #90, opened 2022-08-30
YALB-590: News: Change publication date

nJim:
### Description of work
- Requires and enables the override_node_options module
- Sets permissions for site admins to edit news node 'authored on' date
- Removes the 'promote' option from the manage news form (as we do not need it).
- Fixes warning by checking if sticky form element isset.

### Functional testing steps:
- [x] Create a user with the 'Site Administrator' role. Note: A bug within the CAS module may cause this to fail. Edit `web/modules/custom/yale_cas/yale_cas.module` and comment out the `yale_cas_form_user_form_alter` function as a temporary workaround.
- [x] Login as the site admin. Visit the node add forms for each content type.
- [x] Verify the user can edit the "authored on" date for news but not pages or events.

### DIFF
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -8,8 +8,8 @@
         }
     ],
     "require": {
-        "drupal/admin_toolbar": "^3.1",
         "drupal/address": "^1.10",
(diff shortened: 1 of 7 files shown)


marcb4k (approved):
Functional steps passed. Code looks good. Approved!

PR #93, opened 2022-08-30
YALB-691: Update machine color names to match component library

marcb4k:
### Description of work
- Updates machine color names to match component library

### Functional testing steps:
- [ ] Since there isn't yet anything to test in the UI for this, we can inspect one of the radio buttons for "Blue" on the form at ```/admin/yalesites/themes``` and verify that the value in the DOM is set to ```blue-yale``` instead of simply ```blue```


nJim (approved):
Thank you for following up and combing through this. Just the attention to detail we needed. Approved.

PR #96, opened 2022-08-31
YALB-599: Component overrides fields

marcb4k:
### Description of work
- Adds fields for overriding styles per paragraph for: Callout, Banner, Pull Quote, Text with Image, Quick Links, News and Events (both lists and cards)
- Adds allowed values coming from config to populate the select lists for each field and provides default values for each

### Functional testing steps:
- [ ] Import config: ```drush cim```
- [ ] Import override config: ```drush cim --partial --source=profiles/custom/yalesites_profile/modules/custom/ys_themes/config/install```
- [ ] Create a new page and add any one of the following paragraphs: Callout, Banner, Pull Quote, Text with Image, Quick Links, News and Events (both lists and cards)
- [ ] Verify that there is a tab group on top for Content (default) and Styles
- [ ] In the Styles tab, verify that there are one or more fields (depending on the paragraph) and that the styles select lists are populated (and have defaults set).


web/profiles/custom/yalesites_profile/modules/custom/ys_themes/config/install/ys_themes.component_overrides.yml
```diff
@@ -0,0 +1,77 @@
+callout:
+  field_style_color:
+    values:
+      blue-yale: 'Blue'
+      gray-700: 'Gray'
+      beige: 'Beige'
```
nJim, line 6: Semantics question. What do you think about making in an array instead of a set of properties? Doesn't really matter as long as we are consistant. Just popped in mind.
```yaml
    values:
      - blue-yale: 'Blue'
      - gray-700: 'Gray'
      - beige: 'Beige'
```

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/config/install/ys_themes.component_overrides.yml
```diff
@@ -0,0 +1,77 @@
...
+  field_style_variation:
+    values:
+      'true': 'Featured'
+      'false': 'Secondary'
+    default: 'true'
+event_cards:
+  field_style_variation:
+    values:
+      'true': 'Featured'
+      'false': 'Secondary'
```
nJim, line 76: I know that these are bool in storybook. I'm wondering if it should be. I'm thinking there could be other display modes for cards in in the future so this convention does not give us room to grow. I will start a conversation in Teams. Luckily this is the type of thing that can change later.

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/config/install/ys_themes.component_overrides.yml
```diff
@@ -0,0 +1,77 @@
+callout:
```
nJim, line 1: I think this file is still being config ignored. We should update the rules so that this config is tracked. We will need to add and update these values over time and push the changes out to downstream sites.

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/config/install/ys_themes.component_overrides.yml
nJim: Actually nevermind. We will push this as is. Brian recently suggested that complex card displays should use the future Views-Lite tool. So let's not overthink it at this stage.

nJim (changes requested):
Great work. I like the addition of the config file as it is very readable and organized. Only one change requested -- this config need to be exported and not ignored so that we can update it over time. I also added a syntax question about yaml objects vs arrays but nothing has to change.

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/config/install/ys_themes.component_overrides.yml
marcb4k: Good point, but, yes, if we're consistent I think it's fine to leave as is for now unless we need it to

nJim (approved):
Everything looks great. We will revisit news and events soon to close the loop on those. I'm going to merge in the changes just to close any open work. Good job.

PR #98, opened 2022-09-01
create theme script for development

codeb-a:
### Description of work
- Creates a local script to support wiring tickets when component library work is not-yet-released
- Also exports a change to show images. Unrelated, but needed for a wiring ticket


ModulesUnraveled:
@marcb4k @nJim @vinmassaro I just added you all to review this because I think we've all been doing similar work in various places. This script is to set up a local environment to be able to use code from the component library that's in a not-yet-released branch.

If you could take a look and provide feedback on how we can improve it based on the CI and other work you've done, that would be great!

scripts/local/dev-component-library-install.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+GREEN='\033[1;32m'
+ENDCOLOR='\033[0m'
+
+read -p "Which branch? " BRANCH
+
+echo -e "${GREEN}Move into atomic and checkout develop"
+cd web/themes/contrib/atomic
+git checkout develop
```
marcb4k, line 10: I ran into an issue with npm not using the correct version. I did an ```nvm use``` from the atomic directory and that fixed it. So, perhaps after line 9, add an ```nvm use```

scripts/local/dev-component-library-install.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+GREEN='\033[1;32m'
+ENDCOLOR='\033[0m'
+
+read -p "Which branch? " BRANCH
```
marcb4k, line 6: I'd add the repo name here, i.e. "Which branch from the component-library-twig repo" or something like that so it's clearer which repo you are asking for.

marcb4k (changes requested):
Some small changes, but otherwise worked for me!

scripts/local/dev-component-library-install.sh
ModulesUnraveled: hmm... that's assuming people have nvm installed, but I think that's a standard development tool - at least at 4K. I'll add it for now, and we can tweak it later, if we need to.

scripts/local/dev-component-library-install.sh
ModulesUnraveled: Done

marcb4k (approved):
Chatted with Brian in Slack - Approved!

PR #103, opened 2022-09-09
YALB-681: News/Events: Base path setting and token

nJim:
### Description of work
- Defines a new config for for ys_core settings.
- Adds settings for news and events landing page paths.
- Reorganizes validation the site settings form.
- Creates custom tokens for the new landing page paths.
- Sets up path auto patterns for news and events leveraging these paths.
- Enables the config split as it was disabled.

### Functional testing steps:
- [x] Navigate to the (site settings form)[admin/yalesites/settings] and verify that the new fields exist for news landing page and event landing page paths.
- [x] Try changing the fields. Verify that a starting slash is required.
- [x] Create and publish a new news node. Verify that the URL alias leverages the new landing page path and follows the pattern `	[yale:news_landing_page_path]/[node:created:custom:Y-m-d]-[node:title]` 
- [x] Create and publish a new event node. Verify that the URL alias leverages the new landing page path and follows the pattern `	[yale:event_landing_page_path]/[node:field_event_date:value-custom:Y-m-d]-[node:title]`

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/config_ignore.settings.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/config_ignore.settings.yml
@@ -4,3 +4,4 @@ ignored_config_entities:
   - system.site.name
   - system.site.mail
   - 'system.site.page.*'
+  - 'ys_core.*'
--- a/web/profiles/custom/yalesites_profile/config/sync/config_split.config_split.site_specific_ignored_config.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/config_split.config_split.site_specific_ignored_config.yml
@@ -12,6 +12,6 @@ theme: {  }
 blacklist: {  }
 graylist:
   - 'ys_core.*'
-  - 'ys_themes.*'
+  - 'ys_themes.theme_settings'
 graylist_dependents: true
 graylist_skip_equal: true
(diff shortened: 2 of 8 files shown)


marcb4k (approved):
Functional and code review looks good. Approved! Though a few notes:

1. Validation allows for spaces in the path, but if I added a space, it correctly transformed that into a dash in the URL. I think that's fine unless you want to specifically disallow spaces.
2. Validation allows just a slash with nothing else after that. It works fine for the actual node URLs - it will just be ```sitename.com/news-item```, but if this token will be used eventually for the news or events landing page, that may cause a conflict with the front page.
3. Nothing to change here, but wanted to make a note - if you create a news or event, then go change the site-wide settings, the previously created nodes still have the old URL. I expect this as I wouldn't want the permalinks to change, but wanted to just point that out.

nJim:
@marcb4k All great points.
1. Allowing spaces: Thinking even bigger, this path should follow the rules defined by the pathauto alias settings. Stripping reserved chars, trimming whitespace, etc. Rather than recreate this logic, I instead said that this field must be a valid Drupal path. That is, a node must exist for the news or events landing page. This way the node pathauto validation will take over. One exception is /news or /events (the default values) are acceptable even if this node does not appear. This feels like a compromise and I would love to talk to Yale folks about the best path forward.
2. Front page: I added a validation to ensure these landing pages can not be the root path "/". This is a good thing for us to enforce as logic would otherwise get messy when using this token in other places. This does not fully address your concerns as folks can still do funky things like have the same URL for the news and events landing page. Any maybe that is okay.
3. I will create a follow up ticket for rebuilding aliased content. Possibly a bulk operation button on this form.

nJim:
Approved by Alyssa with some follow up tickets identified.

PR #104, opened 2022-09-09
YALB-286: Add ability to get all theme settings in twig

marcb4k:
### Description of work
- Adds an additional twig function to get all theme settings in an array

### Functional testing steps:
- [x] In a twig template, add either ```{{ dump(getAllThemeSettings()) }}``` or ```{{ kint(getAllThemeSettings()) }}```
- [x] Verify that the output is an array of setting keys and values

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_themes/src/ThemeSettingsManager.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_themes/src/ThemeSettingsManager.php
@@ -134,6 +134,13 @@ class ThemeSettingsManager {
   }
 
   /**
+   * Gets all theme settings from config.
+   */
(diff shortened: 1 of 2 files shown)


ModulesUnraveled (approved):
Functionally, this is great! I'll let @nJim do the code review though, since I don't PHP

nJim (approved):
This works. I am curious what this is being used for. I want to make sure the storage of these settings can change in the future. A method like this starts to require the config file to exist in a specific way. I will touch base with Brian. The work itself is functional and following standards.

marcb4k:
Brian requested it as sometimes he wouldn't know the exact machine names of the configs. That said, the function can be rewritten if we change storage of the config but the function name remain the same just as long as it outputs an array so I think we're all good.

PR #106, opened 2022-09-13
YALB-636: DevOps: fix tests in the Github Actions workflow

vinmassaro:
### Description of work
- Fix the static tests that run as part of the Github Actions CI workflow


vinmassaro:
@nJim Tests are now working and I fixed most of the initial coding standards issues from `phpcs`. The only one it is erroring on now is the below. Are you able to resolve it? Thanks!

```
FILE: ...rofile/modules/custom/ys_mail/src/Config/MailConfigOverrides.php
----------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
----------------------------------------------------------------------
 58 | WARNING | \Drupal calls should be avoided in classes, use
    |         | dependency injection instead
----------------------------------------------------------------------

nJim (approved):
I fixed that service container issue and now the PHP code is passing. Thank you for cleaning up so many other small issue. I hope the beautifier or some other tool was able to automate this for you.

Approving this work as it accomplishes what is in scope. However it also signs a spotlight on how unorganized the general linting and code tools are. CI calls Composer calls NPM for linting for example. I want to tame this in a future ticket. But hat should not hold this work up. Ship it.

PR #108, opened 2022-09-15
YALB 698: Search Setup

marcb4k:
### Description of work
- Uninstalls the core search module
- Adds search API module, enables the module, and sets initial configuration settings
- Adds the following content types to the index: page, event, news
- Adds the following fields to the index: title, text, heading, subheading, caption, and attribution
- This PR does not include any search form, results pages, or views.

### Functional testing steps:
- [ ] Verify settings at ```/admin/config/search/search-api/index/node_index/edit```
- [ ] Verify fields at ```/admin/config/search/search-api/index/node_index/fields```
- [ ] Verify processors at ```/admin/config/search/search-api/index/node_index/processors```


nJim (commented):
All of the work looks great and is functioning correctly. I have a question on the fields used for indexing. Why do we need to index individual content fields if we are already indexing the rendered entity? Is this redundant? field_teaser_text is the only field that I can think of that does not appear on the rendered entity but may have some indexing value.

nJim (approved):
Thank you. Changes look great. Good call on the permissions. Ship it.

PR #110, opened 2022-09-17
YALB-729: Bug: Creating accordion items

nJim:
### Description of work
- Check if the parent paragraph exists before calling the `has` method. This currently fails when creating new accordion item paragraphs as the parent does not exist until the node is saved.

### Functional testing steps:
- [ ] Create a page with an accordion. Add multiple accordion items and save the node.
- [ ] If the node saves and there are no errors, they the bug is resolved.


ModulesUnraveled (approved):
Perfect, I had noticed this issue, but didn't know what had caused it. Works great! :shipit:

PR #114, opened 2022-09-20
YALB-742: Update action banner

ModulesUnraveled:
## [YALB-742: Update action banner](https://yaleits.atlassian.net/browse/YALB-742)

**This PR requires work in the atomic theme here: https://github.com/yalesites-org/atomic/pull/45**

### Description of work
- Creates css for all themes to fix a lpb issue

### Functional testing steps:
- [x] Visit `/node/1` and verify the "Edit Content" button shows over the banner


nJim (approved):
I have these branches on my local and this looks good. I noticed that the build is failing. I wonder if we should keep this live so that Vincent can help us debug the build process?

PR #115, opened 2022-09-22
YALB-700: Search Form and Settings

marcb4k:
### Description of work
- Adds a search form to the utility navigation area
- Adds a search form to the search view
- Enables additional search processors
- Adds a dynamic meta title to the search page if there are search terms
- Adds number of results to the search view
- NOTE: This requires [PR-46](https://github.com/yalesites-org/atomic/pull/46) from atomic and [PR-147](https://github.com/yalesites-org/component-library-twig/pull/147) from component-library-twig

### Functional testing steps:
- [x] Import config: ```drush cim```
- [x] Reindex search: ```drush search-api-index```
- [x] Visit any page and verify that the search box is located in the utility menu area and is styled appropriately (See screenshot)
- [x] Search for something. Examples: "longform" or "interest".
- [x] Verify that the search page has a results counter on top.
- [x] Verify that the meta title (in the tab of the browser) is dynamic and only changes to "Search for [search term]" if there is a search term.

Screen Shot 2022-09-22 at 10 56 36 AM

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
@@ -181,3 +181,16 @@ function ys_themes_theme_suggestions_form_element_label_alter(array &$suggestion
     $suggestions[] = $hook . '__search_form';
   }
 }
+
+/**
+ * Implements hook_metatags_alter().
+ *
+ * If there is a search term, change the title to reflect what was searched for.
+ */
+function ys_themes_metatags_alter(array &$metatags, array &$context) {
+  if (\Drupal::service('path.current')->getPath() == "/search") {
+    if (isset($_GET['keywords']) && $_GET['keywords'] != '') {
+      $metatags['title'] = t("Search for {$_GET['keywords']} | [site:name]");
+    }
+  }
+}
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
@@ -177,7 +177,7 @@ function ys_themes_form_alter(&$form, &$form_state, $form_id) {
  * Add a template to visually hide the search form label and add custom icon.
  */
 function ys_themes_theme_suggestions_form_element_label_alter(array &$suggestions, array $variables, $hook) {
-  if (strpos($variables['element']['#id'], "edit-terms") === 0) {
+  if (strpos($variables['element']['#id'], "edit-keywords") === 0) {
     $suggestions[] = $hook . '__search_form';
   }
 }
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
@@ -181,16 +181,3 @@ function ys_themes_theme_suggestions_form_element_label_alter(array &$suggestion
     $suggestions[] = $hook . '__search_form';
   }
 }
-
-/**
- * Implements hook_metatags_alter().
- *
- * If there is a search term, change the title to reflect what was searched for.
- */
-function ys_themes_metatags_alter(array &$metatags, array &$context) {
-  if (\Drupal::service('path.current')->getPath() == "/search") {
-    if (isset($_GET['keywords']) && $_GET['keywords'] != '') {
-      $metatags['title'] = t("Search for {$_GET['keywords']} | [site:name]");
-    }
-  }
-}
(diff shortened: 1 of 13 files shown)


web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
```diff
@@ -158,3 +158,39 @@ function ys_themes_build_css_variables() : string {
...
+ *
+ * If there is a search term, change the title to reflect what was searched for.
+ */
+function ys_themes_metatags_alter(array &$metatags, array &$context) {
+  if (\Drupal::service('path.current')->getPath() == "/search") {
+    if (isset($_GET['keywords']) && $_GET['keywords'] != '') {
+      $metatags['title'] = t("Search for {$_GET['keywords']} | [site:name]");
+    }
+  }
+}
```
nJim, line 196: The Metatag module is enabled on this site manage these sort of settings. I think we should enable the "Metatag: Views" module so that we can set this pattern within Views. We want to embrace the low-code site building tools whenever possible on this project.

Also just a side note that I get spooked whenever I see $_GET global. It's rarely an issue, but this is one of those random things that leads to possible security issues. (Usually when writing to a DB or interacting with files.) So I tend to always write a validation function when using user-input value. And I typically get these values from symfony instead of using the $_GET directly.

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
```diff
@@ -158,3 +158,39 @@ function ys_themes_build_css_variables() : string {
...
+/**
+ * Implements hook_theme_suggestions_form_element_label_alter().
+ *
+ * Add a template to visually hide the search form label and add custom icon.
+ */
+function ys_themes_theme_suggestions_form_element_label_alter(array &$suggestions, array $variables, $hook) {
+  if (isset($variables['element']['#id']) && strpos($variables['element']['#id'], "edit-keywords") === 0) {
+    $suggestions[] = $hook . '__search_form';
+  }
+}
```
nJim, line 183: I tend to dislike hooks, but they are effective especially when trying to tweak small things without having to recreate something that is in core. None of these things are problematic. But if this list grows, that I would rather create our own block without own template. 

Why all the hate? Who hurt me? It's things like looking at the string position for a specific string. Or checking an autogenerated form id. If these strings change (possibly as part of a core update or installing a third party module) then our code stops firing. 

So my gut says to leave this. It is effective and an appropriate use of Drupal's hooks. But I will also pitch the question over to @ModulesUnraveled. Do you often create a custom search block in Drupal so that you can control the form? Are we near that point or is this just standard theming business?

nJim (changes requested):
One change requested to use the metatag module on the view. I'm glad to see so much of the feedback tucked into this -- iteration for the win. I also posed a question about hooks and welcome a discussion, but the code itself is effective and appropriate.

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
ModulesUnraveled: This is how I've always done it. No one has ever suggested creating a custom block... so I don't have any preference at this time.

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
marcb4k: @nJim - I actually tried the Metatag Views module first, but unfortunately it didn't parse the twig replacement ({{ arguments.search_api_fulltext }}) for the title as we wanted. It left it alone. So, I decided to do it in code. If you think there's another way we can handle this, let me know. I will definitely look into replacing the $_GET, thanks for that suggestion.

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
marcb4k: Thank you for explaining the reasons why hooks can be problematic - I was curious and I understand now. I tried to think of another way we can alter the search form and Drupal doesn't provide template suggestions for this form - so we'd still have to write a template suggestion based on the exposed form name (views-exposed-form-search-page-1) and then also override other templates as well to get the placeholder and ```visually-hidden``` class in there too... so it is possible to do. For now, you said let's ship this version and if we need to later we can refactor?

ModulesUnraveled (approved):
Functionally works. Whenever you and @nJim figure out the code stuff, I say :shipit:

nJim (approved):
Works and the code looks good. Just one thing for you to check...

I was able to get the metatag views piece to work by checking the "use replacement tokens from the first row" box. I've done a lot of custom views plugins work in the past, but I don't understand why this works. An argument should be globally available without having to access a results row. But it works and there is no harm in using this part of the query. I'm happy to replace the _hook with a module. If this change looks good to you, feel free to ship it.

PR #122, opened 2022-09-28
YALB-282: AX: Add button to edit media

nJim:
### Description of work
- Adds the media_library_edit module to the project in the yalesites_profile
- Enables the module via config
- Adds the 'edit' button to all paragraphs using a media entity reference field

### Functional testing steps:
- [ ] Login to the multidev `terminus drush yalesites-platform.pr-122 -- uli`
- [ ] Create a new page
- [ ] Click on the plus sign to add a new paragraph and add a 'Text with Image' component.
- [ ] Add a new image to the paragraph...
  - [ ] Click the 'Add media' button to launch the media library browser. Follow the instructions to 'Add file' to the media library. This may differ from browser to browser. 
  - [ ] Fill out the alternative text and click the 'save' button to add the image to the media library.
  - [ ] With the item in a selected state, click the 'Insert selected' button on the media library browser modal window.
- [ ] Verify that a new edit icon appear on the image thumbnail. This button has a pencil icon.
- [ ] Click the icon. Verify that this open an edit-media modal where the user can change the alt-text and other properties of the media entity.

### DIFF
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -36,6 +36,7 @@
     "drupal/mailchimp_transactional": "^1.0",
     "drupal/mailsystem": "^4.3",
     "drupal/maxlength": "^2.0@RC",
+    "drupal/media_library_edit": "^2.3",
     "drupal/menu_admin_per_menu": "^1.4",
     "drupal/menu_breadcrumb": "^1.16",
     "drupal/metatag": "^1.19",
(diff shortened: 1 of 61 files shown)


nJim:
I need some frontend help. I'm not sure if this is a problem with the module or more likely a gap in how things are styled within the Gin admin theme. When adding the 'Edit Media' link, the link text is appearing when it should be visually hidden.
Screen Shot 2022-09-29 at 3 28 25 PM

ModulesUnraveled:
@nJim Should work now. You might need to clear cache. but the text should be visibly hidden, and not extend past the button (even when it was invisible before, you could still click when the text was to trigger the button)

PR #123, opened 2022-09-28
YALB-762: Bug: Fix CI build - multidev not created

vinmassaro:
### Description of work
Builds are currently failing on the build_frontend step when files being modified are managed by Composer.

This PR updates the build_frontend script to build out the atomic theme dependencies and remove the dependency from composer so that Integrated Composer running on Pantheon does not conflict. This is accomplished by introducing the cut line to `.gitignore`, modifying some of the ignored theme paths, and running `terminus build:gitignore:cut` so that `/web/themes/contrib/atomic/` can be committed when running in CI.

This also fixes an a second issue issue that caused builds to fail around the same time, for another reason: https://github.com/actions/runner/issues/2033

### DIFF
--- a/.ci/build/build_frontend
+++ b/.ci/build/build_frontend
@@ -37,6 +37,9 @@ if [ -n "$(git ls-remote --heads https://github.com/yalesites-org/component-libr
   npm ci --omit=dev --ignore-scripts
   npm install webpack-cli
   npm run build
+  rm -rf node_modules
+
+  # Copy built component library into atomic theme.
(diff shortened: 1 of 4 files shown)


nJim (approved):
I kind of hate that we are doing all of this brute force work to build atomic, but I also don't think we have a choice given the restrictions imposed by Integrated Composer. This change makes sense to me and I'm so glad to see you found a way to get builds working.

I actually do not understand the changes with all of the quotation marks, so I would love to ask you about that in the future.

Let's merge this so we can get builds working. I will start a separate conversation in teams about other possible approaches for the future. Great work!

PR #124, opened 2022-09-28
YALB-736: Feedback: Remove starterkit install hooks

marcb4k:
### Description of work
- Removes install hook that won't do anything for this project. This works to initiate migrations on module install, but not profile install
- Keeps the uninstall hook so if the module is uninstalled, the correct configs are removed

### Functional testing steps:
- [ ] None, this is just maintenance work


nJim (approved):
Approved.

PR #126, opened 2022-09-28
YALB-717: Bug: Breadcrumbs for content without a menu link

marcb4k:
### Description of work
- Fixes bug where content that was not in a menu was showing the breadcrumbs. No breadcrumbs should show for non-menu items

### Functional testing steps:
- [x] Create a new node (page, event, or news) and make sure it is not in the main menu
- [x] Save the page and view it
- [x] Verify that no breadcrumbs show
- [x] Edit the main menu and place one link as a child of another link and save
- [x] Verify that on the child link, the breadcrumbs still show fine

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
@@ -93,11 +93,13 @@ class YaleSitesBreadcrumbBlock extends BlockBase implements ContainerFactoryPlug
     ];
 
     foreach ($breadcrumbs as $breadcrumb) {
-      array_push($links, [
-        'title' => $breadcrumb->getText(),
-        'url' => $breadcrumb->getUrl()->toString(),
-        'is_active' => empty($breadcrumb->getUrl()->toString()),
-      ]);
+      if ($breadcrumb->getText() !== '') {
+        array_push($links, [
+          'title' => $breadcrumb->getText(),
+          'url' => $breadcrumb->getUrl()->toString(),
+          'is_active' => empty($breadcrumb->getUrl()->toString()),
+        ]);
+      }
     }
 
     return [
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
@@ -83,7 +83,7 @@ class YaleSitesBreadcrumbBlock extends BlockBase implements ContainerFactoryPlug
    * {@inheritdoc}
    */
   public function build() {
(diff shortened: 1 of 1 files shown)


ModulesUnraveled (approved):
I'll let @nJim do a code review, but everything functions as described.

web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
```diff
@@ -93,11 +93,13 @@ public function build() {
     ];
```
nJim, line 93: Great. This fixes the issue. I have a suggestion for refactoring the work. Anytime I enter a logic block, I wonder about the business value. In this case the 'if' is about filtering links. This could be a separate method. I prefer breaking working into single-purpose methods. This is for reusability, readability, and hopefully for future unit testing. Take a peek at my suggestion below and we can talk about it. It seems like I am adding a lot of boillerplate but there is a reason for my madness.

```
  /**
   * {@inheritdoc}
   */
  public function build() {
    $breadcrumbs = $this->removeEmptyLinks($this->getBreadcrumbs());
    $links = [
      [
        'title' => $this->t('Home'),
        'url' => $this->urlGenerator->generateFromRoute('<front>', []),
        'is_active' => FALSE,
      ],
    ];

    foreach ($breadcrumbs as $breadcrumb) {
      array_push($links, [
        'title' => $breadcrumb->getText(),
        'url' => $breadcrumb->getUrl()->toString(),
        'is_active' => empty($breadcrumb->getUrl()->toString()),
      ]);
    }

    return [
      '#theme' => 'ys_breadcrumb_block',
      '#items' => $links,
    ];
  }

  /**
   * Get a list of breadcrumb links using the Drupal BreadcrumbBuilder.
   *
   * @return \Drupal\Core\Link[]
   *   An array of Drupal links.
   */
  protected function getBreadcrumbs(): array {
    return $this->breadcrumbManager->build($this->routeMatch)->getLinks();
  }

  /**
   * Remove empty links from a list of links.
   *
   * @param \Drupal\Core\Link[] $links
   *   An array of Drupal links.
   *
   * @return \Drupal\Core\Link[]
   *   An array of Drupal links with empty ones removed.
   */
  protected function removeEmptyLinks(array $links): array {
    return array_filter($links, function($link) {
      return $link->getText() === '';
    });
  }
```

nJim (changes requested):
I'm marking this as 'request changes' but feel free to push back and we can discuss. I like to have small methods whenever possible for a number of reasons. This is something my old boss used to push me into doing and I hated it. Then in 2020 I spent a year doing real Test Driven Development and I saw the light.

web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesBreadcrumbBlock.php
marcb4k: I honestly appreciate your feedback and teaching me new ways to do these things! Making me a better developer! I see what you're doing here. Perhaps a chat at some point about the unit testing you mentioned as I've not really done that before. I'll make those changes and then we can press forward. Thank you!

nJim (approved):
Approved. Ship it.

PR #129, opened 2022-09-30
YALB-702: Fix error message for show/hide search config

marcb4k:
### Description of work
- Fixes error message ```Notice: Trying to access array offset on value of type null in ys_core_preprocess_region() (line 47 of profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module).```

### Functional testing steps:
- [ ] Rebuild to get a clean copy: ```npm run rebuild```
- [ ] Load any page on the front end
- [ ] Verify that there are no more error messages like the one above

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
@@ -43,7 +43,7 @@ function ys_core_form_alter(&$form, $form_state, $form_id) {
  */
 function ys_core_preprocess_region(&$variables) {
   $config = \Drupal::config('ys_core.site');
-  if ($variables['elements']['#region'] == 'header') {
+  if ($variables['elements']['#region'] == 'header' && $config->get('search')) {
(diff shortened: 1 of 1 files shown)


marcb4k:
@nJim - Any idea why the config (```/web/profiles/custom/yalesites_profile/config/sync/ys_core.site.yml```)from the profile isn't setting this correctly?
```
page:
  news: /news
  events: /events
search:
  enable_search_form: 1
```

It should set search.enable_search_form and so that should be available for new sites. But I saw what Brian was seeing - it throws an error. This code will fix that, but I'd like to find the root issue.

ModulesUnraveled (approved):
Functionally works, and makes sense to check if it's set before using it. But like you said, if it should be set by config by default, it'd be good to know why that's not working.

nJim (approved):
Always a good idea to check things before using them. Unfortunately, I could not get this to fail on my local. I'm not sure what was causing the problem. The setting is in the config/sync so it should work. This config is part of the config split, so depending on the order of operations, it possible that the item did not get updated on your local.

PR #131, opened 2022-10-06
YALB-683: News/Events: Breadcrumbs

marcb4k:
### Description of work
- Uses the YaleSites settings for the news/events landing page as the breadcrumb trail when on news/events content types
- Adds the news/event node title to the end of breadcrumbs when on those content types
- Refactors all logic into a new YaleSitesBreadcrumbsManager class
- Adds the custom YaleSites pathauto tokens for news/events to "safe tokens" to allow for correct path parsing
- Note: This only works if the path of the news/event matches the parent path of the landing page. Example: if the path for the landing page for news is ```/news``` and the path for the news article is ```/news/2022-09-16-article-mostly-text``` then the breadcrumbs will show. If the path is different (i.e. the setting was changed and the news/event nodes were not updated) no breadcrumbs will show.

### Functional testing steps:
- [ ] Import config ```drush cim``` to get the new pathauto token settings.
- [ ] Make a note of the current settings for the news landing pages here: ```/admin/yalesites/settings```
- [ ] Create a dummy landing page for news using the path specified in the settings. Make sure it is published.
- [ ] Add this dummy landing page to the main menu.
- [ ] Create a new news node and make sure "Generate automatic URL alias" is checked.
- [ ] Verify that the breadcrumbs show on the news node and that the breadcrumb trail has: 1) a link to the "custom" landing page and 2) the news title shows at the end of the breadcrumb trail.
- [ ] Other tests: 1) Unpublish the news landing page. 2) Disable the landing page in the menu. Remove landing page from the menu. All of these will remove the breadcrumbs from the news node.
- [ ] Verify that sub-levels work: i.e. change the URL of the landing page created above to ```/community/news``` and also change the setting of the news landing page to ```/community/news```. Note, the news nodes will not get the new path so initially the breadcrumbs will not show. Edit the news node and uncheck "Generate automatic URL alias", empty the URL alias, recheck the box, and save the node. This will get the new path and the breadcrumbs should show the correct structure.
- [ ] Test all of the above with events as well.

### DIFF
--- a/.ci/build/build_frontend
+++ b/.ci/build/build_frontend
@@ -37,6 +37,9 @@ if [ -n "$(git ls-remote --heads https://github.com/yalesites-org/component-libr
   npm ci --omit=dev --ignore-scripts
   npm install webpack-cli
   npm run build
+  rm -rf node_modules
+
+  # Copy built component library into atomic theme.
   mv /tmp/component-library-twig "$GITHUB_WORKSPACE"/"$ATOMIC_PATH"/node_modules/@yalesites-org
 
   # Clean up atomic's node_modules and remove gitignores so the changes can be staged.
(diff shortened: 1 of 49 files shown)


marcb4k:
@nJim - All finished with cleanup of code from earlier refactor. Also added in the config for the Menu Breadcrumbs to add the home link via that module.

nJim (approved):
Works great. Ship it.

PR #134, opened 2022-10-06
YALB-752: set max lengths in drupal

codeb-a:
### Description of work
- Updates the set max length text fields for paragraphs in drupal.

### Functional testing steps:
- [x] Go to site
- [x] Go to structure/paragraph types/manage fields/manage form display.
- [x] verify that all text fields for paragraphs are updated to the correct character limit according to the data model.


nJim (approved):
Perfect. Thank you for making all of these updates. I was an issue that was open for a long time. Several folks at Yale will be excited to see this finally in action. Ship it.

PR #135, opened 2022-10-07
YALB-283: AX: Remove confusing remove-button

marcb4k:
### Description of work
- Removes extra unnecessary "remove" button from paragraphs that have cardinality of 1 and the field is required.
- Note: The patch for this was added to the root composer file. If this needs to go into the profile composer file, let me know and I will move it.

### Functional testing steps:
- [ ] Create a page and add a paragraph of type accordion.
- [ ] Verify there is no longer a remove button for the "text" paragraph (internal paragraph)
- [ ] See screenshots for before/after

Before patch:
Screen Shot 2022-10-06 at 3 35 20 PM

After patch:
Screen Shot 2022-10-06 at 3 35 27 PM

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -95,6 +95,9 @@
     "patches": {
       "drupal/layout_paragraphs": {
         "save close tweaks": "https://git.drupalcode.org/project/layout_paragraphs/-/merge_requests/80.patch"
+      },
+      "drupal/paragraphs": {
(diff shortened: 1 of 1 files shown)


nJim (approved):
Approving the change. As you saw in an unrelated conversation, we will be moving these patches to the profile. However you code is still accurate and functional, so let's move it alone. The tech debt is nearly non existent. So glad this patch works.

PR #137, opened 2022-10-12
YALB-777: AX: Atomic theme paragraph previews

marcb4k:
### Description of work
- Correctly styles paragraphs after initial click of "Edit Content" button on the front-end.
- Note: If you then click on the edit button on a specific paragraph, this will open a side panel to edit the paragraph and the side panel still needs styling (like the modal was going to need anyways)
- Thank you to @vinmassaro for finding the gin_lp module which makes this work.

### Functional testing steps:
- [ ] Visit one of the prebuilt pages that has an "Text with Image" components 
- [ ] Hover over the content and click the blue "Edit content" button
- [ ] Verify that the paragraphs will not change styling (i.e. the text with image, the image is the same size and does not shrink) and/or that the CTA banner image doesn't get fuzzy.

### DIFF
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -27,6 +27,7 @@
     "drupal/fast_404": "^2.0@alpha",
     "drupal/field_group": "^3.2",
     "drupal/gin": "^3.0@beta",
+    "drupal/gin_lp": "1.0.x-dev@dev",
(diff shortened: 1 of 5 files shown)


nJim:
I am deleting this branch as this work is being deprioritized. Unfortunately the problems with the layout paragraph frontend editors are much bigger than this ticket. Yale is going to focus on the admin editorial interface rather than supporting two interfaces out of the gate.

PR #138, opened 2022-10-13
YALB-307: redirects on login to frontpage

codeb-a:
### Description of work
- Installs the Redirect After Login module. 
- Redirects all roles to the front page.

### Functional testing steps:
- [ ] Navigate to site
- [ ] Go to configuration/System/Set Log In Destination. 
- [ ] Make sure all roles are redirected to <front>.


nJim (approved):
Just what we needed. I will make a follow up ticket to redirect to the dashboard once route exists. Ship it.

PR #141, opened 2022-10-14
YALB-289: Dashboards: Support form

marcb4k:
### Description of work
- Adds Webform module
- Adds a Support form configured to send requests to Yale Service Now with pre-filled values from the site
- Support form does not save data in Drupal

### Functional testing steps:
- [ ] Visit ```/admin/structure/block``` and click "Place block" in the "Content" region.
- [ ] Click "Place block" next to "Webform"
- [ ] Under "Webform", type "Support" and click on "Support Request"
- [ ] Click "Save block"
- [ ] Visit any page on the front-end
- [ ] Verify that the Webform block is on the page
- [ ] Fill out the "Location", "Contact Number", and "Page You Are Having Trouble With" - this last one is an autocomplete.
- [ ] Click Submit
- [ ] If you aren't logged into your Yale NetID, it will redirect you to the login screen first, proceed with the login.
- [ ] After login, it will redirect to the Service Now "Submit a Ticket" page.
- [ ] Verify that the various fields are pre-filled (Location, Contact number, Short description)
- [ ] In the body of the message, verify that the user agent matches your user agent and that the referral URL is the "page you are having trouble with"


marcb4k:
Since this work has changed scope, new testing instructions: 

- [ ]  Visit /admin/structure/block and click "Place block" in the "Content" region.
- [ ]  Click "Place block" next to "Support Form Block".
- [ ] Click "Save Block" and then "Save Blocks" at the bottom of the block admin interface.
- [ ] Visit any page on the front-end.
- [ ] Verify that the Support Form block is on the page (has a button with the title "Start a Support Request").
- [ ]  Click the button.
- [ ] If you aren't logged into your Yale NetID, it will redirect you to the login screen first, proceed with the login.
- [ ]  After login, it will redirect to the Service Now "Submit a Ticket" page.
- [ ]  Verify that the form contains the following dynamic elements: site URL, site title, user name, user roles, and user agent.

marcb4k:
@nJim - Some notes from this evening's refactor:

1. Redirect after login module disabled and removed from composer
2. Redirect after login code added to ys_core.module - it currently redirects to the front page.
3. Tried to test ```/user/password``` path, but locally I get a mailchimp transactional module error. I tested it on Pantheon and it's fine - this code that was added just changes the redirect URL of the login form so I think the forgot password page should be just fine
4. No need to add a cache tag for that block - it works fine without it for the different users - I think because we're logged in and caching is better with logged in users.
5. Added the block to the dashboard page.

nJim (approved):
Approved. Thank you for taking on all of the small issues and extra work along the way. Ship it.

PR #143, opened 2022-10-16
YALB-768: local development scripts

ModulesUnraveled:
## [YALB-768: local development scripts](https://yaleits.atlassian.net/browse/YALB-768)

**Description of work**
- Updates all of the local scripts to be useful, and work as expected. Follow review steps below for details.

### Setup the project from scratch
- [ ] `lando stop` Kill any running lando instance
- [ ] `docker system prune -a` delete all currently unused docker images and stuff (next time you start, it'll download updated/latest ones)
- [ ] Quit, and restart Docker, just for posterity.
- [ ] Delete your existing `yalesites-project` folder... or rename it or something
- [ ] Clone the repo `git clone git@github.com:yalesites-org/yalesites-project.git`
- [ ] `cd yalesites-project`
- [ ] `npm run setup`
	- Go find something else to do... This will take a LONG time, since it's cloning all of the docker containers from scratch, and doing a full install of the project from scratch. You might think it's frozen a couple of times. IT'S NOT. It just takes a REALLY long time!)
	- You will be asked for your ssh password about mid-way through, so check back occasionally
- [ ] Click the login link, and view your awesome site!

_NOTE: You can re-run the `npm run setup` script later to get a fresh install, and it'll run much faster since you'll already have the docker images cached locally. I just went a little extra-overboard in the steps above to verify new contributors would be successful setting the project up for the first time._


scripts/local/db-get.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if [ -f reference/backup.sql.gz ];
+  then mv reference/backup.sql.gz reference/backup-prev.sql.gz;
```
nJim, line 4: I'm confused by this bit. Seems like you want to keep the old version as a backup so you do not lose it. What's the reason? If you run this command twice then you overwrite it anyhow.

ModulesUnraveled: Yeah, it's really just a light weight "oh crap I wanted to keep using that other db, but I just ran the db:get script, and it's gone now" safeguard. But I didn't think it was necessary to keep anything more than just the previous one. I guess it's not necessary at all, you could always download it directly from pantheon manually. It's something I've used before, but I'm not tied to in any way.

ModulesUnraveled: It would probably be most useful if there was a major database change, but you are already working on a PR before that update happened. This way you could have both database states locally for different branches and import the one you need.

It's definitely an edge case though, so if you're proposing getting rid of this, I wouldn't fight it.

package.json
```diff
@@ -2,31 +2,32 @@
...
-    "local:cl-update": "./scripts/local/dev-component-library-update.sh",
     "lint:php": "composer lint:php",
-    "lint": "./scripts/local/lint.sh",
-    "test": "./scripts/local/test.sh",
-    "husky:commit-msg": "commitlint --edit $1",
-    "husky:pre-commit": "lint-staged"
+    "lint:styles": "stylelint --color {web/modules/custom,web/themes/contrib/atomic,web/themes/custom,web/profiles/custom}/**/*.scss",
+    "local:review-with-cl-branch": "./scripts/local/local-review-with-cl-branch.sh",
+    "local:review-with-atomic-branch": "./scripts/local/local-review-with-atomic-branch.sh",
+    "local:cl-dev": "chokidar 'web/themes/contrib/atomic/component-library-twig/components' -c 'npm run local:review-with-cl-branch'",
```
nJim, line 23: I'm not able to figure out this line. I think you are already doing it, but we should describe some of these in a README file.

nJim (commented):
This is a big step in the right direction. Thank you for cleaning up so much of this. I left a few questions. Feel free to use the feedback in your next round of edits .. I can mark this approved if you want to just move it forward as there is no risk to the application.

scripts/local/db-get.sh
nJim: Let's keep it for now. It doesn't hurt anything and offers a small assurance. But I may change this when we revamp local script in the future.

scripts/local/assets-pull.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+lando pull --code=none --database=test --files=test
```
vinmassaro, line 3: Why does this pull from test?

scripts/local/local-review-with-atomic-branch.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# This script will checkout the specified branch of atomic for local review.
+
+GREEN='\033[1;32m'
+ENDCOLOR='\033[0m'
+
+read -p "Which branch of the atomic repo do you need? " BRANCH
+
+echo -e "${GREEN}Checking out the $BRANCH branch of Atomic"
```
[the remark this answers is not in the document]
vinmassaro: There are a few quick [shellcheck](https://www.shellcheck.net/) fixes that can be made on this file

scripts/local/local-review-with-cl-branch.sh
```diff
@@ -7,14 +7,10 @@
...
 
 echo -e "${GREEN}Move into atomic and checkout develop"
 cd web/themes/contrib/atomic
-nvm use
-git checkout develop
-echo -e "${GREEN}npm ci${ENDCOLOR}"
-npm ci
 echo -e "${GREEN}Delete installed component library${ENDCOLOR}"
 rm -rf node_modules/@yalesites-org/component-library-twig/
 echo -e "${GREEN}Clone component library${ENDCOLOR}"
```
vinmassaro, line 16: There are a few quick [shellcheck](https://www.shellcheck.net/) fixes that can be made on this file

scripts/local/local-review-with-atomic-branch.sh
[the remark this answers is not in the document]
ModulesUnraveled: I fixed this in the follow-up PR Here: https://github.com/yalesites-org/yalesites-project/pull/147

vinmassaro (changes requested):
I added a few minor comments, but other than that, this looks good to me. Thanks for doing this cleanup!

scripts/local/assets-pull.sh
ModulesUnraveled: It shouldn't. I didn't look closely enough. I've updated it in the follow-up PR to pull from dev.

scripts/local/local-review-with-cl-branch.sh
ModulesUnraveled: I've made all of these fixes in the [followup PR](https://github.com/yalesites-org/yalesites-project/pull/147) Thanks!

nJim (approved):
I see the changes in the follow up PR. Thanks Brian. Approved from my point of view. Will also wait for Vincent to sign off to call it ready to merge.

vinmassaro (approved):

PR #144, opened 2022-10-17
YALB-756: Rename pop out image to wrapped image

nJim:
### Description of work
- This is a refactor where the "pop out image" component is being changed to "wrapped image". Any references in code have been updated to use the new name. This spans three repositories:
  - https://github.com/yalesites-org/component-library-twig/pull/154
  - https://github.com/yalesites-org/atomic/pull/54
  - https://github.com/yalesites-org/yalesites-project/pull/144

### Functional testing steps:
- [x] Create your localdev for testing. Since this has a breaking change (renaming things) the automated workflow will not work as the imported database does not match what is in the code. You will need to run the commands below to make sure all repos are on the correct branches, and you have a fresh db install to avoid the breaking changes issues.
	- [x] `npm run local:review-with-atomic-and-cl-branch` (paste in `YALB-756-wrapped-image` for both repos when asked)
	- [x] `npm run build-with-install` to run the site-install script and do a config import etc.
- [x] Create a new page and open the add-component paragraph picker.
- [x] Verify that the 'Wrapped Image' component appears in the list. Select the component to open the form.
- [x] Add some content and verify that the component appears as expected. There should be no functional or visual changes in this work.


codeb-a (approved):
I was able to run the commands and the "Wrapped-Image" component appears in the paragraph picker list. I also added content and can verify that is appears on a page.

nJim:
Let's merge this collection of PRs now. We need to cut new releases, starting with the component library and working our way back to this PR. We can keep working on these things, but we need to get everything back to the default branches to ensure our builds are working correctly.

PR #145, opened 2022-10-17
YALB-793: CI: check for module additions to root composer.json

vinmassaro:
### Description of work
Adds a CI check to make sure root composer.json require array only contains drupal/core*.


nJim (approved):
That gets the job done. Hell of a check. Good work. I'm sure this will save us from making this mistake regularly since this project is organized in this unique way. Approved.

PR #147, opened 2022-10-18
YALB-768: Local Dev Scripts

ModulesUnraveled:
### Description of work
- Creates a theming script
- Documents all of the script updates

### Functional testing steps:
- [x] Review the updated `README.md` and the new `theming.md` docs and make sure everything makes sense.


marcb4k (approved):
This all looks good to me! Approved with a few notes:

1) I noticed that if you perform a db:local-export, it names it with the data and time, so there is no script that will auto re-import that saved database if you want to quickly get back to your work. No problem there, we'd just need to do an import manually. (Reading the date/time would be doable, but a lot more code...)

2) I see the use of ```npm link``` with the local-cl-dev.sh script - I assume this replaces the need to run that older script to update the CL when developing on multiple repos at the same time?

scripts/local/local-cl-dev.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# This script will use npm link under-the-hood (inside the Docker container) to
+# enable frontend developers to work inside Storybook as they normally would,
+# but also make those changes imediately available to the local Drupal instance.
+# This is particularly useful when a wiring ticket surfaces required component
+# library updates.
+
+# Note: Changes to the component library are still tracked ONLY in the component
+# librarn's repository, so in order for the changes to be seen on a remotely
```
marcb4k, line 9: Typo in "library"

ModulesUnraveled:
@marcb4k Yeah, In the past, I've duplicated the backup and/or renamed it to just `backup.sql.gz` and then it can be used by any of the scripts that expect that file. It's a little manual, but this is not a "polished" process. It's still just a "get it done" method.

And yes, using `npm link` is what I would have done from the start, but because we're using Docker, it can't work. Then it dawned on me that since we're already cloning the CL in here anyway, I could just run `npm link` from within the Docker container, and it would work that way! So, yes. That does replace the need for the manual "copy the component library down into node_modules" script.

nJim (approved):
Thank you for putting all of this together. I did not test all of the commands, but they look good and I'm excited to use them in real work. Let's leave this open for Vincent to review as well.

ModulesUnraveled:
Merging now to unblock other work

PR #148, opened 2022-10-19
YALB-330: Admin theme accordion previews

ModulesUnraveled:
### Description of work
- Tweaks the preview of the accordion items to clean up and simplify the admin interface

### Functional testing steps:
- Create a page, and add an accordion with at least two items. Fill in a decent amount of content, and verify the preview does not grow larger than it is as seen in the "AFTER" image below.

**BEFORE**
Screen Shot 2022-10-19 at 3 25 38 PM

**AFTER**
Screen Shot 2022-10-19 at 3 24 14 PM


marcb4k (approved):
Worked great for me. Approved!

nJim (approved):
This is exactly what we needed. I think this is one of the more complex components (because of the nesting) so it will be a good reference going forward. Approved.

PR #156, opened 2022-11-03
YALB-774: TechDebt: Move patch to the profile

marcb4k:
### Description of work
- Moves patches from the root composer.json file to the yalesites_profile composer.json

### Functional testing steps:
- [ ] Not necessary, but if any patches have been added/removed/changed, run ```composer update```
- [ ] Verify that the patches that are currently there still work:
- [ ] Layout paragraphs - on the front-end, click the "Edit Content" above the paragraphs and the save/close button should be on top (where the Edit button used to be)
- [ ] Paragraphs - Edit an accordion on the back-end and there should only be one "remove" button per accordion item, not two.
- [ ] Linkit - Edit a paragraph that has a link field (such as a callout item) and the autocomplete for the link field should include the author and the date/time stamp.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -91,17 +91,6 @@
     "composer-exit-on-patch-failure": true,
     "patchLevel": {
       "drupal/core": "-p2"
-    },
-    "patches": {
-      "drupal/layout_paragraphs": {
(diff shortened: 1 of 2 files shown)


nJim (approved):
Works great. I tested by adding a new patch and the update command worked. Ship it.

PR #158, opened 2022-11-03
YALB-405: Feedback: Breadcrumbs

joetower:
### Description of work
- updating templates to render the breadcrumb block instead of drupal block layout
- review with https://github.com/yalesites-org/atomic/pull/59

### Functional testing steps:
- [x] Checkout this branch and checkout the same branch for https://github.com/yalesites-org/atomic/pull/59
- [x] Create at least two pieces of `page` content and assign each to a menu item hierarchically (first page parent of `about` for example, second page has a parent of `page 1` you created). Make sure to add a `banner` to the `page 2` node.
- [x] Create at least one `event` node and override the `path` settings so that it, too, is a child of `page 2` you created above
- [x] Create at least one `news` node and override the `path` settings so that it, too, is a child of `page 2` you created above
- [x] View each node and verify each piece of content renders the breadcrumbs above the page title

#### page with banner
Screenshot-20221103101728-2505x981

#### News article
Screenshot-20221103101710-1719x613

#### Event node
Screenshot-20221103101706-1474x644


ModulesUnraveled (approved):
Looks good! :shipit:

PR #160, opened 2022-11-03
YALB-810: Feedback: Dashboard - Support Request link displays twice

marcb4k:
### Description of work
- Removes the duplicate support form request block from the dashboard

### Functional testing steps:
- [ ] Config import ```drush cim``` or if using Pantheon multidev, this is already done
- [ ] Visit the dashboard at ```/admin/yalesites/dashboard```
- [ ] Verify that there is only one support request link, not two.


nJim (approved):
Ship it.

PR #162, opened 2022-11-08
YALB-112: overrides text with image paragraph fields

codeb-a:
### Description of work
- Updates Text with Image paragraph fields and overrides them.

### Functional testing steps:
- [x] navigate to site
- [ ] Go to content/add content/page, fill out title and click add content. Add text with image component.
- [ ] Go to styles and verify that the "focus" dropdown has the options "Equal Focused", "Image Focused", and "Content Focused".

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/ys_themes.component_overrides.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/ys_themes.component_overrides.yml
@@ -41,9 +41,9 @@ text_with_image:
     default: feature
   field_style_variation:
     values:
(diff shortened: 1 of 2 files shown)


marcb4k (commented):
Reviewing this, and I see three options that have different values for the inputs, but the text is all the same "Focused". Shouldn't these labels be different?

codeb-a:
@marcb4k The values have been updated. So it should now read as "Equal Focused", "Image Focused", and "Content Focused".

marcb4k (approved):
Works great! Approved!

PR #166, opened 2022-11-10
YALB-770 Search: Forms

marcb4k:
### Description of work
- Replaces the Drupal search form in header and on search page with custom form so searched term will not stay in field after search
- Adds teaser text to search results
- Adds emphasis to the search term in the title of the search results page (This can be edited in the view to add a class for later styling if we need)

### Functional testing steps:
- [x] This requires the atomic branch ```YALB-770-search-refinement``` so when testing locally, run ```local:review-with-atomic-branch``` and use ```YALB-770-search-refinement``` when it asks for the branch name.
- [x] Import config ```drush cim```
- [x] Visit the site and verify that there is a search form in the header
- [x] Type in a search term (ex: long, news, top)
- [x] On the search results page, verify that the title has the search term emphasized
- [x] Verify that there is no search term in either of the search forms prefilled - header or the one on the search page
- [x] Edit one of the nodes that showed up on the search results page and add some teaser text and save
- [x] Review the search results page again and verify that there is teaser text below the "in context" search result.
- [x] Verify that the search results summary is "pluralized" in that if there is only one result, it shows "1 search result" vs multiple search results it will show "X search results"


ModulesUnraveled:
I'm unable to import config. The error doesn't seem related to this PR, but I can't test your changes because of it:

> The import failed due to the following reasons:
  Unexpected error during import with operation rename for paragraphs.paragraphs_type.pop_out_image::paragraphs.paragraphs_type.wrapped_image: The machine name of the &#039;Paragraphs type&#039; bundle cannot be changed.
  Unexpected error during import with operation rename for field.field.paragraph.pop_out_image.field_image::field.field.paragraph.wrapped_image.field_image: Cannot change an existing field&#039;s bundle.
  Unexpected error during import with operation rename for field.field.paragraph.pop_out_image.field_text::field.field.paragraph.wrapped_image.field_text: Cannot change an existing field&#039;s bundle.
  Unexpected error during import with operation rename for field.field.paragraph.pop_out_image.field_caption::field.field.paragraph.wrapped_image.field_caption: Cannot change an existing field&#039;s bundle.

web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/SearchBreadcrumbTwigExtension.php
```diff
@@ -0,0 +1,67 @@
...
+
+        // // Process all breadcrumbs.
+        // foreach ($breadcrumbs['#links'] as $breadcrumb) {
+        //   array_push($links, [
+        //     'title' => $breadcrumb->getText(),
+        //     'url' => $breadcrumb->getUrl()->toString(),
+        //     'is_active' => empty($breadcrumb->getUrl()->toString()),
+        //   ]);
+        // }
+        // return $links;
```
ModulesUnraveled, line 62: If we're not using this, it should be deleted.

ModulesUnraveled (changes requested):
This functionally worked for me. There are a handful of change requests in the related Atomic PR though.

web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/SearchBreadcrumbTwigExtension.php
marcb4k: We can remove, but this file is currently a work in progress to get the breadcrumbs for search results. Jim said if there wasn't anything breaking, to leave the work in place and I've added a @todo to it and a ticket in Jira to return to this work at some point.

marcb4k: It was causing more issues in linting, so I removed this altogether.

PR #175, opened 2022-11-23
YALB-825: Update accordion label

nJim:
### Description of work
- Update the accordion heading field label from "Heading" to "Accordion Component Heading"

### Functional testing steps:
No functional changes. Reviewing the YML should be good enough.


marcb4k (approved):
Works for me, approved

PR #177, opened 2022-12-02
YALB-857: Feedback: Section component content duplicates on the page

marcb4k:
### Description of work
- Removes duplicate content when using sections
- Internal notes: This changes the field formatter widget from "Referenced Entity" to "Layout Paragraphs" but does not use the "Layout Paragraphs Builder" as we had previously, so there are still no controls for editing content on the front-end. I have verified that the markup between "Referenced Entity" and "Layout Paragraphs" is the same but the duplicated content is removed with the latter.

### Functional testing steps:
- [x] If not done already, `drush cim` to import config
- [x] Create a page
- [x] Add a section to the page
- [x] Add a component into each of the sections, primary and secondary
- [x] Publish and save the page
- [x] View the page on the front-end and verify that only the section contains the content from the primary and secondary and that the content from those section areas does not appear below the section.

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.page.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.page.default.yml
@@ -12,6 +12,7 @@ dependencies:
     - node.type.page
   module:
     - entity_reference_revisions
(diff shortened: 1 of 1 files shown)


nJim (approved):
Squashed that bug. Ship it.

PR #191, opened 2023-01-10
Yalb 891  views basic formatter

marcb4k:
### Description of work
- Adds a default formatter that displays a dynamic view based on parameters specified in the custom views basic field

### Functional testing steps:
- [ ] If not yet done, import config `drush cim`
- [ ] If there aren't already some news, event, or page nodes, add at least 2 of each. Make sure to add some teaser titles, teaser text, and a teaser image to events and news nodes.
- [ ] Create a new page, title the page, and mark it as published.
- [ ] Add a new paragraph item for a "Basic View"
- [ ] Choose a combination of what content type to display and what view mode to use.
- [ ] Verify that the view mode drop down changes when the content type dropdown is changed.
- [ ] Save the paragraph
- [ ] Save the page and view on the front-end.
- [ ] Verify that the results of the view match the selection on the back-end.
- [ ] Edit the page, change the basic view parameters, save, and verify that the view results change.


web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/Field/FieldFormatter/ViewsBasicDefaultFormatter.php
```diff
@@ -99,26 +113,39 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
+      $running = TRUE;
+
+      // Set up the view and initial decoded parameters.
+      $view = Views::getView('views_basic_scaffold');
+      $view->setDisplay('block_1');
+      $paramsDecoded = json_decode($item->getValue()['params'], TRUE);
+
+      // Overrides filters using our custom views filter - ViewsBasicFilter.
+      $filters = $view->display_handler->getOption('filters');
+      $filters['views_basic_filter']['value'] = $paramsDecoded;
```
nJim, line 131: For now this is exactly what we need. But I think in the next iteration, we should move some of this to the manager service. I don't think the views plugin should worry about decoding the params and then finding the right value. That's all business logic that can exist in the single management class. 

Instead I would have a method for getFilter() or getFitlers() or getFiltersbyType() or whatever we may need. The management class returns a string. It can even do validation and default values if needed. It's likely other views plugins will need similar functions, so might as well centralize them.

No change at the moment. But I think we will add some additional filters next sprint and that will be the time to experiment a little.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/views/filter/ViewsBasicFilter.php
```diff
@@ -0,0 +1,30 @@
...
+   */
+  public function query() {
+    // Ensure the main table for this handler is in the query.
+    $this->ensureMyTable();
+
+    // Parse content type filters.
+    // @todo Add taxonomy term as a filter. Code here: https://gist.github.com/marcb4k/fd4716eb133aa846f90a9e42c80def8c.
+    foreach ($this->value['filters']['types'] as $content_type) {
+      $this->query->addWhere($this->options['group'], 'type', $content_type);
+    }
```
nJim, line 27:
```suggestion
    // Ensure the main table for this handler is in the query.
    $this->ensureMyTable();

    /** @var \Drupal\views\Plugin\views\query\Sql $query */
    $query = $this->query;

    // Parse content type filters.
    // @todo Add taxonomy term as a filter. Code here: https://gist.github.com/marcb4k/fd4716eb133aa846f90a9e42c80def8c.
    foreach ($this->value['filters']['types'] as $content_type) {
      $query->addWhere($this->options['group'], 'type', $content_type);
    }
```

nJim: Just and FYI. If you want to appease the IDE overlords, then this is the common way to do it. The reason `addWhere` is not recognized is that not all view can follow this syntax. I think SOLR data has a different set of methods for querying data. I'm pretty sure external entities from a webservice don't explicitly use a where clause. 

We know that all YaleSites database data will be a flavor of SQL. The IDE is just barking because it technically does not have to be true in Drupal.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
```diff
@@ -16,14 +16,42 @@ class ViewsBasicManager extends ControllerBase implements ContainerInjectionInte
...
+        'card' => 'Event cards',
+        'list_item' => 'Event list items',
+      ],
+    ],
+    'page' => [
+      'label' => 'Pages',
+      'view_modes' => [
+        'teaser' => 'Teasers',
+      ],
+    ],
```
nJim, line 54: Like you, I would love to replace this with something Drupalier. Let's keep on this path for a little to see how ugly things get. This current path gives us a lot of freedom and flexibility. We can create something more structured once we get some more bells and whistles.

nJim (approved):
I left a few comments. One where I suggested a non-functional change. Most of them just reacting and thinking about the future. This looks great and can be merged in. I want to let this get more complicated before we overthink the best way to organize everything.

PR #192, opened 2023-01-12
YALB-772: Nav: Add top level link to fly out

marcb4k:
### Description of work
- If there is a submenu on a main menu link, this will add the top level link as the first item into that submenu.
- If the link is a node, the link text will change from the link title to the node title.

### Functional testing steps:
- [ ] If not already done, add some links into the main menu.
- [ ] Reorder the main menu links to include some that do not have any subitems, and some that do have subitems.
- [ ] Save the menu.
- [ ] On the front-end, verify that opening any main navigation item that has a submenu also has a new first link that going to the same page as the top level link.
- [ ] Verify that the title of this new link is the node title and not the original link title.


web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
```diff
@@ -84,3 +84,27 @@ function ys_core_user_login_form_submit($form, FormStateInterface $form_state) {
...
+      $routeParams = $clonedMenuItem['original_link']->getRouteParameters();
+      if (!empty($routeParams['node'])) {
+        $node = \Drupal::entityTypeManager()->getStorage('node')->load($routeParams['node']);
+        $clonedMenuItem['title'] = $node->getTitle();
+      }
+
+      // Add cloned item to the beginning of the menu.
+      array_unshift($variables['items'][$menuItem]['below'], $clonedMenuItem);
+    }
+  }
```
nJim, line 109:
```suggestion
  // If submenu, add the top level link as the first link in the submenu.
  foreach ($variables['items'] as &$menuItem) {
    // Skip menu links without children.
    if (empty($menuItem['below'])) {
      continue;
    }
    // Clone top level menu item.
    $clonedMenuItem = $menuItem;
    
    // If a node, use the node title instead of link title.
    $routeParams = $clonedMenuItem['original_link']->getRouteParameters();
    if (!empty($routeParams['node'])) {
      /** @var \Drupal\node\Entity\Node $node */
      $node = \Drupal::entityTypeManager()->getStorage('node')->load($routeParams['node']);
      $clonedMenuItem['title'] = $node->getTitle();
    }

    // Add cloned item to the beginning of the menu.
    array_unshift($menuItem['below'], $clonedMenuItem);
  }
```

nJim: Just making a few suggestions:

1. No need to gather the keys and then foreach the keys. You can just foreach and it will iterate the same.
2. If you pass the iterator by reference, you can say `$menuItem` instead of `$variables['items'][$menuItem]`.
3. I tend to avoid wrapping code in nested if statements. Instead, I exit early if something doesn't apply. So if there are no children, just continue to the next one. This helps more when you have lots of nested functions and it's a challenge to trace where you are in the code.

These are all preferences. While my code has some performance gains, it really doesn't matter in the context of Drupal. Adopt whatever parts you find more readable.

nJim: We could even remove the clone piece by simply overriding the title after adding it to the array. But I think having a variable named 'cloneMenuItem' creates some solid self documenting code.

web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
```diff
@@ -99,3 +99,35 @@ function ys_core_user_login_form_submit($form, FormStateInterface $form_state) {
...
+    // Skip menu items that do not have subitems.
+    if (empty($menuItem['below'])) {
+      continue;
+    }
+
+    // Clone top level menu item.
+    $clonedMenuItem = $menuItem;
+
+    // We don't want any other subitems, just the top level.
+    $clonedMenuItem['below'] = NULL;
```
marcb4k, line 119: Good that we re-reviewed this. There was a small bug where it was cloning ALL subitems into the menu. Since we only want the top level item, I set all the below items to `NULL`.

nJim: Smart

nJim (approved):
Approved.

PR #198, opened 2023-01-19
Bug: roles are removed from users if not part of Grouper group

vinmassaro:
### Description of work
CAS users who have been given a role but are not in the Yalesites Grouper group are stripped of that role on their next login because of the CAS Attributes setting "Remove role from user if match fails?". This changes the value to false to stop this from happening.


nJim (approved):
Approved. While I did not test this with a real user this settings change seems straight forward and would fix the issue you described.

PR #199, opened 2023-01-20
YALB-937: Views: Filter by tag

marcb4k:
### Description of work
- Adds the ability to filter views basic lists by a taxonomy term in the tags vocabulary
- At the moment only one term is supported and help text is added to inform the content editor
- There is some work in Drupal core over single value autocomplete widgets not warning about cardinality: https://www.drupal.org/project/drupal/issues/2951134 so for the moment, if there are multiple values, the last value is used.

### Functional testing steps:
- [ ] Create a few News nodes and tag them with some tags. Make sure to reuse a tag or two for testing the view.
- [ ] Create a page and add a Views Basic paragraph to it
- [ ] Select "Display News articles as News cards"
- [ ] Start typing the name of one of the tags used above for autocomplete and select that tag
- [ ] Save the paragraph and verify that the preview adds the "Filtered by tag "{tagname}"
- [ ] Save the page
- [ ] Verify that the view displays only the news nodes tagged with the tag selected.

### DIFF
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -89,6 +89,9 @@
       },
       "drupal/core": {
         "plural results summary https://www.drupal.org/project/drupal/issues/2888320": "https://www.drupal.org/files/issues/2021-12-15/2888320-78.patch"
+      },
+      "drupal/hide_revision_field": {
+        "fix deprecated function unserialize https://www.drupal.org/project/hide_revision_field/issues/3260101": "https://www.drupal.org/files/issues/2022-01-24/3260101-2.patch"
       }
     }
   }
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/Field/FieldWidget/ViewsBasicDefaultWidget.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/Field/FieldWidget/ViewsBasicDefaultWidget.php
@@ -112,6 +112,11 @@ class ViewsBasicDefaultWidget extends WidgetBase implements ContainerFactoryPlug
       ],
     ];
 
+    $form['group_user_selection']['markup_pre_entity_types'] = [
+      '#type' => 'markup',
+      '#markup' => '<div class="grouped-items">',
+    ];
+
     $form['group_user_selection']['entity_types'] = [
       '#type' => 'select',
       '#options' => $this->viewsBasicManager->entityTypeList(),
@@ -156,6 +161,24 @@ class ViewsBasicDefaultWidget extends WidgetBase implements ContainerFactoryPlug
       '#suffix' => '</div>',
     ];
 
+    $form['group_user_selection']['markup_post_entity_types'] = [
+      '#type' => 'markup',
+      '#markup' => '</div>',
+    ];
+
+    // @todo add validation for only one term.
+    // More info: https://www.drupal.org/project/drupal/issues/2951134
+    $form['group_user_selection']['tags'] = [
+      '#title' => $this->t('Filtered by tag'),
+      '#description' => $this->t('At this time, only one term is supported. If multiple terms are added, only the last one will be used.'),
+      '#type' => 'entity_autocomplete',
+      '#target_type' => 'taxonomy_term',
+      '#default_value' => ($items[$delta]->params) ? $this->viewsBasicManager->getDefaultParamValue('tags', $items[$delta]->params) : NULL,
+      '#selection_settings' => [
+        'target_bundles' => ['tags'],
+      ],
+    ];
+
     $element['group_params']['params'] = [
       '#type' => 'textarea',
       '#title' => $this->t('Params'),
@@ -183,6 +206,9 @@ class ViewsBasicDefaultWidget extends WidgetBase implements ContainerFactoryPlug
         "filters" => [
           "types" => [
             $form['group_user_selection']['entity_types']['#value'],
+          ],
+          "tags" => [
+            $form_state->getValue(['group_user_selection', 'tags']),
           ],
         ],
       ];
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/views/filter/ViewsBasicFilter.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/views/filter/ViewsBasicFilter.php
@@ -29,9 +29,11 @@ class ViewsBasicFilter extends FilterPluginBase {
     }
 
     // Parse tag filters.
-    $tag_table = $query->addTable('node__field_tags');
-    foreach ($this->value['filters']['tags'] as $tid) {
-      $query->addWhere(0, "$tag_table.field_tags_target_id", $tid);
+    if (($this->value['filters']['tags'][0])) {
+      $tag_table = $query->addTable('node__field_tags');
+      foreach ($this->value['filters']['tags'] as $tid) {
+        $query->addWhere(0, "$tag_table.field_tags_target_id", $tid);
+      }
     }
   }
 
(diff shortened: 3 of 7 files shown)


web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/Field/FieldWidget/ViewsBasicDefaultWidget.php
```diff
@@ -156,6 +161,24 @@ public function formElement(
       '#suffix' => '</div>',
     ];
 
+    $form['group_user_selection']['markup_post_entity_types'] = [
+      '#type' => 'markup',
+      '#markup' => '</div>',
+    ];
+
+    // @todo add validation for only one term.
```
nJim, line 169: Let's leave this for now as I want to keep charing forward. But if this becomes and issue we can change this to a text input and create our own autocomplete callback. The entity_autocomplete type is doing all of the work for us, but it sucks that it does not have enough options.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/views/filter/ViewsBasicFilter.php
```diff
@@ -24,10 +24,15 @@ public function query() {
     $query = $this->query;
 
     // Parse content type filters.
-    // @todo Add taxonomy term as a filter. Code here: https://gist.github.com/marcb4k/fd4716eb133aa846f90a9e42c80def8c.
     foreach ($this->value['filters']['types'] as $content_type) {
       $query->addWhere($this->options['group'], 'type', $content_type);
     }
+
+    // Parse tag filters.
+    $tag_table = $query->addTable('node__field_tags');
```
nJim, line 32: WHAT? How did I not know `addTable` exists? I've been writing joins for years like a chump.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
```diff
@@ -108,6 +108,26 @@ public function viewModeList($content_type) {
     return $viewModes;
   }
 
+  /**
+   * Returns an array of taxonomy tags.
+   */
+  public function tagsList() {
```
nJim, line 114: I don't think this method is in use. Should we delete it? or do you have something planned for it?

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
```diff
@@ -122,11 +142,20 @@ public function getViewModeLabel($type, $view_mode) {
     return self::ALLOWED_ENTITIES[$type]['view_modes'][$view_mode];
   }
 
+  /**
+   * Returns a tag label given a term ID.
+   */
+  public function getTagLabel($tag) {
+    $term = $this->entityTypeManager()->getStorage('taxonomy_term')->load($tag);
+    return $term->name->value;
```
nJim, line 150: $term may be null if the term does not exist. For example, if someone deletes the taxonomy term. Then I think these node forms will white-screen when trying to generate previews.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
```diff
@@ -136,6 +165,11 @@ public function getDefaultParamValue($type, $params) {
         $defaultParam = $paramsDecoded['filters']['types'][0];
         break;
 
+      case 'tags':
+        $tid = (int) $paramsDecoded['filters']['tags'][0];
```
nJim, line 169: I have a feeling that this method will change a lot in the near future as we make the data more complex. So I don't want to overthinking it. 

The only think we may want to change right now is avoiding warnings around `$paramsDecoded['filters']['types'][0]`. It is possible that these keys will not exist. (Okay it is not really possible since this tool is not live, but I'm approaching this from a future-proof point of view.) So we may want to check if !empty($paramsDecoded['filters']['types'][0]) to avoid "illegal offset" and "Undefined array key" warnings.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
```diff
@@ -122,11 +142,20 @@ public function getViewModeLabel($type, $view_mode) {
     return self::ALLOWED_ENTITIES[$type]['view_modes'][$view_mode];
   }
 
+  /**
+   * Returns a tag label given a term ID.
+   */
+  public function getTagLabel($tag) {
```
nJim, line 148: I didn't think of this before, but would you mind typehinting functions? For example:
```
/**
  * Returns a tag label given a term ID.
  * 
  * @param int $tag
  *   The ID of a taxonomy term.
  * 
  * @return string
  *   The label of the given taxonomy term or an empty string.
  */
  public function getTagLabel(int $tag): string {
```

web/profiles/custom/yalesites_profile/composer.json
```diff
@@ -89,6 +89,9 @@
       },
       "drupal/core": {
         "plural results summary https://www.drupal.org/project/drupal/issues/2888320": "https://www.drupal.org/files/issues/2021-12-15/2888320-78.patch"
+      },
+      "drupal/hide_revision_field": {
+        "fix deprecated function unserialize https://www.drupal.org/project/hide_revision_field/issues/3260101": "https://www.drupal.org/files/issues/2022-01-24/3260101-2.patch"
```
nJim, line 94: Is this related?

nJim (changes requested):
Everything works great. I added some commentary to parts of the code. A few changes requested that are mostly from an opinionated place.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/Field/FieldWidget/ViewsBasicDefaultWidget.php
marcb4k: Yah, from that discussion on d.o, it seems that work is progressing. We could try applying that patch if you'd like?

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/views/filter/ViewsBasicFilter.php
marcb4k: It had been a while since I've done queries like that, so I looked it up and saw that - pretty neat! I recall doing joins before as well.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
marcb4k: Good catch - before I was using the `entity_autocomplete`, I was going to write my own. Removed.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
marcb4k: Good call. Fixed.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
marcb4k: Yup, thanks for the reminder. Added for all functions.

web/profiles/custom/yalesites_profile/composer.json
marcb4k: Not to this work, but I received an error about this because of our upgraded PHP version. I can remove this and add to a different PR if you'd like?

nJim: Nope it's good. Just wan't sure.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/Field/FieldWidget/ViewsBasicDefaultWidget.php
nJim: It's good for now. We just want something stable enough to get client feedback. We'll just keep things moving for now.

nJim (approved):
Approving change. Thanks for all the little and big things in this one.

PR #202, opened 2023-01-25
YALB-938: Views: Numeric limit

marcb4k:
### Description of work
- Adds numeric limit to views basic
- Defaults to 0 to show all items (This can be changed if we want to set it to a default of 10, for example)
- Fixes an issue with filtering by tags - if tags are empty, show all.

### Functional testing steps:
- [ ] Import config `drush cim`
- [ ] If not already done, add about 3-4 news articles with teaser titles, text, and an image
- [ ] Add a page and on that page, add a paragraph of "Basic View"
- [ ] Specify to show a list of News Articles and leave the default "0" for the "Limit to"
- [ ] Save the page and verify that the number of articles matches what exists in the system
- [ ] Edit the page and the Basic View and change the limit to 1 or 2
- [ ] Save the page and verify that the number of articles shown matches the limit specified


web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/Field/FieldWidget/ViewsBasicDefaultWidget.php
```diff
@@ -179,6 +179,26 @@ public function formElement(
...
+      '#default_value' => ($items[$delta]->params) ? $this->viewsBasicManager->getDefaultParamValue('limit', $items[$delta]->params) : 0,
+      '#min' => 0,
+      '#required' => TRUE,
+    ];
+
+    $form['group_user_selection']['markup_post_limit'] = [
+      '#type' => 'markup',
+      '#markup' => '</div>',
+    ];
+
```
nJim, line 201: I never see this pattern in Drupal where HTML is started in one place and ended in another. (Talking about the wrapper div). Although it is extremely common in WordPress FWIW.

I suggest either using #prefix and #suffix on the limit item:
```
    $form['group_user_selection']['limit'] = [
      '#title' => $this->t('Limit to'),
      '#description' => $this->t('Enter 0 to show all items.'),
      '#prefix' => '<div class="grouped-items">',
      '#suffix' => $this->t('item(s)') . '</div>',
      '#type' => 'number',
      '#default_value' => ($items[$delta]->params) ? $this->viewsBasicManager->getDefaultParamValue('limit', $items[$delta]->params) : 0,
      '#min' => 0,
      '#required' => TRUE,
    ];
```
Or using a wrapping element such as a fieldset or a container:
```
    $form['group_user_selection']['limit_group'] = [
      '#type' => 'container',
      '#attributes' => [
        'class' => ['grouped-items'],
      ],
    ];
    $form['group_user_selection']['limit_group']['limit'] = [
      '#title' => $this->t('Limit to'),
      '#description' => $this->t('Enter 0 to show all items.'),
      '#suffix' => $this->t('item(s)'),
      '#type' => 'number',
      '#default_value' => ($items[$delta]->params) ? $this->viewsBasicManager->getDefaultParamValue('limit', $items[$delta]->params) : 0,
      '#min' => 0,
      '#required' => TRUE,
    ];
```

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/views/filter/ViewsBasicFilter.php
```diff
@@ -29,9 +29,11 @@ public function query() {
     }
 
     // Parse tag filters.
-    $tag_table = $query->addTable('node__field_tags');
-    foreach ($this->value['filters']['tags'] as $tid) {
-      $query->addWhere(0, "$tag_table.field_tags_target_id", $tid);
+    if (($this->value['filters']['tags'][0])) {
```
nJim, line 32: Checking if($var) will prevent the block of code from running but will also result in an error if the variable is not set. I suggest if(!empty($var)).

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
```diff
@@ -166,10 +166,18 @@ public function getDefaultParamValue($type, $params) {
         break;
 
       case 'tags':
+        if (!isset($paramsDecoded['filters']['tags'])) {
```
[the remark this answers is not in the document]
nJim: If change it to empty and it will catch that case as well `empty($paramsDecoded['filters']['tags'])`

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
```diff
@@ -166,10 +166,18 @@ public function getDefaultParamValue($type, $params) {
...
+          $defaultParam = NULL;
+          break;
+        }
         $tid = (int) $paramsDecoded['filters']['tags'][0];
         $defaultParam = $this->entityTypeManager()->getStorage('taxonomy_term')->load($tid);
         break;
 
+      case 'limit':
+        $defaultParam = (empty($paramsDecoded['limit'])) ? 0 : $paramsDecoded['limit'];
```
nJim, line 178: This works, but I also think you can case it to an int and all cases will be covered:
`$defaultParam = (int) $paramsDecoded['limit'];`

nJim (changes requested):
Works great. A few pedantic changes requested.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/Field/FieldWidget/ViewsBasicDefaultWidget.php
marcb4k: Fixed. I actually removed all of that and essentially copied what Views does, so no suffix.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/views/filter/ViewsBasicFilter.php
marcb4k: Good catch. Fixed.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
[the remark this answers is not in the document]
marcb4k: Fixed in prior PR.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
marcb4k: Fixed.

nJim (approved):
Approved.

PR #209, opened 2023-02-01
YALB-978: Clear and Informative UI Epic

nJim:
### Description of work
- Wrapped Image caption now had maxlength setting of 120
- Wrapped Image text field is now required
- Icons and descriptions added to multiple components
- Components renamed

### Functional testing steps:
- [ ] Login to the multidev for testig
- [ ] Create a new page and click the plus (+) button to open the new paragraph picker
- [ ] Verify the following configuration were change:
  - [ ] "Pull Quote" is renamed to "Quote"
  - [ ] "Basic View" is renamed to "View"
  - [ ] Icons are added to Grand Hero, View, Calendar List, and Newsfeed
  - [ ] Descriptions area added to View, Calendar List, and Newsfeed and can viewed on the paragraph edit form.
  - [ ] The wrapped image content field is now required
  - [ ] The wrapped image caption field uses a soft limit of 120 chars


miketullo95:
These look great! One thing, for the soft cap character limit on the wrapped image caption, we probably want to have some visual indication for soft caps. We haven't used that mechanic yet and at the moment it looks similar (if not the same?) to the hard cap. Would that be included here? or a different ticket/PR?    @nJim

nJim:
@miketullo95 There are a couple of tickets around validation and char limits. So let's continue that work in a different ticket. Good points. I'm curious if we are going to change direction and/or pick a different tool to give the user better feedback. See tickets YALB-895 and YALB-961.

PR #214, opened 2023-02-04
YALB-651: CAS-restricted nodes (new)

marcb4k:
### Description of work
- Adds a field to pages, events, and news to require a CAS login
- Adds a custom block that will show up on 403 pages (default or custom) that has a link to login via CAS including a destination parameter to return the user back to the original page.
- Adds custom code to set Drupal grants.

### Functional testing steps:
- [ ] Import config `drush cim`
- [x] Edit an existing node or create a new node
- [x] In the sidebar, open the "Publish Information" details and check "CAS Login Required"
- [x] View the node while logged in and verify that it is accessible.
- [x] View the same node on a incognito / private (logged out) browser window
- [ ] Verify that same node is not accessible and that the 403 page has a link to login via CAS
- [ ] Click the link and login via CAS
- [ ] After CAS login, verify that you are taken back to the originating page


vinmassaro:
@marcb4k this is looking much more like I envisioned, thanks.

Can it be simplified further by removing block_in_page_403 and using methods directly from the CAS module to force login when accessing a protected node? I think it's OK to remove this extra step where you hit a 403 and the link to login with CAS is displayed, especially if we plan to style links to CAS protected nodes with a padlock or other icon. The CAS module implements this forced login behavior directly where you can supply page paths in the CAS module config. We aren't using it because we don't want to give users access to all of the CAS configurations, but we should mimic this behavior. Thanks!

marcb4k:
@vinmassaro - No problem, I can make that change. My thought was that we allow them to supply a custom 403 page in the settings, and if we bypassed that with CAS login pages, then no one would ever see that 403 page. But I can see that if it's set to CAS enabled with that field and we have links styled correctly, then the 403 would still be shown to standard Drupal access denied (like `/admin` for example). Cool, I'll change this up and re-submit for review.

marcb4k:
@vinmassaro - OK, fixed the dependency injection listing and is ready for your re-review. CAS protected nodes will now redirect directly to the CAS login now without an interstitial page.

rigoucr:
@marcb4k is there a Test user for the CAS login?

vinmassaro (approved):

web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/src/EventSubscriber/NodeAccessEventSubscriber.php
```diff
@@ -0,0 +1,97 @@
...
+      // Make sure this is a node.
+      if ($entityType == 'node') {
+        /** @var \Drupal\node\Entity $node */
+        $node = $this->entityTypeManager->getStorage($entityType)->load($params[$entityType]);
+
+        /* Check to see if node is set to require login only.
+         * If so, redirect to CAS login.
+         */
+        if ($node->hasField('field_login_required')) {
+          if ($node->get('field_login_required')->getValue()[0]['value']) {
```
rigoucr, line 87: Maybe here you could use:` $node->field_login_required->isEmpty() ` to validate if there is data into the field

rigoucr (changes requested):
A couple of possible improvements here @marcb4k

web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/src/EventSubscriber/NodeAccessEventSubscriber.php
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * Manage the 403 exceptions.
+   *
+   * @param Symfony\Component\HttpKernel\Event\ExceptionEvent $event
+   *   The Incoming event.
+   */
+  public function on403(ExceptionEvent $event) {
+    if ($this->currentUser->isAnonymous()) {
+      // Get path alias.
+      $path = $event->getRequest()->getPathInfo();
```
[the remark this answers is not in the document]
marcb4k: So much better! Refactored! TY!

web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/src/EventSubscriber/NodeAccessEventSubscriber.php
marcb4k: Great idea - since this is only a boolean, we just need to check if it's empty or not. Refactored.

web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/src/EventSubscriber/NodeAccessEventSubscriber.php
```diff
@@ -0,0 +1,87 @@
...
+   * @var \Drupal\path_alias\AliasManager
+   */
+  protected $pathAlias;
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManager
+   */
+  protected $entityTypeManager;
```
rigoucr, line 36: @marcb4k you forgot to remove the entityTypemanger  manager definition

web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/src/EventSubscriber/NodeAccessEventSubscriber.php
```diff
@@ -0,0 +1,87 @@
...
+   * The construct for this class.
+   *
+   * @param Drupal\Core\Session\AccountInterface $current_user
+   *   The current user service.
+   * @param Drupal\path_alias\AliasManager $path_alias
+   *   The path alias manager.
+   */
+  public function __construct(AccountInterface $current_user, AliasManager $path_alias) {
+    $this->currentUser = $current_user;
+    $this->pathAlias = $path_alias;
```
rigoucr, line 48: Also, reading this again, noticed that the AliasManager services could be removed, you are not currently using it in your current logic

web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/src/EventSubscriber/NodeAccessEventSubscriber.php
```diff
@@ -0,0 +1,87 @@
...
+   * Manage the 403 exceptions.
+   *
+   * @param Symfony\Component\HttpKernel\Event\ExceptionEvent $event
+   *   The Incoming event.
+   */
+  public function on403(ExceptionEvent $event) {
+    if ($this->currentUser->isAnonymous()) {
+
+      // Get path alias.
+      $path = $event->getRequest()->getPathInfo();
```
rigoucr, line 68: Also this could be removed and in the line 78 , you could could simply call to the toUrl() function of Node class to get the current node url, example:
`$casRedirectUrl = Url::fromRoute('cas.login', ['destination' => $node->toUrl()->toString()])->toString();`

rigoucr (approved):
Almost there , just a couple of details

web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/src/EventSubscriber/NodeAccessEventSubscriber.php
marcb4k: Removed.

web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/src/EventSubscriber/NodeAccessEventSubscriber.php
marcb4k: Yup, good call. Changed.

marcb4k:
@vinmassaro - Good idea. I'll put in a ticket for that change on moving items around in the sidebar. I think that would be a good one for the UX folks to look at first to talk about naming the sections so it makes sense. I like your suggestion of "visibility".

web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/src/EventSubscriber/NodeAccessEventSubscriber.php
marcb4k: Removed, thanks!

rigoucr (approved):
The code looks good and it works as expected ,thx

PR #217, opened 2023-02-06
YALB-879: AX: Alerts management interface fe

mariannuar:
### Description of work
- Display `Alert Types` radio buttons as illustrations.
- The description of the alert type appears in a screenreader friendly way.
- The `Are you sure` form includes custom styles to show the importance of this action.

### Functional testing steps:
- [ ] Login to the multidev via [CAS login](https://pr-217-yalesites-platform.pantheonsite.io/cas) or `terminus drush yalesites-platform.pr-217 -- uli`
- [ ] Navigate to the [Alerts Settings interface](http://pr-217-yalesites-platform.pantheonsite.io/admin/yalesites/alert)
- [ ] Verify the 3 alert types are displayed as illustrations
- [ ] Verify the form looks similar to the [design](https://www.figma.com/file/Goj25us0SUbg0qw2xY8uyF/Yalesites-Alert?node-id=0%3A1&t=fnOPRmitIqgnibxu-0)
- [ ] Select the `Emergency` and verify the `Are you sure` modal form looks similar to the design
- [ ] Enable `Voice Over` or any screenreader
- [ ] Navigate through the `Alert Types` and confirm the screean reader reads the description every time a new option is selected.


nJim (approved):
This is exactly what we needed. Thank you for digging into the form to make changes. This work can get messy to divide between frontend and backend. Your solution was perfect. Everything is well organized including the new method in the AlertManager.

All of this work is approved. I'm going to tag Yale's accessibility folks to give it a final review.

ntay90 (changes requested):
This looks good; accessibility-wise, the only thing that I'm noticing is that the alert radio buttons are missing a focus indicator when they receive keyboard focus. Additionally, the "Are you sure?" dialog is missing an `aria-modal="true"` attribute.

- [ ] Add focus indicator to alert radio buttons.
- [ ] Add `aria-modal="true"` to "Are you sure?" alert dialog.

nJim:
@ntay90 We got the focus state added. The other issue you mentioned is adding aria-modal="true" to the modals. There is not an easy way for us to do this as we are using Drupal's core Modal handler. So this issue is present for all modals (media library, paragraph picker, embed content form, some of the edit dialogs.) 
 
I've documented this in a follow up ticket so we can decide the best path to improve this code. Ideally, we would provide a patch to the Core Drupal project so that everyone can get the improved version of the modal. https://yaleits.atlassian.net/browse/YALB-1000
 
I hope it is okay to move this work forward as is. If you consider this a blocker I could always remove the modal part of the settings form until we can work on the aria attribute. Thoughts?

ntay90:
@nJim I think the follow-up ticket for the modal is reasonable, and the focus state looks great. I think this is good to go.

PR #220, opened 2023-02-14
YALB-651: Bugfix for CAS-restricted nodes

marcb4k:
### Description of work
- Fixes caching issue when reverting a CAS required page to not require CAS login, cache will no longer redirect viewing of node

### Functional testing steps:
- [x] Edit a page, news, or event
- [x] In the sidebar under "Publish Options", select "CAS Login Required" and then save the node.
- [x] View the node logged out (incognito window) and verify that you get redirected to the CAS login page
- [x] Edit the node and uncheck the "CAS Login Required" and save the node.
- [x] View the node logged out (incognito window) and verify that you can see the page fine, with no need to clear the cache.


rigoucr (approved):
Works like a charm

PR #221, opened 2023-02-16
YALB-1010: Views Sorting

marcb4k:
### Description of work
- Adds sorting ability to the Views Basic interface.
- For now, the options are: Published Date (News), Event Date (Events), and Titles (Pages) with both ASC and DESC orders for each. Each sort label, field, and orders can and will be changed after Yale feedback.

### Additional contextual notes
Views Basic allows content editors to select some criteria to show lists of other content as you would normally do via the Views interface, but with a more controlled and limited interface. For now, the content type of "Pages" is a placeholder as we intend to replace pages with people profiles. Therefore, the Pages sort is an example of how we can change the context of the sort from date-based sorts (for News and Events) to text-based sorts.

### Functional testing steps:
- [x] Add three published News nodes to the site. In the sidebar, open "Publish Information" and change the publish date for each to a different date. Title the node with the date used. (This is because at the moment, the "Publish Date" field is not shown on the front-end template)
- [x] Add 3 published Event nodes to the site. Under the "Logistics" tab, make sure to add a date and time for the event and make all three different dates. (Events do show the event date correctly on the front-end)
- [x] Create a Page and add a paragraph of type "View" to that page.
- [x] In the "Create new View" modal, select either the "Posts" or "Events" at the top, and first verify that the "Sorting by" field changes to match the content type selected. Also verify that the view mode list (to the right of the content type) also changes when the content type is changed.
- [x] Select one of the "Sorting by" options and then save the view.
- [x] Publish and save the page and verify that the list of content in the view are in the correct order on the front-end as per the "Sorting by" option selected.
- [x] Go back and edit the page and change the content type and/or sort order and verify that the correct content type sort order is shown on the front-end.

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/views.view.views_basic_scaffold.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/views.view.views_basic_scaffold.yml
@@ -11,7 +11,7 @@ dependencies:
 id: views_basic_scaffold
 label: 'Views Basic Scaffold'
 module: views
-description: ''
+description: 'View that is overridden by Views Basic custom module code'
 tag: ''
 base_table: node_field_data
 base_field: nid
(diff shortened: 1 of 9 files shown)


rigoucr (approved):
Code looks good and it works as expected

PR #227, opened 2023-02-24
YALB-85: Create profile content type

nJim:
### Description of work
- Create the profile content type as defined in [the data model](https://yaleedu.sharepoint.com/:x:/r/sites/YaleSitesUpgradeProgram/_layouts/15/Doc.aspx?sourcedoc=%7B2B574C06-BAC3-41E2-924A-F4562054F849%7D&file=Proposed%20Content%20Model.xlsx&action=default&mobileredirect=true&wdOrigin=TEAMS-ELECTRON.teams.files&wdExp=TEAMS-CONTROL&wdhostclicktime=1644421397192&cid=ec49a607-8b17-4aad-82cb-ab5eb45ecb7a).
- Sets all content types to be published by default
- Updates text with image to use the new name "Content Spotlight"
- Adds an icon and instructions to the webform component

### Testing
This work is not yet in a testable place and is only intended to lay groundwork for our next sprint where we integrate profiles with the directory service. There is also no frontend work associated with this content type. As such, the content type is hidden from users. A screenshot is added to this PR so reviewers can ask questions and verify that we are going in the right direction.
local yale edu_node_add_profile (1)


nJim:
Assigning to Alyssa for review via Teams.

rigoucr:
I took a look over the PR and its functionality  and it looks fine but since I can't access the data model link I wasn't able to compare the new Content type against it.

And just double checking :
When you say: `the content type is hidden from users`, does it means that you simply are not adding permissions To any role aside of the admin Role ?

PR #233, opened 2023-03-14
YALB-1016: DevOps: Post Deploy Commands

vinmassaro:
### Description of work
This ticket moves the existing confim.php script into `/web/private/scripts/` since Pantheon platform can only run them from this directory. It also adds the `sync_code` hook so that it will run after code sync, upstream updates, CI pushes to multidevs, etc., and updates paths in pantheon.yml.


nJim (approved):
Code looks good. Thank you for debugging this. I will merge the changes since there are some releases going out shortly. I will verify that this runs on our sandbox sites.

PR #234, opened 2023-03-16
Layout builder blocks

nJim:
### Description of work
- Creates new custom blocks by recreating the configuration for the previous paragraphs. This includes: embed, video, image, and webform components
- Templates are also recreated in this atomic PR https://github.com/yalesites-org/atomic/pull/101

### Functional testing steps:
- [x] Create a new page for testing and navigate to the layout interface for that node
- [x] Testing the 'image' component
  - [x] Verify the component exists within the layout builder tools in an appropriate group
  - [x] Confirm that the fields match their previous definitions from the related paragraph.
  - [x] Check that instructions are added to the widget
  - [x] There is no preview image at this time so ignore the generic thumbnail
  - [x] Add an example component to the page
  - [ ] Make sure the component renders through the correct template to appear like the previous paragraph based iteration.
- [x] Repeat this process for the other component (embed, video, and webform components)


rigoucr:
Hey @nJim, 

- I found that For the webform component in its paragraph version the title was not required , but for the block it is required, is that right ?
Screenshot 2023-03-16 at 2 19 44 PM

- In your instructions you added this instructions ` Make sure the component renders through the correct template to appear like the previous paragraph based iteration.` but  I look into your code and there were only changes related to config nothing related to templates definition , I Asume that there is some magic happening here , because at least to me everything looks great to me, I'm attaching a screen recording of the components I created .

https://user-images.githubusercontent.com/2217827/225746185-696f3259-8c8b-42a6-9d3c-ccb7bafdef76.mov


- Finally I noticed that the Default title for each of the components stills showing when you are creating a block , so as suggestion, I think you should Use this module https://www.drupal.org/project/auto_entitylabel , to automatically populate the block titles , the module also hide the titles from the user forms .

let me know what you think .

marcb4k (changes requested):
Overall it looks good!

@rigoucr mentioned the webform component title not being required on the paragraph but is required on the block.

Also, I saw that on the webform on the form display, the component title is on top for the paragraph but on the block it is on the bottom.

Finally, the templates have changed recently (as per a discussion I had with Emulsify folks on 3/16). See Notion doc for changes. No changes for this repo, but Atomic needs changes and a new branch / PR in the component library as there are some minor changes to those components as well. We can discuss if you'd like or I'm happy to make those changes too, just let me know.

marcb4k:
@rigoucr - I played with the auto entity label module yesterday to see if that would work for us and unfortunately, it's not ready for primetime yet. 1) it requires a patch (no biggie) but 2) I tested the patch and it doesn't seem to work. See here for more: https://www.drupal.org/project/auto_entitylabel/issues/3065687#comment-14971241. If I have time, I'd love to dig into that code and possibly update the patch to fix the issue because it does seem this would be a great way to auto-label some of those components.

rigoucr:
@marcb4k if that module is not ready to prime time you could also try something like this:
`/**
 * Implements hook_form_alter().
 */
function manati_layout_builder_form_alter(array &$form, FormStateInterface $form_state, string $form_id) {
  // Removing the block name field from all the block_content forms inside
  // Layout builder.
  $form_ids = ['layout_builder_update_block', 'layout_builder_add_block'];
  if (in_array($form_id, $form_ids)) {
    // Removig not required fields for block form inside Layout Builder.
    if (isset($form['settings']['block_form'])) {
      // Getting current time to autogenerate a name for the block.
      $current_time = \Drupal::service('datetime.time')->getCurrentTime();
      $date_formatted = \Drupal::service('date.formatter')->format($current_time, 'short');
      // Altering Block name field - removing it from the form.
      $form['settings']['label']['#access'] = FALSE;
      $form['settings']['label']['#default_value'] = $form['settings']['block_form']['#block']->bundle() . ' ' . $date_formatted;
      // Also removing option to show the Block name in its display.
      $form['settings']['label_display']['#default_value'] = FALSE;
      $form['settings']['label_display']['#access'] = FALSE;
      if (!empty($form['settings']['view_mode']['#access'])) {
        unset($form['settings']['view_mode']['#options']['default']);
        $form['settings']['view_mode']['#title'] = t('Presentación');
        $form['settings']['view_mode']['#required'] = TRUE;
        $form['settings']['admin_label']['#weight'] = 0;
        $form['settings']['view_mode']['#weight'] = 0;
      }
    }
   }
    `
Our approach was to hide those fields with code and also set the value we required for our usage case

nJim:
Thank you for the careful review and all of the feedback. I am going to merge this work into the feature branch as we have time work resolve these issues and it will be easier to keep everything on one branch this week. But I will address these things.

PR #238, opened 2023-04-03
YALB-1072: Dedicated edit/layout toolbar links

nJim:
### Description of work
- Organizes existing and new toolbar code in the ys_toolbar module.
- Enables the new module via configuration.
- Adds links to the toolbar though a hook and custom service.
- Edit, Layout, and Publish links added based on business logic.

### Functional testing steps:
- [ ] Create a new unpublished page and view the page.
  - [ ] Verify that an 'Edit' link appears in the toolbar. 
  - [ ] Clicking this link takes the users to the nodes edit form.
- [ ] View the node again.
  - [ ] Verify that a 'Layout' link appears in the toolbar. 
  - [ ] Clicking this link takes the users to the layout builder interface.
- [ ] View the node again.
  - [ ] Verify that a 'Publish' link appears in the toolbar. 
  - [ ] Clicking this link reloads the page. The node is now published and the 'Publish' link no longer appears.
- [ ] All tasks appear in the toolbar in a 'Manage this Page' tray
  - [ ] Clone, Revisions, and many other tasks appear in this space for nodes.
  - [ ] This tray reverts to it's default label "Local Tasks" when viewing a route that is not a node.
  - [ ] It is possible that the "Local" tasks tray is empty for some routes. This is not an clean fix for this bug but we will discuss it in another ticket.


marcb4k (changes requested):
Most of this looks good. I'm seeing the edit and layout buttons show up on the front-end. However a few issues:
1. With the `entity_redirect` module, after adding a page, it does redirect, but to `/{page-title}/layout` but should be `/node/{nid}/layout` for that to work. Otherwise it gives an error: "The requested page could not be found."
2. I am not seeing the "Publish" button in the toolbar when the node is unpublished. I checked the code and see you are checking on the access of the page which I did a quick debug and got `FALSE` for access, even though I checked the permissions page and I should be able to view my own unpublished nodes. I tried both from the admin user as well as my NetID user.

nJim:
> 1. With the `entity_redirect` module, after adding a page, it does redirect, but to `/{page-title}/layout` but should be `/node/{nid}/layout` for that to work. Otherwise it gives an error: "The requested page could not be found."

Good find. Somehow this alluded me if then it was clearly broken. I've created a patch to fix the module. https://www.drupal.org/project/entity_redirect/issues/3352265

> 2. I am not seeing the "Publish" button in the toolbar when the node is unpublished...

Looks like I forgot to run config export. I'm leveraging the existing 'publishcontent' module to create these route. I exported the updated yml and all should work now.

marcb4k (approved):
This works well for me now! I did see a small linting error (inline comment I think) but otherwise, approved!

PR #239, opened 2023-04-04
YALB-1096: Use Simple Sitemap instead of XML Sitemap

vinmassaro:
### Description of work
- Uninstalls XML Sitemap module
- Installs Simple Sitemap module
- Configures event, page, news, and main menu navigation in the default sitemap.xml
- Adds a separate main menu only sitemap for VRT usage
- Unrelated to this ticket: Adds "develop" as the default branch in the CI frontend build script when a matching branch of the component library is not found to fix broken builds

### Functional testing steps:
- [x] Visit https://pr-239-yalesites-platform.pantheonsite.io/sitemap.xml for default sitemap
- [x] Visit https://pr-239-yalesites-platform.pantheonsite.io/menu-links-only/sitemap.xml for main menu links sitemap for VRT use


vinmassaro:
Created followup task [YALB-1097](https://yaleits.atlassian.net/browse/YALB-1097) to remove XML Sitemap after the next release.

web/profiles/custom/yalesites_profile/composer.json
marcb4k, line 1: Would you also want to remove `xmlsitemap` from composer.json since we're switching to `simple_sitemap`?

marcb4k (approved):
This all looks good to me! One comment for a possible future ticket - since we have that "Requires CAS login" option for nodes, would you want those nodes to be excluded from the sitemap? Some custom code for that, but `simple_sitemap` has some hooks that we could use for that.

web/profiles/custom/yalesites_profile/composer.json
vinmassaro: Thanks Marc, we'll remove XML Sitemap in YALB-1097 in the next release so it can be uninstalled and then removed from code base later.

vinmassaro: Forgot to address your first question: yes, we should prevent CAS-protected nodes from appearing in sitemaps. I opened up [YALB-1103](https://yaleits.atlassian.net/browse/YALB-1103) for that. Thanks!

PR #249, opened 2023-04-27
YALB-1113: Enable modal layout browser

nJim:
### Description of work
- Updates a layout browser setting to use the modal version. This resolves a current issue with the block search form.

### Functional testing steps:
- [x] Create a new page and navigate to the layout interface
- [x] Click the 'Add Block' button to expand the layout builder browser
- [x] Type 'image' into the block search tool and verify that the list is filtered to the two blocks with 'image' in the name.


dblanken-yale (approved):
Looks great and works as expected.

PR #252, opened 2023-04-28
YALB-989: Levers: Stub admin interface

codechefmarc:
### Description of work
- Adds the theme settings form as an off-canvas fly-out form that dynamically changes the page depending on what options are chosen.
- Theme settings are not saved until form is saved.
- Cache for theme settings is cleared after saving form.
- Removes settings that will eventually become dials in components.
- Note: At this time, "Line Thickness" and "Line Color" do not do anything - needs discussion and decision. Possibly being moved to dials?

### Functional testing steps:
- [x] View any node page
- [x] Verify that there is a new menu item in the toolbar called "Theme Settings"
- [x] Click on "Theme Settings"
- [x] Choose some options in "Global Theme", "Navigation Position", "Navigation Type", "Header Theme", and "Footer Theme"
- [x] Verify that the new labels from Mike T. are in the theme settings panel for Header Theme and Footer Theme
- [x] Verify that the site changes when changing the settings. Try different combinations of the "Global Theme" and "Header" and "Footer" and verify you get different palettes.
- [x] Save the form at the bottom
- [x] Verify changes made in the settings are kept immediately, no need to clear cache.


web/profiles/custom/yalesites_profile/modules/custom/ys_toolbar/src/ToolbarItemsService.php
```diff
@@ -235,4 +253,54 @@ protected function buildButton(string $route, string $label): array {
...
+        '#url' => Url::fromRoute(
+          $route,
+          $this->redirectDestination->getAsArray(),
+        ),
+        '#access' => $this->accessManager->checkNamedRoute(
+          $route,
+          $this->getNodeRouteParams(),
+          $this->account
+        ),
+        '#attributes' => [
```
nJim, line 282: I was trying to find a way to disable the ajax loading spinner, but I do not see anything avilable within the toolbar_item render element. The spinning circle is distracting, but not a real problem. I will pitch this to the frontend folks for theming. Mentioning here in case you have any ideas.

web/profiles/custom/yalesites_profile/modules/custom/ys_toolbar/src/ToolbarItemsService.php
```diff
@@ -124,6 +137,11 @@ public function addItems(): array {
       );
     }
 
+    $this->toolbarItems['toolbar_theme_settings_link'] = $this->buildOffCanvasButton(
```
nJim, line 140: We got some cross module calls. Let's either:
1. List ys_themes as a dependency of the ys_toolbar module
2. Or create a separate service in ys_themes so it contains its own code

web/profiles/custom/yalesites_profile/modules/custom/ys_toolbar/src/ToolbarItemsService.php
```diff
@@ -235,4 +253,54 @@ protected function buildButton(string $route, string $label): array {
...
+        '#type' => 'link',
+        '#title' => $label,
+        '#url' => Url::fromRoute(
+          $route,
+          $this->redirectDestination->getAsArray(),
+        ),
+        '#access' => $this->accessManager->checkNamedRoute(
+          $route,
+          $this->getNodeRouteParams(),
+          $this->account
```
nJim, line 280: I'm not sure this is what we want. The ability to update theme settings is note necessarily tied to one's permissions to update a node. What do you think about using the existing `yalesites manage settings` permission here? I admit it does not matter at the moment as there is only one real role. But I suspect in the future we will allow folks to edit content but not site settings.

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
```diff
@@ -138,6 +138,26 @@ function ys_themes_preprocess_block(&$variables) {
...
 
+/**
+ * Implements hook_preprocess_node().
+ */
+function ys_themes_preprocess_node(&$variables) {
+  // Add the cache tag, so that the theme setting information is rebuilt when
+  // the config is saved.
+  // Via: https://drupal.stackexchange.com/questions/266379/how-to-clear-cache-for-config-entity-after-making-changes
+  $config = \Drupal::config('ys_themes.theme_settings');
+  \Drupal::service('renderer')->addCacheableDependency($variables, $config);
```
nJim, line 149: Great technique. I wonder if this should be moved to the preprocess_page? Right now this is only clearing the cache for nodes. So caches or will clear for other entities or for dynamically generated pages like the 404 for the search results pages.

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.libraries.yml
```diff
@@ -2,3 +2,9 @@ ys_themes:
   css:
     theme:
       css/ys-themes.css : {}
+levers:
+  css:
+    theme:
+      css/levers.css : {}
+  js:
+    js/levers.js: {}
```
nJim, line 10: I think you need to add a dependency to 'Drupal' since you are using it in the JS. Something else (likely the toolbar) is already loading it, so there is not failure at the moment. But we should still add it here.
```
dependencies:
    - core/drupal
```

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/js/levers.js
```diff
@@ -0,0 +1,52 @@
+((Drupal) => {
+  Drupal.behaviors.ysLevers = {
+    attach: function (context, settings) { // eslint-disable-line
+      once("ysLevers", ".layout-container", context).forEach((element) => {
```
nJim, line 4: Just thinking aloud, the once...forEach pattern is so common in Drupal that I don't think we want to change this. But in reality, this is superfluous code. Even if there are multiple 'layout-containers' on a site, we don't want to change the :root element multiple times. Nothing to resolve unless you want to pull at that string.

nJim (changes requested):
A couple of smaller changes requested. I want to dive deeper into levers.js and there may be a way to simplify this work, but I think this can wait until the rest of the tickets in this series is complete.

joetower (approved):
@codechefmarc This all looks great and functions well, to me! Excellent work :+1:

web/profiles/custom/yalesites_profile/config/sync/ys_themes.theme_settings.yml
```diff
@@ -1,9 +1,7 @@
-action_color: blue-yale
-pull_quote_color: one
 line_color: gray-500
```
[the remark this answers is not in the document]
codechefmarc: Good catch. Removed from the module config.

web/profiles/custom/yalesites_profile/modules/custom/ys_toolbar/src/ToolbarItemsService.php
codechefmarc: I disabled via CSS, just hid the spinner if it was attached to the toolbar items. I couldn't find another way to do it either - I tried a few things inside that service, but most of the other ways I found people doing this was for more regular form API render arrays, not the toolbar items.

web/profiles/custom/yalesites_profile/modules/custom/ys_toolbar/src/ToolbarItemsService.php
codechefmarc: Added `ys_themes` as a dependency, good call.

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module
codechefmarc: Ahh, great idea - moved into the `preprocess_page`

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.libraries.yml
codechefmarc: Added dependencies.

web/profiles/custom/yalesites_profile/modules/custom/ys_themes/js/levers.js
codechefmarc: I get why Drupal does this - for all those pesky ajax calls. No changes made, but I hear ya and I'd love to explore other options another time.

codechefmarc:
@nJim - thanks for your comments! Made those changes and re-requested review. I'll add the line settings work into a new branch so we can get this one merged when approved.

nJim (approved):
Approved.

PR #255, opened 2023-05-02
YALB-1181 - Documentation: Document component and global themes

joetower:
### Description of work
- Adds documentation for component themes and global themes

### Functional testing steps:
- [ ] Read through new documentation
- [ ] Verify it makes sense

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ The YaleSites platform empowers the Yale community to create digital experiences
 - [Getting started as a developer](/docs/development.md)
 - [Learn about the project architecture](/docs/upstream.md)
 - [Frontend review and development](/docs/theming.md)
+- [Working with component themes and global themes](/docs/color-theme.md)
 
 ## Common Scripts
 
--- a/docs/color-theme.md
+++ b/docs/color-theme.md
@@ -1,0 +1,344 @@
+# Working with color and themes
+
+## Jump to:
+- [Adding a new color value](#adding-a-new-color-value)
+- [Adding a new component theme](#adding-a-new-component-theme)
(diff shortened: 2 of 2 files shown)


codechefmarc (approved):
Wow, this is fantastic to see! I didn't run through it directly - I'd like to - but my question is, do you need full releases of the tokens repo to be able to continue development on the component library? If not, I would add a section about this and instructions how to (if it's possible) pull in a feature branch of the tokens library into the component library to be able to test things out ahead of a release. What do you think?

joetower:
@codechefmarc Thank you!

You do not need to create a new tokens release to test things locally. But, as far as I'm aware, you do need to create a new release if you want the changes to be testable on Netlify as part of a deployment from the component library. Do you think that would be helpful to spell out, or does it feel kinda obvious? It kinda feels obvious to me because the changes can be tested locally as you build the tokens repository. But maybe, as written, that is not obvious? 

I'm unsure of a solid workaround for testing on Netlify without a release, being that the component library pulls specific releases of the tokens repository. Forcing a version change (in package.json) for the sake of testing on Netlify might get gross, especially if you forget to change it and merge the component-library PR.

nJim (approved):
This is much needed. Looks good. I didn't test the process, nor do I need to for approval. But I'm eager for us to try adding another color as a proof. Thank you much. Ship it.

codechefmarc:
@joetower - Of course, I should have thought about it for half a second more, of course we can do all of this locally to test before creating the PR's. 👍 I'll add to my list to eventually test this myself as well as I'm super curious how the whole process goes. I like knowing how the sausage is made. :P

PR #260, opened 2023-05-05
YALB-855: Close modal after adding block

nJim:
### Description of work
- Applies a patch to close the block modal after adding the block to the page.
- See https://www.drupal.org/project/layout_builder_browser/issues/3352754

### Functional testing steps:
- [x] Add a block to the page. Any type will do.
- [x] Fill out all fields on the Configure block form.
- [x] Click the 'Add Block' button on the bottom of the form.
- [x] Verify that the modal 'Configure block' form closes after adding the block.


codechefmarc (approved):
Works perfectly. Approved!

PR #261, opened 2023-05-08
Hide revision field patch cleanup

codechefmarc:
### Description of work
- Removes patch that has now been incorporated not the latest version of the hide revision field module.
- Fixes an error during `composer install` where this patch could not be applied.

### Functional testing steps:
- [ ] Verify there are no `Deprecated function` errors loading the site
- [ ] Verify there are no `Deprecated function` errors in the logs


nJim (approved):
I need this! Thank you for getting ahead on this.

PR #262, opened 2023-05-09
fix(YALB-1123): Cache invalidated on alert save

dblanken-yale:
### Description of work
- Add ys_alert preprocessor hook to add a cache dependency such that updating invalidates rendered cache

### Functional testing steps:
- [x] Click `Settings->Alert Settings`
- [x] Change something about the current alert
- [x] Click the `Save configuration` button, waiting to see that the configuration was successfully saved
- [x] Visit the root site again
- [x] Verify that the alert has the changes you made

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_alert/ys_alert.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_alert/ys_alert.module
@@ -24,3 +24,17 @@ function ys_alert_theme($existing, $type, $theme, $path) {
     ],
   ];
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function ys_alert_preprocess_page(&$variables) {
+  $config = \Drupal::config('ys_alert.settings');
+  $variables['alert'] = $config->get('alert');
+
+  // Add the cache tag, so that the theme setting information is rebuilt when
+  // the config is saved.
+  // Via: https://drupal.stackexchange.com/questions/266379/how-to-clear-cache-for-config-entity-after-making-changes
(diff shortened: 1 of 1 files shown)


web/profiles/custom/yalesites_profile/modules/custom/ys_alert/ys_alert.module
```diff
@@ -24,3 +24,17 @@ function ys_alert_theme($existing, $type, $theme, $path) {
     ],
   ];
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function ys_alert_preprocess_page(&$variables) {
+  $config = \Drupal::config('ys_alert.settings');
+  $variables['alert'] = $config->get('alert');
```
codechefmarc, line 33: I don't think you need this line to set the `$variables['alert']` - that was in the `ys_themes` module to set the global theme for twig templates, but for the alert, you should only need the `$config` to send to the Drupal service. Otherwise, tested this out and all works great!

codechefmarc (changes requested):
I tested this out and it worked great for me. Just remove the one line and you're good to go!

web/profiles/custom/yalesites_profile/modules/custom/ys_alert/ys_alert.module
dblanken-yale: You are correct.  Thanks for catching that.  The config is all we needed.  Thanks again for your help!

dblanken-yale (commented):
Doh I didn't see you did this until I already pushed.  :D Sorry about that.

codechefmarc (approved):
Works great, approved!

PR #263, opened 2023-05-09
YALB-395: Metadata: Fallback for teaser images (BE)

codechefmarc:
### Description of work
- Adds a fallback teaser image picker to the Site Settings form
- Adds Twig function and a service for retrieving and setting YaleSites core settings (replicated from `ys_themes`)
- Requires [Atomic PR-121](https://github.com/yalesites-org/atomic/pull/121)
- [Testing multidev here](https://pr-263-yalesites-platform.pantheonsite.io/)

### Functional testing steps:
- [ ] Visit Site Settings form here `/admin/yalesites/settings`
- [ ] At the bottom of the form, select a "Fallback teaser image" from the media library (or upload your own) and then save the configuration
- [ ] Add 2 event nodes to the site. For one of them, select a teaser image in the edit screen.
- [ ] For both events, you must also set the date and time of the event to be in the future.
- [ ] Add or edit 2 post nodes. For one of them, select a teaser image in the edit screen.
- [ ] Add a page to the site, save and enter layout builder after save
- [ ] Add a View block and set the following criteria: Show posts as a post card grid, no tags, display all items.
- [ ] Add the block and save the page.
- [ ] Verify that the post that has a teaser image set is showing correctly and the post that does not have a teaser image set is the fallback image that was set in Site Settings
- [ ] Go back into layout builder, configure the view block, and change the view from "Post Card Grid" to "Post List"
- [ ] Update the block, save the layout, and verify that the images follow the same rules for fallback vs selected image.
- [ ] Repeat the last few steps but instead select "Events" instead of "Posts" in the view configuration and verify that event teasers follow the same rules.


web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/CoreSettingsManager.php
```diff
@@ -0,0 +1,69 @@
+<?php
```
nJim, line 1: I'm curious about CoreSettingsManager.php. I like having these kinds of services to better organize the business logic. I like encapsulating work that has a single purpose. It's especially important when we need to DRY up code.

But currently this file does not seem to do much. get calls the configs get, and set calls the configs set. So why use this middleware to manage that transaction? It would be different if there was some filter or other business logic taking over, but right now it's a clear pass through.

I'm not against it, but would like to hear your take on it.

web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/CoreSettingsManager.php
```diff
@@ -0,0 +1,69 @@
...
+    );
+  }
+
+  /**
+   * Gets core setting from config.
+   *
+   * @param string $setting_name
+   *   Setting machine name.
+   */
+  public function getSetting($setting_name) {
```
nJim, line 45: Please include @retrun type params (I believe it is an immutable config) and also type hint at least the arguments on the function declaration. 
`public function getSetting(string $setting_name)`

nJim (changes requested):
Great work. I've added a big question on the service being used. Nothing needs to change, but at first glance the code does not seem to be offering a lot of value compared to using the config service directly. 

My only requested changes are more of style. Please make sure all methods have fully defined annotations about their params and returns. Also please use the type hinting on the arguments themselves. I left a comment on one method, but please check them all.

web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/CoreSettingsManager.php
codechefmarc: I can remove this and replace with direct config calls - no problem there. I was following the pattern we established with the `ys_themes` module so if we needed to add any more logic later, we could do it here. But, agreed, we can cross that bridge if we need to later.

codechefmarc:
@nJim - Removed the manager and refactored. Ready for re-review. Thanks!

nJim (approved):
Approved. Ship it.

PR #264, opened 2023-05-09
YALB-457: sets max-length for teaser text fields

codeb-a:
### Description of work
- Sets the "Teaser Text" field to have a maximum length of 150 characters, on all content types.
- Sets the "Teaser Text" field to have a hard limit for all content types.

### Functional testing steps:
- [ ] Go to structure/content types
- [ ] Go to manage fields/manage form display for all content types (pages, posts, events)
- [ ] Check that the maximum length is set to 150 characters and "hard limit" is toggled on.


nJim (approved):
Perfect. Just what we needed.

PR #265, opened 2023-05-09
YALB-814: Metatags Open Graph (SB/BE)

codechefmarc:
### Description of work
- Adds a new token to get the teaser title if set, but if not, fallback to the node title.
- Enables and configures Open Graph metadata globally and also for articles

### Functional testing steps:
- [x] Edit a page on the site and add some teaser text (do not add teaser title for now), and a teaser image
- [x] Save the page
- [x] View the page and inspect the HTML
- [x] Verify that there are the following meta tags in the `head`:
     - [x] `description` - teaser text
     - [x] `og:site_name` - the site title
     - [x] `og:type` - "website"
     - [x] `og:url` - the URL of the current page
     - [x] `og:title` - the title of the current page
     - [x] `og:description` - teaser text
     - [x] `og:image` - URL to the teaser image
     - [x] `og:image:alt` - Alt text of the teaser image
- [x] Edit the page again and add a teaser title
- [x] Save the page
- [x] View the page and inspect the HTML
- [x] Verify that the following meta tag has changed in the `head`:
     - [x] `og:title` - teaser title instead of the page title
- [x] View a post node and inspect the HTML
- [x] Verify that there are the following additional meta tags in the `head`:
     - [x] `og:type` - changed to "article" instead of "website"
     - [x] `article:published_time` - matches the date in the Publish Date field in the post

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.tokens.inc
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.tokens.inc
@@ -24,6 +24,10 @@ function ys_core_token_info() {
     'name' => t('Events landing page'),
     'description' => t('The path alias for the events landing page.'),
   ];
+  $info['tokens']['yale']['teaser_title_node_title'] = [
+    'name' => t('Teaser title or node title'),
+    'description' => t('Will use the teaser title of the node if set, and if not will fallback to the node title'),
+  ];
   return $info;
 }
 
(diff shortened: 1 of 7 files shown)


web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.tokens.inc
```diff
@@ -51,6 +55,14 @@ function ys_core_tokens($type, array $tokens, array $data, array $options, Bubbl
           $config = \Drupal::config('ys_core.site');
           $replacements[$original] = $config->get('page.events') ?: '/events';
           break;
+
+        case 'teaser_title_node_title':
+          if (!empty($data['node'])) {
+            $nodeTitle = $data['node']->getTitle();
+            $teaserTitle = ($data['node']->field_teaser_title->first()) ? $data['node']->field_teaser_title->first()->getString() : NULL;
```
nJim, line 62: This works fine but may not be future proof. `($data['node']->field_teaser_title->first())` will throw a fatal error if a node does not have a the teaser title field -- as you can not call the first method on null. But I *think* if you change this expression to `($data['node']->field_teaser_title)` it will fail gracefully.

nJim (changes requested):
Everything works perfectly. One change requested.

web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.tokens.inc
codechefmarc: Very good catch. Turns out we need to do both - check for the field to exist, but also to check for the `first()` as if that doesn't exist, no teaser text is inputed. Fixed and re-review.

nJim (approved):
Approved. Ship it.

PR #266, opened 2023-05-10
YALB-839: Enable and configure Google Search Console

nJim:
### Description of work

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/metatag.metatag_defaults.global.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/metatag.metatag_defaults.global.yml
@@ -8,6 +8,7 @@ id: global
 label: Global
 tags:
   canonical_url: '[current-page:url]'
+  google_site_verification: '[yale:ys_google_site_verification]'
   og_site_name: '[site:name]'
   og_type: website
   og_url: '[current-page:url]'
--- a/web/profiles/custom/yalesites_profile/config/sync/metatag.metatag_defaults.global.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/metatag.metatag_defaults.global.yml
@@ -8,4 +8,7 @@ id: global
 label: Global
 tags:
   canonical_url: '[current-page:url]'
+  og_site_name: '[site:name]'
+  og_type: website
+  og_url: '[current-page:url]'
   title: '[current-page:title] | [site:name]'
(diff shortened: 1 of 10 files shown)


web/profiles/custom/yalesites_profile/config/sync/metatag.metatag_defaults.global.yml
```diff
@@ -8,6 +8,7 @@ id: global
 label: Global
 tags:
   canonical_url: '[current-page:url]'
+  google_site_verification: '[yale:ys_google_site_verification]'
```
dblanken-yale: Just verifying: In [6476a316ce86eba687f33f130d44450e63b22ae8](https://github.com/yalesites-org/yalesites-project/pull/266/commits/6476a316ce86eba687f33f130d44450e63b22ae8) you replaced `ys_google_site_verification` with `google_site_verification`.  Is that also meant for here or is `ys_google_site_verification` referring to something different in the yml now?

dblanken-yale: If it's too long to explain, it's ok, but is this the sync yml that's defining the meta tags to render when visiting any HTML page?  If so, how is it distinguishing the underscores to mean either a dash in the case of meta name `google-site-verification` and a colon in the meta property for like `go:site-name`; looks like magic at the moment.  :D

dblanken-yale:
I was able to follow the directions and saw the meta tag with my content.  Great work!

web/profiles/custom/yalesites_profile/config/sync/metatag.metatag_defaults.global.yml
nJim: @dblanken-yale I had a change of heart when reviewing my own code. Any installs should work but it's possible the multidev was still out of date. All the name ys_google_site_verification was changed to google_site_verification. Since this token is in a Yale-namespaced group, I did not see a reason for superfluous ys_ prefix. So I removed it. This is similar to the other tokens in this group. Either name would work but we need to be consistant, so what's in my token declaration matches the name in the yml. Good question. Let me know if you want to dig deeper.

Connecting it all together.
- The settings form has a variable `google_site_verification`
- This variable is saved as configuration in the `ys_core.site.yml` file and is accessible as `seo.google_site_verification`. I bet we will add other seo value here in the near future.
- This change also defines a new token `$info['tokens']['yale']['google_site_verification']` which can be used to render the value that we stored in config.
- This token was added using the Drupal Metatags Verification module. This will render that meta tag on all pages.

dblanken-yale: It's very cool to see how it all goes together.  Thanks for explaining that part.  I agree, the namespace makes up for it for sure.

PR #267, opened 2023-05-11
YALB-1163 - Create Button Link component

joetower:
### Description of work
- Adds button-link block type to layout-builder
- Allows two `link` fields to be added to the new button-link component

### Functional testing steps:
- [ ] Test here:


codechefmarc (changes requested):
Almost all of this works great for me. The one thing I noticed when setting the grand hero to "Tall", images work fine, but the video doesn't extend to the bottom of the component. I know videos can be strange in this way. Attached screenshot:

Screenshot 2023-05-12 at 9 03 11 AM

joetower:
@codechefmarc great catch, thank you! I pushed up a fix. This should be ready for another look. 
https://pr-267-yalesites-platform.pantheonsite.io/components

codechefmarc (approved):
Wrapped image and video height on grand hero both work great now. Approved!

PR #272, opened 2023-05-12
YALB-1190: Fixes alert setting select bug

codeb-a:
### Description of work
- Fixes a bug on the alert form so the current selected option has a blue border and a darker border when using the keyboard tab.

### Functional testing steps:
- [ ] Go to the site and click on settings/alert settings in the admin menu.
- [ ] Verify the currently selected alert type has a light blue border
- [ ] Verify using the keyboard to tab and the arrows keys to select different alert types the border is dark blue.
- [ ] Verify that the light blue border no longer stays stuck on one option when using the keyboard navigation. (Should be moving with the arrows keys.)


nJim (approved):
Perfect. Just what we needed. Ship it.

PR #273, opened 2023-05-16
YALB1145: Levers form: Add icons and custom controls

joetower:
### Description of work
- Themes the global theme color palette options
- Themes the component options
- Dynamically reads the currentTheme and applies the currentTheme to the appropriate inline css styles on the form input span elements

### Functional testing steps:
- [ ] Login here: https://pr-273-yalesites-platform.pantheonsite.io
- [ ] After logging in, navigate to a page like https://pr-273-yalesites-platform.pantheonsite.io/components and click the "Theme Settings"
- [ ] Change the color palette and observe how the change affects the visual representation of each component lever

https://github.com/yalesites-org/yalesites-project/assets/366413/dfeeeb1a-ab84-409e-9853-b097c2ca5911

- [ ] Note that the Header and Footer themes are named appropriately, according to the two slot names in use.
- [ ] Footer theme and header theme slot names represent the border color (`base`) and the background color (`white`).


ahitson:
This is so dope. But, here are my changes:

- Header themes shouldn't use the color names, because they are combinations of the colors and don't have the same meaning. They need their own color set. Like white, light grey, dark...or something.

- Footer themes also need their own names and can't use the slot names. 

- Can the circle outlines be a lighter grey instead of black? Like Grey 500?

- The selection is different on each and would like to unify them. Can they all have a selected highlight like in the color selection? And standardize that the outline goes around the title and image.

- There is some link color weirdness happening on the header themes. Navigation shouldn't change colors between light options.

- Footer link hovers should just add an underline instead of changing colors. It's creating some inaccessible combos.

joetower:
@ahitson Thank you! 

This is ready for another look when you have time. I did not change the header or footer theme names, yet. We should discuss this a bit internally, I'd like to better understand from you, Jim, and Marc why the current naming convention was adopted. It is not a big deal to change any of them, though. 

- The color swatch outlines/circles are now gray 500
- Each radio button has the same styles - border, and background color when selected
- I have fixed the header and footer link colors :partying_face: . You should now see header hovers change, only, mapped to slot-2. Footer links should underline on hover and not change color.  

https://github.com/yalesites-org/yalesites-project/assets/366413/986e93b4-4104-490a-b9e8-fac26ab286ca

ahitson:
Looking good! Now let's change the names. Let's do the header as White & Base, Light Grey & Highlight, Base & Highlight (in that order). And footer would be White & Base, White & Deep, Light Grey & Highlight, Dark Grey & Highlight, Base & Highlight (in that order).

ahitson (approved):

codechefmarc (approved):
This all looks good to me! Approved! The only thing I can see that we may want to look at is the JS for this is currently using `document` instead of `context` which is more a Drupal standard. I did that as well in my code in the JS file and didn't test `context`. That said, I think some of the `document` calls are correct as we're touching the root instead of just the context which is only the slide out drawer. This all works as is and it's not a blocker so this can be moved forward.

miketullo95:
This looks great! can we get the same naming convention with the button theme too? right now it has one, two and three, change to base, highlight, and deep (?). @ahitson are these terms accurate?

ahitson:
@miketullo95, close. The options would be Base, Action, and Deep.

joetower:
@miketullo95 @ahitson I updated the button theme labels: 

Screenshot-20230522140140-380x501


Also, Mike, I visually-hid the `radio` inputs. 
Screenshot-20230522140407-2552x1359


https://pr-273-yalesites-platform.pantheonsite.io/components

miketullo95 (approved):

kara-franco (changes requested):
Hiya! Impressive, the fieldset/legend is really nice, sounds great on a screen reader. 

The only issue I came across while testing on both screen readers and my keyboard is that when a user selects radio input with the "enter" key, the whole modal shuts. This behavior does not occur when a user uses the "spacebar" key. 

- [ ] Inputs should be selectable with both "spacebar" and "enter". 

If you need any further insight, @ntay90 can provide guidance!

joetower:
@kara-franco Thank you! 

I want to make sure I understand the resolution here. If we make the form submit/close using the `spacebar` key would that solve the issue? 

cc: @ntay90

nJim (approved):
Approving. Works fantastic. Follow up tickets are created for a possible refactor and a keyboard accessibility issue.

PR #275, opened 2023-05-17
YALB-1064: Add preview icons

codeb-a:
### Description of work
- Adds preview icons to layout builder blocks.

### Functional testing steps:
- [ ] go to a page content type
- [ ] click layout builder at the top of the page to edit
- [ ] verify that when clicked on blocks for each section the previews for each block is displaying.

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/layout_builder_browser.layout_builder_browser_block.accordion.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/layout_builder_browser.layout_builder_browser_block.accordion.yml
@@ -6,6 +6,6 @@ id: accordion
 block_id: 'inline_block:accordion'
 category: all_blocks
 label: Accordion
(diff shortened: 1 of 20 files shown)


nJim (changes requested):
I have a few changes; my appoligies as some of them are me being pedantic.

1. Could you please move these image into the yalesites-project repo? right now these images do not appear as they are not stored anywhere. I'm thinking they should go in an image folder inside the '/home/jim/sites/yale/yalesites-project/web/profiles/custom/yalesites_profile/modules/custom/ys_core' module.
2. Can you please rename the files to use all lowercase letters? And also use hyphens instead of spaces or underscores in the file names? This is just a Drupal standard that I want to make sure we are following.

codeb-a:
@nJim This should be good to go now.

nJim (approved):
Just what we needed.

PR #279, opened 2023-05-19
YALB-971: LocalDev theme setup scripts

dblanken-yale:
### Description of work
- Includes [YALB-715](https://yaleits.atlassian.net/jira/software/projects/YALB/boards/368?selectedIssue=YALB-715) as it's closely related
  - Script prepends script name to any output so you can know what is running (useful later when other scripts attempt to use this one)
- New script `local-git-checkout.sh` that can safely change branches on any of the 3 repositories (`token`, `component-library-twig`, and `atomic`)
- Create symlinks to each repo at the base directory to quickly get to the development versions (gitignored)
- Give error messages should the script error, such as directories not existing
- Clears Drupal cache if the `atomic` branch is changed
- Other scripts utilize new `say` utility method to help readability: i.e. `local-cl-dev.sh`
- It will not switch repos if there are conflicting git changes uncommitted

### Usage

For Help:
```bash
npm run local:git-checkout -- -h
```

To attempt to change all repo branches to the same (reverts back to main (or develop for yalesites) if doesn't exist):
```bash
npm run local:git-checkout -- -b YALB-xx-whatever-your-branch-name-is
```

To target individual branch names for each repo (-c = Component Library, -a = Atomic, -t = Token, -m = YaleSites main repo)
```bash
npm run local:git-checkout -- -c YALB-11 -a YALB-33 -t figma -m YALB-12
```

To debug any problems you might see:
```bash
npm run local:git-checkout -- -d
```

To get even more verbose execution messages:
```bash
npm run local:git-checkout -- -v
```

To set up CL, Atomic, and Tokens to main branch, and YaleSites main repo to develop, run without arguments:
```bash
npm run local:git-checkout
```

### Functional testing steps:
- [x] `cd` to `yalesites-project` directory (best on fresh version)
- [x] Run `npm run local:git-checkout`
- [x] Verify no errors occurred
- [x] Run `ls`
- [x] Verify that the following symlinks exist: `atomic`, `component-library-twig`, and `tokens`
- [x] Visit your local yalesites website and verify that the pages look like tokens are being loaded
- [x] Use it with a repo setup you're currently needing in case I missed a permutation

[YALB-715]: https://yaleits.atlassian.net/browse/YALB-715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,3 +125,8 @@ reference
 # Yalesites profile assets
 web/profiles/custom/yalesites_profile/composer.lock
 web/profiles/custom/yalesites_profile/vendor/
+
+# Local development symlinks
+atomic
+component-library-twig
+tokens
--- a/.gitignore
+++ b/.gitignore
@@ -125,8 +125,3 @@ reference
 # Yalesites profile assets
 web/profiles/custom/yalesites_profile/composer.lock
 web/profiles/custom/yalesites_profile/vendor/
-
-# Local development symlinks
-atomic
-component-library-twig
-tokens
--- a/.gitignore
+++ b/.gitignore
@@ -125,3 +125,8 @@ reference
 # Yalesites profile assets
 web/profiles/custom/yalesites_profile/composer.lock
 web/profiles/custom/yalesites_profile/vendor/
+
+# Ignore symlinks for dev
+/atomic
+/component-library-twig
+/tokens
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "local:review-with-cl-branch": "./scripts/local/local-review-with-cl-branch.sh",
     "local:cl-dev": "./scripts/local/local-cl-dev.sh",
     "local:theme-link": "./scripts/local/local-theme-link.sh",
+    "local:git-checkout": "./scripts/local/local-git-checkout.sh",
     "prettier": "prettier '{.github,docs,web/profiles/custom,web/modules/custom,web/themes/contrib/atomic,web/themes/custom}/**/*.{json,md,js,html,scss,html}' --ignore-unknown --list-different",
     "setup": "./scripts/local/setup.sh",
     "test": "./scripts/local/test.sh"
--- a/scripts/local/local-cl-dev.sh
+++ b/scripts/local/local-cl-dev.sh
@@ -11,46 +11,14 @@
 # through the regular component library release process, and be included in an
 # official release of Atomic.
 
-[ $# -eq 1 ] && [ "$1" == "-h" ] && echo "Usage: $0 [-d]" && exit
-[ $# -eq 1 ] && [ "$1" == "-d" ] && _say "Debug mode enabled" && set -x
-
-[ -f "./scripts/local/util/say.sh" ] || (echo -e "You must run this from the yalesites root directory" && exit)
+[ -e "./scripts/local/util/say.sh" ] || (echo -e "[$0] Say utility not found.  You must run this from the yalesites root directory: " && exit 1)
 source ./scripts/local/util/say.sh
 
- _say "Moving to atomic repo"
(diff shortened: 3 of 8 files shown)


codechefmarc (commented):
Amazing work, @dblanken-yale! I ran through this a few times with some different branches and all looked to be working just fine for me. Love how it adds the synlinks to the root - MUCH easier to get to them rather than `cd`ing all over the place.

Approved from my end, but want to be sure to get @joetower and @nJim eyes on this as well before fully approving. A few comments:

1. the `-b` option is a tad confusing (I figured it out) in that the regular command `git checkout -b` will create a new local branch where with yours it's not creating branches but switching to existing branches. Perhaps use a different letter?
2. Noticed this doesn't switch the main repo over, only the theming repos. I think this is fine, but wanted to check in with the other devs to see what they thought?
3. After switching my main repo and using your new script, I, of course, had to run an `npm run build-with-assets` or `npm run build-with-install` which is no problem. But then, after that, I needed to re-run `npm run local:cl-dev` in order for the new components to show up from the new component library repo. Again, I think this is fine as long as we know the steps, but wanted to make a note of it to see if we wanted to "automate" more things too? For example, perhaps another option in your script to switch the main repo too, run the `assets` or `install` npm command, then `cl-dev`? But make all that optional unless you specify the flag?

codechefmarc:
On second thought, my point about running one of the `assets` or `install` commands may not be a good idea as that is potentially destructive change locally, so perhaps having us run those scripts manually is a better idea rather than adding it to your script so we are sure that is what we are doing. That said, I still like the idea of possibly running `cl-dev` as part of your script? But I'd like Joe to weigh in on that idea.

dblanken-yale:
@codechefmarc Thanks a lot for trying it out!

Ya I wasn't sure whether to include the yalesites branch or not.  It's trivial to add it if we want to include it.  My thought was that without any parameters, this would set up the dev environment so you could choose to go into any part of it at that point and change or create branches, so this was basically setting up the three repos so you could do something else to the YS branch, so I left it as is.  But we can be flexible with it and add options to do either if that'd help.

For the -b, I was trying to have a shortcut where if you knew you had three repos with the same name you wouldn't have to be so verbose about which to change, and my original thought was -a for all, but then atomic had taken that, so -b for branches was where I went.  We can even remove that for now since I agree it can be confusing and sort of against git.  Or, the alternative could be we could have it create branches if it didn't exist, and not revert back to main.

For the cl library, I found that after running the script, you can run `npm run build` and it'll pick it up (with a `lando drush cr`). That's been my workflow, at least, and it's been pretty seamless.

Yes, we can definitely set up a flag to do any of that.  I'll let others chime in too; I'll be happy to adjust it as needed.  I'm just glad it's working.  :D  I've been using it during the SoundCloud PR and so far I'm able to function pretty well in all branches.

dblanken-yale:
Maybe if you want the main branch to change as well we could add a "-y" for YaleSites or a "-m" for main repo?

codechefmarc:
> Maybe if you want the main branch to change as well we could add a "-y" for YaleSites or a "-m" for main repo?

Humm, (naming is hard), I'd avoid `-y` as that implies "answer yes". `-m` would work though!

dblanken-yale:
Main branch change options pushed.  `-m` or using `-b` should now attempt to change the yalesites branch, but remember it'll still check for uncommitted changes and not allow you to change otherwise.  Defaults to develop branch.

joetower (approved):
@dblanken-yale This is pretty awesome! Thank you for working on this. I think it's a great improvement. :clap: nice work!

dblanken-yale:
Committing after some git squashing.

PR #281, opened 2023-05-22
YALB-1199: Media: Organize media on install (BE)

codechefmarc:
### Description of work
- Adds additional images into the content on install
- Changes location of original images for new versioning directory structure on S3

### Functional testing steps:
- [ ] Visit the media library `/admin/content/media-grid`
- [ ] Verify there are now 38 images in the library.
- [ ] Edit any of the new images and verify there is alt text inputted for the image

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_starterkit/content/images.json
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_starterkit/content/images.json
@@ -3,53 +3,193 @@
     "images": [
       {
         "image_id": 1,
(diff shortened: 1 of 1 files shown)


codechefmarc:
@vinmassaro - I noticed that after creating the multidev for this, the images weren't imported automatically. I checked the deploy to Pantheon step and it does a config import, but not a migrate import. I ran the migration for this PR manually and all worked well.

I don't think we need to do this every time, of course, but wanted to ask your opinion about subsequent times we want to import new images into the platform for all sites. I assume as part of your deploy work, there may be a step you can add (optionally?) to import images / other content as we need to in the future. Just something to think about.

PR #289, opened 2023-05-25
YALB-818 - Views: Add icons as controls

joetower:
### Description of work
- Adds svg icons to use in views basic
- Updates `cta__component_theme` lever/dial variable to be added to the `button_link` component only, instead of all inline blocks.

### Functional testing steps:
- [x] View this page: https://pr-289-yalesites-platform.pantheonsite.io/components
- [x] Note that if you change the `global_theme` color palette or the `button_theme`, the CTA in the banner does not change color. The `button_link` component _does_ change. 

https://github.com/yalesites-org/yalesites-project/assets/366413/b2d19680-cc06-45fd-9bd5-b18187ac2afa

--- 
- [x] Add a view block to the page
  - [x] Verify the block icons render well 

https://github.com/yalesites-org/yalesites-project/assets/366413/debab7fd-89f6-4bc8-b2ea-8576a5b1c762

---
- [x] Edit the view block you created
  - [x] Verify the block icons render well in the condensed sidebar instance 

https://github.com/yalesites-org/yalesites-project/assets/366413/5bbef6ad-dc4c-4a8f-97ef-866cfba9e6ed


web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/assets/js/views-basic.js
```diff
@@ -0,0 +1,39 @@
+((Drupal) => {
+  Drupal.behaviors.ysLevers = {
```
codechefmarc, line 2: Rename this to `Drupal.behaviors.ysViewsBasic` as there already is a `Drupal.behaviors.ysLevers` in the `ys_themes` module. I don't think it would cause a conflict, but just to be safe.

codechefmarc, line 2: Rename this to `Drupal.behaviors.ysViewsBasic` as there already is a `Drupal.behaviors.ysLevers` in the `ys_themes` module. I don't think it would cause a conflict, but just to be safe.

codechefmarc (changes requested):
Works as expected and looks great! One minor change to the JS and then you're good to go!

joetower:
@codechefmarc Great catch, thank you! 

I updated the behavior name and the JS code comment. I also updated some files with an empty line because I missed that, too. :facepalm:

codechefmarc (approved):
Thanks for the changes. Tested and approved!

nJim (approved):
Approved.

miketullo95 (approved):

PR #276, opened 2023-05-17
YALB-1198: Events: Exposed filters on dynamic list (BE)

codechefmarc:
### Description of work
- Adds event category taxonomy vocabulary
- Adds event category as a term reference into events
- Adds event category exposed filter to calendar list - limits terms to only ones in use
- Adds event category filtering to views basic - visible only when event is selected
- Adds operator (AND/OR) functionality to views basic
- Allows multiple terms to be selected / filtered on for tags and event categories (hence why operator was added)
- Adds "No results found." message to view if no results are found.

### Functional testing steps:
- [ ] Visit `/admin/structure/taxonomy/manage/event_category/add` and `/admin/structure/taxonomy/manage/tags/overview` and add a few terms to each vocabulary.
- [ ] Add multiple events to the site. Make sure to add dates in the future or else they won't show up in views.
- [x] Verify there is a new "Category" taxonomy field on the event edit form.
- [x] Verify the new autocomplete deluxe field allows to select existing and create new terms - this works best with some terms already in the system.

Screenshot 2023-05-17 at 3 47 12 PM

- [x] For each event, tag and categorize with some events getting the same terms and others getting different terms. For testing, use the term in the title to help with filtering later.
- [x] Create a new page, title it, save, and enter layout builder.
- [x] Add a block of type "Calendar List".  After adding the block, save layout builder.
- [x] Verify that the calendar list has an "Event Category" filter on the top. Note: This is currently un-styled.

Screenshot 2023-05-17 at 3 52 15 PM

- [ ] Verify that the terms listed in the filter are the only ones that are currently in use - it does not show all terms.
- [x] Re-enter layout builder and add a block of type "View"
- [x] Verify that there is a new "Match Content That Has" area with two options that control how terms should be matched to nodes (Any = OR, All = AND)

Screenshot 2023-05-17 at 4 00 18 PM

- [x] Select "Events" under "I Want To Show"
- [x] Verify that the "Event Category" autocomplete text box is added to the form

Screenshot 2023-05-17 at 4 00 23 PM

- [x] Add in some tags / categories via autocomplete to filter by and select either the "Any" or the "All" under the "Match Content That Has". Note, tags and categories can support multiple terms now.
- [x] Add the block.
- [x] Verify that there are results for the filters (Can be done without saving layout builder)
- [x] If there are filters that don't match any content, verify that there is a "No results found." message that appears for that block.
[...]

PR #290, opened 2023-05-25
YALB-1211: Exclude a term in views

codechefmarc:
### Description of work
- Includes work from [PR 276 for YALB-1198](https://github.com/yalesites-org/yalesites-project/pull/276) as a starting point
- Adds the ability to exclude terms in a view
- Reworks how terms are chosen on the view settings form: there are two fields, one for including terms, and the other for excluding terms.
- When including terms, the "Any (OR) /All (AND)" operator works as expected
- When excluding terms, this will ignore the "Any/All" operator and instead default to "Any (OR)" as otherwise things get very confusing very quickly. We can change this if we want to add another "operator" control to control just the exclude in addition to the include.
- Adds an entity reference view to help when choosing terms for the two reference fields above. This new view simply adds the vocabulary name in parentheses when performing an autocomplete search to help find the correct term - say if the same term name was used in two different vocabularies.

### Functional testing steps:
- [x] Add some event categories and tags to the site. If logged in as an admin, this can be done [here for events](https://pr-290-yalesites-platform.pantheonsite.io/admin/structure/taxonomy/manage/event_category/add) and [here for tags](https://pr-290-yalesites-platform.pantheonsite.io/admin/structure/taxonomy/manage/tags/add). If not an admin, these can be added on the event and news edit pages as part of those fields.
- [x] Create some events on the site and title them in a way to help test out which categories and tags are part of each event. Add some categories and tags to each. Also, make sure to add a date or else they won't show up in the view.
- [x] Add a page to the site, add a title, and save. In Layout builder, after save, add a view.
- [x] Verify there are 2 fields for entering taxonomy terms: "Include content that uses the following terms" and "Exclude content that uses the following terms"

Screenshot 2023-05-26 at 1 46 47 PM

- [x] Select "Events" under "I want To Show"
- [x] Add some terms via autocomplete into one or both boxes
- [x] Verify that when performing an autocomplete, the suggestions show the taxonomy vocabulary in parentheses next to the term to verify which vocabulary the term comes from.

Screenshot 2023-05-26 at 1 47 24 PM

- [x] Save the block and verify that the view changes to represent the taxonomy filters.
- [x] Repeat as necessary to test different combinations of terms: multiple terms included with both the "Any" and "All" selector below, also testing the ability to exclude one or multiple.
- [x] Verify that excluding multiple terms will mean that those terms don't have to show up on the same node in order to be excluded. (Forced to be an "OR" for excluding).

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/field.field.block_content.view.field_view_params.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/field.field.block_content.view.field_view_params.yml
@@ -18,8 +18,8 @@ translatable: false
 default_value:
   -
     group_params:
-      params: '{"view_mode":"card","filters":{"types":["post"],"tags":null,"event_category":null},"operator":"or","sort_by":null,"display":null,"limit":0}'
-    params: '{"view_mode":"card","filters":{"types":["post"],"tags":null,"event_category":null},"operator":"+","sort_by":null,"display":null,"limit":0}'
+      params: '{"view_mode":"card","filters":{"types":["post"],"terms_include":null,"terms_exclude":null},"operator":"+","sort_by":null,"display":null,"limit":0}'
+    params: '{"view_mode":"card","filters":{"types":["post"],"terms_include":null,"terms_exclude":null},"operator":"+","sort_by":null,"display":null,"limit":0}'
 default_value_callback: ''
 settings: {  }
 field_type: views_basic_params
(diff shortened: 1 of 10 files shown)


nJim (approved):
Approved. We have a few more changes coming around this work, so I want to see how complex this can get before considering any big shifts. This may be the most direct way to write things without messing up AJAX. I find this approach very readable. It's just a lot of code to work though. Really excited to see it working!

PR #293, opened 2023-05-31
YALB-971: LocalDev: Theme setup scripts fix

dblanken-yale:
fix(YALB-971): force composer for atomic

When setting up a yalesites project, composer does not use the name 'origin', but instead uses a remote name of 'composer'.  This caused some issues when attepmting to switch branches on atomic, as there would be two tracking branches, and it didn't know which to take.  This now passes the 'composer' remote only for atomic and passes it through.

### Description of work
- Fix script to allow the `composer` remote for atomic only
- YaleSites repo defaults to `develop` if branch doesn't exist

### Functional testing steps:
- [ ] Create a fresh git clone of the project repo
- [ ] Run `npm run setup` to set up initial items
- [ ] Run `npm run local:git-checkout -- -b <branch name you care about)`
- [ ] Verify that you successfully retrieved the branches you needed, specifically from atomic, but others as well

### DIFF
--- a/scripts/local/local-git-checkout.sh
+++ b/scripts/local/local-git-checkout.sh
@@ -179,10 +179,12 @@ function repo_has_changes() {
 # Main function that does the work
 function _local-git-checkout() {
   # Default values
-  local cl_branch="main"
-  local token_branch="main"
-  local atomic_branch="main"
-  local yalesites_branch="develop"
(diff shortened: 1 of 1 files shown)


codechefmarc (approved):
Tested this out with actual work branches and works great! Approved.

PR #295, opened 2023-06-01
YALB-1239: Filter to show past events

codechefmarc:
### Description of work
- Adds the ability to filter events in the views tool to show only future events, only past events, or all events.
- Forces all views to only show published nodes.

### Functional testing steps:
- [x] Create some events on the site and set the event date of a few of them in the past, and a few others in the future. You may want to title them with "future" or "past" for ease of testing.
- [x] Create a page, title it, save, and enter layout builder.
- [x] Add a block of type "View"
- [x] Fill out the Administrative label and select "Events" under "I want to show"
- [x] Verify that when "Events" is selected, a new filter option becomes available lower on the form to filter when the event takes place.
Screenshot 2023-05-31 at 6 12 38 PM

- [x] Make sure "Future" is selected under "Event Time Period"
- [x] Add the block.
- [x] Verify that the results of the view show only future events.
Screenshot 2023-05-31 at 6 13 24 PM

- [x] Configure the block and change the Time Period to "Past Events" and update the block.
- [x] Verify that the results change to only show past events.
- [x] Finally, change the option to "All Events" and verify that all events are shown.

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/field.field.block_content.view.field_view_params.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/field.field.block_content.view.field_view_params.yml
@@ -18,8 +18,8 @@ translatable: false
 default_value:
   -
     group_params:
-      params: '{"view_mode":"card","filters":{"types":["post"],"terms_include":null,"terms_exclude":null},"operator":"+","sort_by":null,"display":null,"limit":0}'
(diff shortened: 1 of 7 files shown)


dblanken-yale (commented):
Great job!  I was able to follow the tests and the events were showing as expected.  All good here, but will defer to @nJim in case I missed anything.

dblanken-yale:
I forgot to add: and only published events were showing.  Great job again!

nJim (approved):
This works perfectly. I still would like to create an entity property to run this calculation for us, but I have not had the time. Thank you for just running with it!

PR #296, opened 2023-06-01
YALB-1246: Bug: Linkit widget broken

codechefmarc:
### Description of work
- Removes Linkit from URL fields for: CTA Banner, Content Spotlight, Grand Hero, Quick Links, Event Content Type, Callout, and Custom Cards
- Uninstalls Linkit module

### Functional testing steps:
- [x] Add an event to the site
- [x] In the Call To Action URL, start typing and choose an existing piece of content that is suggested
- [x] Verify that once selected, the drop down goes away and the title of the selected content is filled in the URL field.
- [x] Verify that after saving, the internal path (like `/node/2`) is replaced by the path alias.
- [x] Also verify that on the front-end, the alert link is the path alias, not the internal Drupal link.

Screenshot 2023-06-01 at 9 21 56 AM

- [x] Verify that the following components also use the Drupal link field instead of Linkit: CTA Banner, Content Spotlight, Grand Hero, Quick Links, Event Content Type, Callout, and Custom Cards


dblanken-yale (commented):
Nice job!  They all worked for me.

nJim (dismissed):
Everything looks good but you are missing one item. Linkit is currently used here too web/profiles/custom/yalesites_profile/modules/custom/ys_alert/src/Form/AlertSettings.php

codechefmarc:
@nJim - OK, this one is ready for re-review. Changes made:
1. Re-enabled Linkit
2. Updated Linkit to 6.0@RC
3. Linkit is still in use for the Alert Settings form - tested and it works again with the new version
4. Linkit is still disabled on all components and instead uses Drupal's core autocomplete for links
5. Linkit is enabled for WYSIWYG - I figured that would be a better experience for those areas.

codechefmarc:
@dblanken-yale - Since Jim is out at Drupalcon, if you can give this one more look and then if approved, I'll merge it in. Thanks!

dblanken-yale:
@codechefmarc While the admin side does work to select the CTA URL on a new event, now when I click on the link generated after it has been added, it is pointing to `entity:node/7` and I get the message `Safari cannot open the page because the address is invalid`.  Chrome doesn't follow the link at all.  :(

dblanken-yale:
On the alert form, the LinkIt drop down seems to work fine, and results in a URL the browser can follow.

dblanken-yale:
WYSIWYG link works great and links to a proper URL.

Although I did notice that in the editor, it links to the URL alias, while in the Alert form, it's linking to the node number.  Still working though in both cases as it goes to the right page.

dblanken-yale:
All of the other items mentioned are resolving to the right URL, so it's only the event CTA that seems to be not working right when generating the URL to go to.  I think if we can figure that out, it'll be ready to go!

dblanken-yale:
As far as the event CTA URL goes, I think I might be the culprit? :(

In my last ticket regarding the add to calendar link and cleanup in Atomic, I used the `node.field_event_cta.uri` to pass to the component as content was not available.  I wonder if that is not translating into an actual URL while the other blocks are.  When I did that ticket I was testing against external links, like `https://google.com`.

It seems hacky, but adding the following lines to `node--event--full.html.twig` in Atomic and setting  `event_meta__cta_primary__href` to the new `cta_url` variable does seem to work:

```twig
{% set cta_url = node.field_event_cta.uri %}
{% if node.field_event_cta.uri starts with 'entity:' %}
  {% set cta_url = node.field_event_cta.uri|replace({'entity:': '/'}) %}
{% endif %}
```

Is there a better reference than `uri` that would successfully convert the URL?  I figure the other blocks are doing it right, but they’re not using the `node` object to get their data.  I didn't see anything else when using `dump` that would yield a better URL.

dblanken-yale:
As far as the scope of this ticket goes, I do think you've done what was asked.  I'm going to submit a new bug ticket on the event link if that's ok.  So I think this is good to go, unless you can think of a way to convert that reference before it gets to the twig file using a preprocess or something.

dblanken-yale (approved):

codechefmarc:
> All of the other items mentioned are resolving to the right URL, so it's only the event CTA that seems to be not working right when generating the URL to go to. I think if we can figure that out, it'll be ready to go!

I'll look into why the event CTA isn't working an will let you know. Good catch.

codechefmarc:
@dblanken-yale - I think I figured it out for your Atomic template for the event full display. Usually within templates there is the `content` variable and we use that to pull from for all of the fields. I see one in `node--event--full.html.twig` but it's not at all in the right format that I'm used to - I think because it's using layout builder? Anyways, if it was working, we'd use ` content.field_link.0['#url_title']` to get the path alias URL (see `block--inline-block--cta-banner.html.twig`)

So, in this case, I see you are using the `node` variable so I was able to figure out that you can get the path alias via: `node.field_event_cta.0.getUrl().toString()` so go ahead and try that in Atomic and let me know if that works for you.

One last thing for me to change is to get the path alias into the Alert, and once that is done, I'll let you know. Thanks for your great feedback and testing!

codechefmarc:
@dblanken-yale - I know you're out this week, but wanted to let you know the alert link will now show the path alias to an internal link instead of the internal link itself. I'll ask @codeb-a to do a quick review of this one and then we'll ship it. Thanks!

codechefmarc:
@codeb-a - This one has been revised for the most part, but I did make some changes and since some of the folks are out this week, I'd love your eyes on it before I merge. Thank you!

codeb-a (approved):
Looks good!

PR #298, opened 2023-06-01
YALB-1294: Install and configure Bugherd module for testing

vinmassaro:
### Description of work
- Adds and configures Bugherd module with Yalesites Upgrade Project ID
- Applies [patch](https://www.drupal.org/project/bugherd/issues/3364305) to allow anonymous users to fill in the email field when it is required

### Functional testing steps:
:warning: Disable any ad blocking extensions in your browser (Adblock Plus, uBlock Origin, etc)
- [ ] As an anonymous user, load a page and see the "Send feedback" button at the lower right
- [ ] As an authenticated user, load a page and see the "Send feedback" button at the lower right. When submitting, your email address should be pre-populated


nJim (approved):
Everything looks good on the Drupal side of the house.

PR #303, opened 2023-06-07
YALB-608, YALB-731, YALB-1225, YALB-1236: Performance Local vs Production

codechefmarc:
## Performance Local vs Production

- [YALB-608](https://yaleits.atlassian.net/browse/YALB-608)
- [YALB-731](https://yaleits.atlassian.net/browse/YALB-731)
- [YALB-1225](https://yaleits.atlassian.net/browse/YALB-1225)
- [YALB-1236](https://yaleits.atlassian.net/browse/YALB-1236)

### Description of work
- Enables config split for local and production environments, allows for different modules to be enabled/disabled per environment
- For local development: adds a local.settings.php file during `npm run setup` which uses config split to enable/disable modules, loads `theming.services.yml` for twig debugging and disabling of cache
- For production (and multidevs on Pantheon): enables caching, disables twig debug, enables Pantheon Advanced Page Cache, disables devel and devel kint modules

### Functional testing steps:
- [x] This requires testing on both the Pantheon mulidev (https://pr-303-yalesites-platform.pantheonsite.io/) and local environment to compare changes

## Multidev (Mimics production)

- [x] Login to the multidev and visit the modules page at https://pr-303-yalesites-platform.pantheonsite.io/admin/modules
- [x] Verify that the following modules are disabled: Devel and Devel Kint Extas
- [x] Verify that the following module is enabled: Pantheon Advanced Page Cache
- [x] View the front-end of the site in an incognito window as a logged out user
- [x] Inspect the DOM and verify that there are no twig debugging comments
Screenshot 2023-06-08 at 1 14 34 PM

- [x] Still in the inspector, click on the "Network tab". Refresh the page, then click on `pr-303-yalesites-platform.pantheonsite.io` at the top and look at the "Response Headers" -> "Cache-Control"
- [x] Verify that the `max-age` is set to "86400"
Screenshot 2023-06-08 at 1 12 51 PM

## Local

- [x] Pull down a clean copy of the `yalesites-project` repo
- [x] Before running `npm run setup`, checkout this branch `YALB-608--config-split`
- [x] Run `npm run setup`
- [x] Login to the local site and visit the modules page at /admin/modules
- [x] Verify that the following modules are enabled: Devel and Devel Kint Extas
- [x] Verify that the following module is disabled: Pantheon Advanced Page Cache
- [x] View the front-end of the site in an incognito window as a logged out user
- [x] Inspect the DOM and verify that there are twig debugging comments in the DOM
[...]

PR #304, opened 2023-06-08
YALB-1241: Deploy Siteimprove analytics JavaScript

codechefmarc:
### Description of work
- Adds Siteimprove Javascript to production environments only
- This PR contains commits from [PR-303](https://github.com/yalesites-org/yalesites-project/pull/303)

### Functional testing steps:
- [x] Testing requires local and multidev (https://pr-304-yalesites-platform.pantheonsite.io/)
- [x] On multidev, load a page and inspect the DOM
- [x] Search for "siteimprove" and verify that the Siteimprove javascript is loading near the end of the DOM close to the closing body tag: `<script src="https://siteimproveanalytics.com/js/siteanalyze_66356571.js" async></script>`
- [x] For local testing: pull down this repo
- [x] Change into this branch (`YALB-1241--siteimprove`)
- [x] Run `npm run setup`
- [x] Once it has finished building, visit the site, and inspect the DOM
- [x] Search for "siteimprove" and verify that the Siteimprove javascript is _not_ loading at all. We don't want this to load locally.


dblanken-yale (approved):
Functions great!  I don't have SiteImprove locally!

I did notice that in the `theming.services.yml`, it states a comment saying to enable it to follow the example file that doesn't seem to exist.  Instead we're just using the local settings if it exists, correct?  Would it be worth removing those comments since the example file isn't located near it and initially threw me off as I was trying to figure out what was enabling debug in my local environment?

Passing as this work is great and doing what it should!

https://github.com/yalesites-org/yalesites-project/blob/0924e316f61ef2c700486b394824204b9fe45f25/web/sites/theming.services.yml#L3-L4

PR #308, opened 2023-06-09
YALB-1235: Media: Configure image toolkit/settings

codechefmarc:
### Description of work
- Adds ImageMagick module
- Switches Image toolkit to use ImageMagick and set quality to 100%

### Functional testing steps:
- [x] Login to the site and go to `/admin/config/media/image-toolkit`
- [x] Verify that ImageMagick is set as the toolkit, and the image quality is set to 100%
- [x] Create a page with some components that include images - include images from within the media library and uploaded
- [x] View the page and verify that images output correctly.


dblanken-yale (approved):
It works well!  Nice job!

The only oddity I saw was with the image I uploaded and used, but this could be a learning moment for me as to how Drupal deals with uploaded images.  I uploaded a 83kb jpg, but when I downloaded the same image after adding it to a block, it was 206kb.  Would Drupal be using ImageMagick flags that would cause it to be larger in size?

codechefmarc:
@dblanken-yale - if you download the image from where it shows on the site (like from within a component), then an image style was applied to make the image both dimensionally smaller (if it needs it) and file size (with ImageMagick) so, yes, it'll be a smaller file size that way. If you visit the media library, you can get the link to the original file which is stored as well.

Thanks for testing, I'll merge this one in too.

PR #319, opened 2023-06-21
YALB-579: Improve homepage settings controls

alan-scrimenti:
### Description of work
- Replaced homepage text field, that is used to set the node of the homepage, with an autocomplete field
- Used Alerts setting form autocomplete field element as an example
- Note: Could not get it to work without setting attribute 'data-linkit-widget-title-autofill-enabled' => 'false', but was told the Alert setting autocomplete works without setting this?

### Functional testing steps:
- [ ] Open the settings page at '/admin/yalesites/settings'
- [ ] In the 'Front page' field, start typing the Title of the page you would like to set as the front page. Matching pages will be displayed.
- [ ] Once you select a page, the field should be populated with the node address of that page.
- [ ]  Save the settings page and verify that page is now the new homepage.

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Form/SiteSettingsForm.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Form/SiteSettingsForm.php
@@ -110,10 +110,17 @@ class SiteSettingsForm extends ConfigFormBase {
     ];
 
     $form['site_page_front'] = [
-      '#type' => 'textfield',
-      '#description' => $this->t("Specify a relative URL to display as the front page."),
+      '#type' => 'linkit',
+      '#description' => $this->t("Specify a relative URL or autocomplete to display as the front page."),
       '#title' => $this->t('Front page'),
+      '#autocomplete_route_name' => 'linkit.autocomplete',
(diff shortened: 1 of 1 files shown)


codechefmarc (changes requested):
I got this to work just fine, so you're on the right path. I don't think you need that `#attributes`, though - I removed it and it still worked for me. One of the things I did was to merge in the `develop` branch so you'll be up to date with the rest of the codebase and you'll get the newer version of linkit included in that. So, here are my thoughts for you:

1. Merge develop into your branch (this should also fix the build error in the PR)
2. Run `npm run build-with-assets` to rebuild your local with the latest changes (including the newer version of linkit)
3. Remove the `#attributes`
4. Optional, but if you want to follow the code in `web/profiles/custom/yalesites_profile/modules/custom/ys_alert/src/Form/AlertSettings.php` to save the path alias instead of the Drupal internal link ("/my-cool-page" instead of "/node/7", for example). I'd be happy to jump on a Zoom call to go over what I did there to do that. There's an open issue on the Linkit issue queue where custom fields don't get the path alias which is super strange and (if I get some time) I'd like to create a patch for that one. Here is that issue: https://www.drupal.org/project/linkit/issues/3274952

PR #320, opened 2023-06-21
YALB-1288: Bug: Search no results message (BE)

codechefmarc:
### Description of work
- Fixes bug where no results text wasn't showing properly.
- No results text will now show if search terms are entered and no results are found.
- New search text will now show if no search terms are entered. (i.e. on `/search` with no keywords query parameter)
- Requires [Atomic PR-136](https://github.com/yalesites-org/atomic/pull/136)

### Functional testing steps:
- [x] Use the top right search box and enter a search term that would produce no results (such as "asdfg") and hit Enter
- [x] Verify that the no results text shows up: "There were no results for your search."
- [x] Edit the URL of the page to simply `/search` and load the search page with no query parameters
- [x] Verify the new search text shows up: "Enter some search terms and then click search."

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/views.view.search.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/views.view.search.yml
@@ -115,7 +115,7 @@ display:
             offset: false
             offset_label: Offset
       exposed_form:
-        type: input_required
+        type: basic
         options:
           submit_button: Search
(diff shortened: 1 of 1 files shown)


dblanken-yale (approved):
Works great!  Nice job!  Blows me aways sometimes seeing the changes needed to get such good functionality.

PR #324, opened 2023-06-23
fix(YALB-1189): Site admins should not be able to configure site-wide…

codechefmarc:
### Description of work
- Removes the ability for site admin role to see pencil icon and edit system-level blocks.
- System-level blocks are still editable by platform admins.
- Site admins can still edit layout builder blocks.

### Functional testing steps:
- [x] Login to the site as a site administrator role in an incognito window and also have either user 1 or platform admin role in another browser window. Note: if using CAS, double check the role of the testing user (with user 1) AFTER logging in because CAS Can add in the platform admin role and mess up testing.
- [x] Visit the front-end of the site
- [x] Hover over the alert banner and verify that the pencil icon on the right-hand side is missing and there is no way to configure / remove the alert banner as a block.

You should NOT see this when hovering as a site admin:
Screenshot 2023-06-23 at 2 09 09 PM

- [x] Visit the `/search` page as a site administrator and hover over the title "Search" and verify that there is no pencil icon on the right-hand side.
- [x] Verify that as user 1 or platform admin role, these pencil icons do show up.
- [x] Add a page, title, save, and enter layout builder.
- [x] Verify that you can add and edit layout builder blocks as a site administrator.

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/user.role.site_admin.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/user.role.site_admin.yml
@@ -18,7 +18,6 @@ dependencies:
     - taxonomy.vocabulary.post_category
     - taxonomy.vocabulary.tags
   module:
-    - block
     - contextual
     - editoria11y
     - file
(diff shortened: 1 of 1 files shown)


dblanken-yale (approved):
Great job!  Works as intended!

PR #332, opened 2023-06-28
YALB-1295: Bug: Posts display date of creation and not Publish Date

codechefmarc:
### Description of work
- Changes post sorting for cards and lists to use "Publish Date" field instead of creation date.
- Also has an Atomic piece in [PR-142](https://github.com/yalesites-org/atomic/pull/142) that changes the visual display of the post date (in cards, lists, and default node view) from the creation date to the date in the "Publish Date" field.

### Functional testing steps:
- [x] Create 3 posts on the site with different publish dates
- [x] On a page, in layout builder, add a block of type "Post feed" and save the layout
- [x] Verify that on the page in the post feed, the posts are ordered in reverse chronological order and that the date listed in the feed is the published date set, not the creation date.

Screenshot 2023-06-28 at 3 46 02 PM

- [x] Click through to one of the posts and verify that on the detail page, the post date is the published date and not the creation date.

Screenshot 2023-06-28 at 3 47 07 PM

- [x] Go back to the page and re-enter layout builder
- [x] Add a block of type "View"
- [x] Use the following parameters: "Posts" as "Post Card Grid" sorted by "Publish Date - older first".
- [x] Save the layout
- [x] Verify that the sort order of the posts in the view are from oldest to newest and that the date listed in the view is the published date set, not the creation date.

Screenshot 2023-06-28 at 3 49 41 PM

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/pathauto.pattern.post.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/pathauto.pattern.post.yml
@@ -7,12 +7,12 @@ dependencies:
 id: post
 label: Post
 type: 'canonical_entities:node'
-pattern: '[yale:post_landing_page_path]/[node:created:custom:Y-m-d]-[node:title]'
+pattern: '[yale:post_landing_page_path]/[node:field_publish_date:date:custom:Y-m-d]-[node:title]'
 selection_criteria:
-  8a744808-8c9b-41fd-bb78-ada30c5622e8:
+  d2b465dc-45d7-44e9-bd6f-a0b2794f2767:
     id: 'entity_bundle:node'
     negate: false
-    uuid: 8a744808-8c9b-41fd-bb78-ada30c5622e8
+    uuid: d2b465dc-45d7-44e9-bd6f-a0b2794f2767
     context_mapping:
       node: node
     bundles:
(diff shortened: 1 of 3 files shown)


dblanken-yale (approved):
I was able to get through all steps without issue.  I did have one question: if two posts have the same publish date, does it then sort by anything as a backup?  Like title, or creation date at that point?

Works great though!  Nice job.

joetower:
@codechefmarc I have one question, this looks great :+1: 

Should we also be changing the `url` structure? I noticed after creating 2 posts in May, today, June 29, the url still assigns `06-29`.

https://github.com/yalesites-org/yalesites-project/assets/366413/98157319-77df-47ca-b75a-0ac70da21c95

codechefmarc:
@dblanken-yale - It doesn't currently have any other sorts after the publish date, so I think Drupal uses the node ID, but certainly we can add whatever other criteria we want. I'll let Yale decide that and I'm happy to implement it.

@joetower - Good catch! Wow, I should have thought about that. I'll update that and re-submit for testing when finished.

codechefmarc:
@joetower - OK, changed the path for posts to the publish date. It won't change existing unless you edit and save. New ones will get the new path. Also, cool, is that older posts created will get a redirect to the new path. Tagged you for testing and then if all good, I'll merge this one in. Thanks!

joetower (approved):
@codechefmarc Works wonderfully, great work! :+1: :shipit:

PR #335, opened 2023-07-07
YALB-1290: Bug: Grand Hero Text alignment

codechefmarc:
### Description of work
- Title is misleading. Another issue on this ticket was brought up by Ashley: "Text CTAs are also too close to the text (in action banners)."
- Heading HTML text format will no longer strip out `<p>` tags
- Fixes the issue with the text being too close to the CTA in action banners
- Tested all other components and all looked fine except for
- Quote component attribution broke - we need to address this in the component library CSS is my guess

### Functional testing steps:
- [ ] Create a page, title, save and enter layout builder
- [ ] Add a component of type "Action Banner"
- [ ] Fill out the heading, content, image, and link with dummy text
- [ ] Add the block and save the layout
- [ ] Reduce browser width to 990px or less
- [ ] Verify that the text and button have sufficient padding around each other, see screenshot

Screenshot 2023-07-10 at 9 46 26 AM

### After Joe has made a CSS fix (probably in Component Library):
- [ ] Add a component of type Quote to the page
- [ ] Fill out all of the information, save the block, and save the layout
- [ ] Verify that the quote attribution has the dash next to the attribution, not above it. **Current screenshot is the broken quote**:

Screenshot 2023-07-07 at 1 27 45 PM

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/filter.format.heading_html.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/filter.format.heading_html.yml
@@ -14,7 +14,7 @@ filters:
     status: true
     weight: -10
     settings:
-      allowed_html: '<em> <strong>'
+      allowed_html: '<em> <strong><p>'
       filter_html_help: false
       filter_html_nofollow: false
   improve_line_breaks_filter:


joetower:
@codechefmarc I pushed up a fix: https://github.com/yalesites-org/component-library-twig/pull/264 and rebuilt the multidev in this PR. 
This should be ready to test here: https://pr-335-yalesites-platform.pantheonsite.io/components

codechefmarc:
@joetower - Easy review as it's just the main repo with both fixes in place now: Action banner padding and quote fix you just put in. Thank you!

nJim (approved):
Testing on my local and it seems to be working. I do not see a quote issue, thought I have the most up to date component library which was released since this conversation began. I'm going to consider this work approved but ask the client to double check this test plan once we got out to a dev environment. Thank you for the coordination needed to make a small change,

PR #337, opened 2023-07-10
YALB-1256: Consistant 'add' label for reference fields

nJim:
### Description of work
- Update entity reference widgets to use button variation. This applies to the 'custom card' and 'gallery' blocks. We want to use the same controls as other blocks.

### Functional testing steps:
- [ ] Login and create a page.
- [ ] Add a 'custom card' block. Verify that the bottom of the content-add module it includes a link to 'add custom card' but does not include the suffix "to Cards". This is consistant with our other blocks like image grid, tabs, and accordions.

For reference, this controls originally looked like this on two of the block types:
-old
But are now simply a button:
-new


nJim:
Functionally approved by Alyssa in a meeting.

codechefmarc (approved):
Works as expected. Approved.

PR #343, opened 2023-07-11
YALB-1448: Bug - Media library wont allow item selections

nJim:
### Description of work
- Fixes a bug where site admins can not select media in the media library browser when adding content in a custom block. This is related to a Drupal Core bug that appears in specific configuration of using layout overrides and custom content blocks.
- To resolve this issue, we must give users explicit permission to add the block. This is strange as users already have these permissions, but somehow this resolved the issues. This is made possible by enabling the module: https://www.drupal.org/project/block_content_permissions
- Tech debt: Custom blocs are treated as content in a way consistant with nodes starting in Drupal 10.1. Once we upgrade, we can disable this module and use the new core Drupal permissions.

### Functional testing steps:
- [x] Login to Drupal as a site admin
- [x] Edit the layout of a page and add a 'Video' block. Verify that the user can add an existing video from the media library. Previously this silently failed as there was an 'access denied' on the ajax process.
- [x] Conduct some sanity checks on other blocks including media focused ones, including a grand hero, imaage, and divider block.


nJim:
Functional testing approved by Alyssa in a meeting.

codechefmarc (approved):
Works as expected. Approved!

PR #359, opened 2023-07-20
Bug: CAS user email addresses not being updated on login

vinmassaro:
### Description of work
- When a user is added manually as a CAS user, a dummy email address is required to create the account. This was supposed to be updated with the user's real email address from a CAS Attribute after login, but is not working. This adds an additional configuration to use the mail CAS Attribute.
- Also adds `noreply.yale.edu` as the default email hostname when creating a new user manually, so end users don't need to fill it in. The email on the user account gets updated from CAS Attributes on the user's next CAS login.

### Functional testing steps:
- [ ] Log in as user 1, create a new CAS user with your netID and set the email to username@noreply.yale.edu
- [ ] In a new browser session, log in through CAS with newly created user and edit the user account, verifying the email address has been updated from username@noreply.yale.edu

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/cas.settings.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/cas.settings.yml
@@ -29,8 +29,8 @@ forced_login:
 user_accounts:
   prevent_normal_login: true
   auto_register: true
-  email_assignment_strategy: 1
-  email_hostname: ''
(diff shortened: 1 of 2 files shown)


nJim (approved):
Approved. This works for me. I wonder if we should set these as install-configs in the yale_cas module. Would be good for other's who want to use this service.

PR #367, opened 2023-08-02
YALB-1370: Bug: Search results for "Bread" and "Crumb"

dblanken-yale:
### Description of work
- Add and enable [Search API HTML Element Filter extension](https://www.drupal.org/project/search_api_html_element_filter)
  - Add `.visually-hidden` as a selector to remove before indexing
- Enable `Tokenizer` and `HTML Filter`

### Functional testing steps:
- [ ] Add a new Page
- [ ] Search for "Bread" or "crumb" or "Breadcrumb"
- [ ] Verify no results are found (unless they were actual content in a page)
- [ ] Edit the new Page
- [ ] Save the new Page
- [ ] Search for "Bread" or "crumb" or "Breadcrumb"
- [ ] Verify no results are found (unless they were actual content in a page)
- [ ] Log out of the site
- [ ] Search for "Bread" or "crumb" or "Breadcrumb"
- [ ] Verify no results are found (unless they were actual content in a page)
- [ ] Search for text that you do know exists on a page -- remember that the menu name in some cases does not match the title of the page it's linked to
- [ ] Verify results are properly found for the search term

NOTE: I ran it here, but existing sites will need to re-index for it to take the new processors.  Either go into the admin side and clear and re-index, or run: `terminus drush <site> -- search-api:index`

### DIFF
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -73,6 +73,7 @@
     "drupal/redirect": "^1.7",
     "drupal/search_api": "^1.25",
     "drupal/search_api_exclude": "^2.0",
+    "drupal/search_api_html_element_filter": "^1.0",
     "drupal/section_library": "^1.1",
     "drupal/selective_better_exposed_filters": "^3.0@beta",
     "drupal/simple_sitemap": "^4.1",
@@ -81,9 +82,9 @@
     "drupal/upgrade_status": "^3.18",
     "drupal/webform": "^6.2@beta",
     "drupal/wingsuit_companion": "^2.0@RC",
-    "yalesites-org/atomic": "1.16.3",
     "oomphinc/composer-installers-extender": "^2.0",
     "wikimedia/composer-merge-plugin": "^2.1",
+    "yalesites-org/atomic": "1.16.3",
     "yalesites-org/yale_cas": "^1.0"
   },
   "minimum-stability": "dev",
--- a/web/profiles/custom/yalesites_profile/config/sync/core.extension.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.extension.yml
@@ -101,6 +101,7 @@ module:
   search_api: 0
   search_api_db: 0
   search_api_exclude: 0
+  search_api_html_element_filter: 0
   section_library: 0
   simple_sitemap: 0
   smart_date: 0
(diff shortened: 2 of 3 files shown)


codechefmarc (changes requested):
Hi Dave, tried testing this one out and I wasn't able to get any search results at all - for any term. So, I tried rebuilding the index and it said "-34 items could not be indexed. Check the logs for details." Checking the logs and there are a bunch for the `ys_views_basic` module that I wrote and so I'm wondering if that is holding up indexing? Are you able to search for other keywords and get results? I'll see if I can fix those Views Basic errors on this branch and try again.

codechefmarc:
@dblanken-yale - Update for you. I did perform the fixes to `ys_views_basic` and that removed the errors from the logs. You can see the changes here (https://github.com/yalesites-org/yalesites-project/tree/views-basic-fixes) But, I was still getting errors and things not indexing. So, I turned off your `ys_index` module and things started to index again.

I was going to ask if your module is needed? I did a search on "bread" after turning off your module (but the other module you found `search_api_html_element_filter` is still on) and no results were found for "crumb" and 2 results for "bread" but they were legitimate - "breadth" was the word it found. So, I don't think your module is required? Let me know your thoughts. Thanks!

dblanken-yale:
@codechefmarc Doh, I apologize for not completing the steps for the reason I was making it.  Yes, re-indexing does fix the issue, however, when you then modify or create a new node, breadcrumb becomes searchable again for that node.  My hope was to inject a refresh of the index like we would do in the admin side on save, but it sounds like it's doing more destructive things than I had hoped.

I'll update the steps above to include this case.  I am all for removing this if it's not needed, but I will also take a look with your changes to see if maybe that was the culprit and I missed it.

dblanken-yale:
@codechefmarc Yes, disabling the module results in breadcrumb being searchable for the last saved item (at least, locally it does).  It's strange because I've attempted to create hook logging and it is calling the search_api hooks on save.  I'm not sure why breadcrumb would be there on save, but not there when we manually decide to reindex.

I'll keep looking into this and see what I can come up with.  Thanks for providing those changes to views though!

dblanken-yale:
After spending half the day debugging into search_api, I think I found the solution.  The Stemmer was tokenizing the original text for some reason.  In its description, it states that using the tokenizer before the Stemmer is recommended, so combining the addition of the tokenizer with the HTML filter results in a reliable search.

Looking in the db, it was indexing all kinds of things we wouldn't want, like the menu items.  Their respective pages will be indexed, but the pages that reference that page won't be since it's in the menu.

dblanken-yale:
@codechefmarc Whenever you can, give it another try.  Hopefully you'll have a MUCH better time with search.  :D

codechefmarc (approved):
Beautiful - works great! I tested adding a new node as well and nothing showed up. Also tested re-indexing and no issues there. Approved!

PR #375, opened 2023-08-08
ChosenJS composer fix on develop for Multidev

dblanken-yale:
DO NOT MERGE


dblanken-yale:
Tested and working.  Closing so we can merge with master, not develop.

PR #376, opened 2023-08-09
YALB-1512: Docs: Verify and update developer setup documentation

dblanken-yale:
## [YALB-1512: Docs: Verify and update developer setup documentation](https://yaleits.atlassian.net/browse/YALB-1512)

For more senior developers, this allows an at a glance look on the requirements to have set up, while allowing a deeper dive into the document to find specifics.

### Description of work
- Adds documentation geared toward a more mid/senior level developer
- Keep existing documentation in tact so that questions regarding higher level explanation can be answered by digging deeper

[Branch markdown file with the setup information](https://github.com/yalesites-org/yalesites-project/blob/YALB-1512-setup-docs/docs/setup.md)

### Functional testing steps:
- [ ] Read through
- [ ] Too verbose?  Can you easily identify ways to circumvent the verbosity?
- [ ] Are there parts that you feel are irrelevant to a mid/senior level Drupal developer?
- [ ] Is there any piece missing?
- [ ] Does it properly convey how the different repositories interact and exist for?

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -92,42 +92,13 @@ To clone the project, the above requirements must be met first.  If you have not
 
 `git clone git@github.com:yaleitsites-org/yalesites-project.git`
 
-## Project setup script
+## Project setup
 
 This repository contains a custom Pantheon Upstream used to create and manage every site on the YaleSites platform. Out of the box this project is not connected to an individual Drupal site. To contribute to this project, we need to connect the local development environment to a Drupal site to leverage the site's files and database.
 
 ```bash
 # Executing the setup script will prepare the local development environment.
 npm run setup
-```
-
-## Alternative manual setup instructions
-
-A detailed explanation of the setup script appears below.
-
-```bash
-# Step 1: Clone this repository and enter the project directory.
-git clone git@github.com:yalesites-org/yalesites-project.git
-cd yalesites-project
-
-# Step 2: Ignore the composer.lock file on local dev only.
-grep -qxF 'composer.lock' .git/info/exclude || echo 'composer.lock' >> .git/info/exclude
-```
-
-The Pantheon Lando recipe can connect the local development environment to a remotely hosted site. Connection information is stored in a local Lando settings file and includes a Pantheon site name and UUID. While we can connect to any site running this upstream, a particular integration environment has been provisioned for engineers working on this repository or the associated YaleSites profile. Connection information for the integration environment is stored in the example Lando local file.
-
-```bash
-# Step 3: Create a local Lando settings file to connect to the external environment.
-cp .lando.local.example.yml .lando.local.yml
-```
-
-Starting Lando will provision the containers required to run a local development environment. The `pull` command can then sync the database and files from a Pantheon-hosted site with the new local site.
-
-```bash
-# Step 4: Start Lando and import the remote files and database.
-lando start
-lando pull --database=dev --files=dev --code=none
-lando drush cr
 ```
 
 Visit the local dev site [https://yalesites-project.lndo.site/](https://yalesites-project.lndo.site/) or run `lando drush uli` to obtain a login link.
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -18,6 +18,17 @@ __NOTE: Documentation assumes you are using MacOS or a Linux environment.__
    npm run setup
 ```
 10. [Project Commands](#project-commands)
+   
+## Repository Links
+
+Repositories used to make the platform:
+
+* [Yalesites Project](https://github.com/yalesites-org/yalesites-project): Drupal Platform Site (This repo)
+  * [Yalsites Project Profile modules](https://github.com/yalesites-org/yalesites-project/tree/develop/web/profiles/custom/yalesites_profile/modules/custom)
+* [Atomic Theme](https://github.com/yalesites-org/atomic): Atomic Drupal Theme
+* [Component Library Twig](https://github.com/yalesites-org/component-library-twig): Component Library
+* [Tokens](https://github.com/yalesites-org/tokens): Style tokens from Figma
+* [ESLint Config and Other Formatting](https://github.com/yalesites-org/eslint-config-and-other-formatting): Reusable Linting/Formatting
 
 ## Requirement Details
 
@@ -66,7 +77,7 @@ A detailed explanation of the setup script appears below.
 
 ```bash
 # Step 1: Clone this repository and enter the project directory.
-git clone https://github.com/yalesites-org/yalesites-project.git
+git clone git@github.com:yalesites-org/yalesites-project.git
 cd yalesites-project
 
 # Step 2: Ignore the composer.lock file on local dev only.
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -1,6 +1,6 @@
 # Local Development Environment
 
-__NOTE: Documentation assumes you are using MacOS with at least 8GB of memory and an Intel, M1, or M2 processor__
+__NOTE: Documentation assumes you are using MacOS with at least 8GB of memory on an Intel, M1, or M2 processor__
 
 :clap: Thank you for considering to contribute to the YaleSites Platform!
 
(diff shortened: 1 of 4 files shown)


myvaughn (approved):
I'm going to defer to the others on this. It looks good to me and appropriate for an experienced developer.

docs/setup.md
```diff
@@ -1,10 +1,52 @@
...
+__NOTE: Documentation assumes you are using MacOS with at least 8GB of memory on an Intel, M1, or M2 processor__
+
+:clap: Thank you for considering to contribute to the YaleSites Platform!
+
+## Requirements & Setup
+
+1. [Contact YaleSites](mailto:yalesites@yale.edu?subject=Developer%20Access%20Request) for access to resources such as Pantheon, Figma designs, etc.
+2. [GitHub](#github)
+   1. SSH authorization key
+   2. [Github Package Personal Access Token](https://github.com/yalesites-org/component-library-twig) in environment variable `YALESITES_BUILD_TOKEN`
```
codechefmarc, line 12: Why does this one link to the Component Library? Is that what requires the build token? I thought that the token was required for the main repo too? (I forget!)

docs/setup.md
```diff
@@ -21,15 +63,11 @@ This project supports development with Lando using the Pantheon recipe. This pro
...
 ### Additional tools
 
-- [Composer](https://getcomposer.org/download/): PHP package manager. Version 2.x.
-- [NVM](https://github.com/nvm-sh/nvm#install--update-script): Recommended for installing and switching between Node versions.
-- Node.js (>=8.0). We recommend installing via NVM.
+- [Composer](https://getcomposer.org/download/): PHP package manager. Version 2.x. (Can use lando instead if you prefer)
+- [NVM](https://github.com/nvm-sh/nvm#install--update-script): Node Version Manager
+- Node.js (>=8.0,<11.0). Via NVM.
 
 ## Project setup script
```
codechefmarc, line 73: For this area, I'd also add something like "If you haven't already cloned down the repo..." and then then command to clone down.

Do we need the "alternate manual setup instructions"? I like the idea of explaining what each of the scripts do, including the setup script, but, not sure we want people running the individual commands? Especially if the `setup` script is updated but the docs aren't? I'm not sure about this one so I'd love your input.

docs/setup.md
```diff
@@ -46,7 +84,7 @@ A detailed explanation of the setup script appears below.
 
 ```bash
 # Step 1: Clone this repository and enter the project directory.
-git clone https://github.com/yalesites-org/yalesites-project.git
+git clone git@github.com:yalesites-org/yalesites-project.git
 cd yalesites-project
 
 # Step 2: Ignore the composer.lock file on local dev only.
```
codechefmarc, line 90: "When working on a Yale-managed package" -> "When working on a Yale-managed package such as Atomic or the Component Library Twig"

docs/setup.md
```diff
@@ -100,3 +138,73 @@ git -C web/themes/contrib/atomic ls-remote --get-url
...
+npm run build-with-assets                       # Build/Rebuild with replaced assets
+```
+```bash
+npm run build-with-install                      # Rebuild with fresh imports
+```
+```bash
+npm run local:cl-dev                            # npm link component library with local environment
+```
+```bash
+npm run local:review-with-atomic-and-cl-branch  # Use different branches of each repo
```
codechefmarc, line 161: I'm curious if we should get rid of these scripts / not talk about them because your `git-checkout` script is so good I'm inclined to say we can replace those others with yours. :)

codechefmarc (dismissed):
Looking really good - just a few changes / questions for you that I've commented per line. I'm fine with essentially approving as is, but thought a little discussion would be good too. Let me know your thoughts!

docs/setup.md
dblanken-yale: Yes, that is where the current documentation was to generate the token.  I had thought about duplicating it, but instead linked to it so we wouldn't have to keep it in sync.  I also thought that I should update this to add (see prerequisites) if I continue to link to component library.

Yes it is needed here as well as it uses the component library in the setup.  Do you think it'd be bad to migrate that over to this document and then reference it from the component library README?  That then make this more of the source of truth in setup.

docs/setup.md
dblanken-yale: I agree that the manual isn't technically needed, but was there from before.  I know that Mike V has used the manual before as a way to track down issues that occurred using setup sometimes, so I decided to leave it as is.

I do like the idea of adding something about getting the repo for completeness; that's sort of where I was going toward the top of the document with the quick copy/paste bash commands.

docs/setup.md
dblanken-yale: Love that change!  Much more explicit.  Thank you!

docs/setup.md
dblanken-yale: 😁 That makes me so happy.  I'm happy for us to discuss this and we could talk about retiring them or archiving them?  Maybe we'll wait on that till Jim gets back so we can get his input.  I'd hate to remove any if someone is using it still and prefer it.

docs/setup.md
dblanken-yale: Thinking about it more, I think we may attempt to duplicate how to do this here as it is pretty important to getting up and running, and I know I'd hate being redirected there and then have to realize prerequisites are collapsed.  Thanks for bringing that up and speaking about it.

docs/setup.md
vinmassaro: +1 to remove scripts if they are not being used!

docs/setup.md
```diff
@@ -1,10 +1,69 @@
...
+* [Component Library Twig](https://github.com/yalesites-org/component-library-twig): Component Library
+   * [Deployed Storybook](https://yalesites-org.github.io/component-library-twig)
+* [Tokens](https://github.com/yalesites-org/tokens): Style tokens from Figma used to drive design of the platform and components
+* [ESLint Config and Other Formatting](https://github.com/yalesites-org/eslint-config-and-other-formatting): Reusable Linting/Formatting included in the project
+
+## Requirement Details
+
+### GitHub
+
+While the project can be checked out and run locally without it, one must setup an SSH key on their development machine if they wish to push code on any of the yalesites repositories.
```
vinmassaro, line 44: This should probably say "cloned" instead of "checked out"

dblanken-yale: Good catch.  Done!

docs/setup.md
dblanken-yale: Added, and I feel better when clicking that link now.

docs/setup.md
dblanken-yale: I added a small section for cloning the repo for completeness.

vinmassaro: I agree with @codechefmarc on this one, it is less to maintain and if you really want to run the steps manually, you can break things down by looking at what `npm run setup` is doing.

vinmassaro:
@dblanken-yale Looking good!

You didn't write this part but I noticed something that is minor, but should be corrected:

"This repository contains a custom Pantheon Upstream" should be changed to "This repository is a Pantheon Custom Upstream"

Thanks!

docs/setup.md
dblanken-yale: Sorry @myvaughn; if you need them we can look at the script.  :D

I agree, keeping the manual in sync with the script longer term might be problematic.

docs/setup.md
dblanken-yale: Ok I removed some of them from the docs to instead attempt to use the git-checkout.

dblanken-yale:
@vinmassaro @codechefmarc If you have any time tomorrow, give it one more look over.  If it's ok with both of you, I'll release it so that they can provide it in their communication.  Thanks so much for looking at it and giving this feedback.

docs/setup.md
myvaughn: +1 to remove scripts that aren't being used (outside the scope of this PR), to undo unnecessary nesting of scripts, and to be sure scripts have good comments. Since we're removing the details of what the scripts are doing on the README (which I think is a good idea) we'll want the scripts to be easy to decipher.

PR #381, opened 2023-08-14
Update develop from master

vinmassaro:
### Description of work
- `master` has seen a number of hotfixes applied to it that have not been applied to `develop`. This PR will bring the branches in sync.


codechefmarc:
@vinmassaro - I was trying to test this on the multidev and, strange, I see the build error about `bartik` not having the `core_version_requirement` - I checked my local and it doesn't have it. Also checked Drupal 9.5.10 and it doesn't have it either... so not sure why it is complaining about that. I'm going to try to re-run that job to see if it fixes itself.

vinmassaro:
I’m not sure why CI is failing either, and will try building locally to see if I can reproduce.

codechefmarc:
Looks like it is trying to load Bartik from the `profiles/custom` directory but we don't have Bartik in there... strange!

dblanken-yale:
I ran into a similar issue with Chosen at one point. It was thinking it was at the root but instead was in the profile directory, and could not find bartik, which seemed to be part of core Drupal.

This was the fix I think, but strange that we need it now but did not need it for the hot fix.

https://github.com/yalesites-org/yalesites-project/commit/d7f61c67d618468d8012a0e311642760c6fa23a8

dblanken-yale:
@vinmassaro Sorry I didn't just apply it last night--had the kiddo.  :D Made a branch to merge into this one.  First did it on develop so the multidev could create for us to check.

https://github.com/yalesites-org/yalesites-project/pull/382

vinmassaro:
This PR is cursed, closing.

PR #386, opened 2023-08-16
YALB-1483: Footer: Add LinkedIn setting and link

dblanken-yale:
## [YALB-1483: Footer: Add LinkedIn setting and link](https://yaleits.atlassian.net/browse/YALB-1483)

Linked to: [Component Library PR](https://github.com/yalesites-org/component-library-twig/pull/281)

### Description of work
- Adds LinkedIn as a footer option

### Functional testing steps:
- [x] Log in with your CAS login
- [x] Visit Settings->Footer Settings
- [x] Verify that you see LinkedIn as an option
- [x] Enter a URL for LinkedIn and Save
- [x] Visit a page to verify that the LinkedIn icon appears and is clickable to the location you specified
- [ ] Visit Settings->Footer Settings
- [ ] Delete the URL you created for LinkedIn and Save
- [ ] Visit a page to verify that the icon no longer shows


codechefmarc (changes requested):
Looks like our good friend caching is preventing this from working directly. If I just set it and try it out on the front end, I don't get the new URL. But if I clear cache, it works fine. Probably have to do one of those things we did in one of those pre-process functions so, something like this but you'll have to modify for the specific config, let me know if you need help with that piece: https://github.com/yalesites-org/yalesites-project/blob/9cfa96591f5ad88230748b14b2f69933c3b31180/web/profiles/custom/yalesites_profile/modules/custom/ys_themes/ys_themes.module#L145

As to your question about the color of the LinkedIn logo - I'd check with @miketullo95 and Katy to see what they think.

dblanken-yale:
@codechefmarc Oh darn!  Wonder why it seemed to work for me.  Maybe I did something that let time elapse so long it cleared.  Ugh.  Thanks for pointing it out, I'll see if I can add that cache refresh.

Ya, the linkedin color is fixed.  @joetower is awesome and brought up that the icon already existed, so I didn't need to copy over it.  hahaha. Thanks again, Joe.

dblanken-yale:
@codechefmarc Is there a "Drupal way" for a global location to call that as a static method for us across the codebase?  Just didn't want shotgun surgery if we ever have to change it.  Or do you think it'll be ok?

dblanken-yale:
I guess if we did that, all of our modules would be dependent on it.  Maybe the way we have it is the best for now.

codechefmarc:
@dblanken-yale - Not a bad idea and one worth exploring for sure. Since `ys_core` is sort of our base of many things for all sites, perhaps we could have a function in there that grabs each config that we need to keep track of and run that command once. We'd obviously want to make sure the other modules (like `ys_themes`) are a dependency of `ys_core` but that's easy.

Perhaps put in a tech debt ticket to look into that?

codechefmarc (approved):
Works great now! Approved!

PR #387, opened 2023-08-16
YALB-1519: Add semantic-release to project repo

vinmassaro:
### Description of work
- Adds semantic-release to this repository on pushes to `master`
- Automatically increments yalesites_profile.info.yml version number on new release
- Adds intial `version: 1.0` to yalesites_profile.info.yml so semantic-release can modify it on next release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,23 @@
+name: Create a release
+on:
+  push:
+    branches: master
+env:
+  GH_TOKEN: ${{ secrets.YALESITES_BUILD_TOKEN }}
+jobs:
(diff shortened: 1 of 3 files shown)


vinmassaro:
Work blocked here by outdated dependencies in https://github.com/yalesites-org/eslint-config-and-other-formatting, so I opened up https://github.com/yalesites-org/eslint-config-and-other-formatting/pull/5.

dblanken-yale (commented):
Changes look sound to me.  Interesting to see how the replace plugin makes the changes in the yaml file.  I'll still defer to @nJim but it looks like it would work to me.

PR #391, opened 2023-08-17
YALB-1503: Controls: Style Lever

dblanken-yale:
## [YALB-1503: Controls: Style Lever](https://yaleits.atlassian.net/browse/YALB-1503)

This version was originally created since the code was created in such a way to want radio button inputs for all options.  

Screenshot 2023-08-18 at 2 31 41 PM

### Description of work
- Add animation style options to theme settings
  - Options: minimal|Minimal, artistic|Artistic
- Install by default with a minimal default setting

### Functional testing steps:
- [x] Visit a page
- [x] Go to theme settings
- [x] Scroll toward the bottom of the page
- [x] Verify that `Animation Style` exists with the following options:
  - [x] Minimal (default selected)
  - [x] Artistic
- [x] Select whichever one is not currently selected
- [x] Verify that the newly selected item is highlighted
- [x] Save
- [x] Go back into theme settings
- [x] Verify that the animation style you previously selected is still selected

PR #393, opened 2023-08-18
YALB-1503: Controls: Style Lever (Select Edition)

dblanken-yale:
## [YALB-1503: Controls: Style Lever (Select Edition)](https://yaleits.atlassian.net/browse/YALB-1503)

[Version with radio buttons](https://github.com/yalesites-org/yalesites-project/pull/391)

This version was created to conform to the specs, although the original code was created in such a way to want radio button inputs for all options. I wanted feedback on which to use since using selects causes a bit more code/css.

While this does look good, CSS had to be borrowed from places outside of ys_theme, and the logic to produce different elements had to be made.  It still allows you to define new settings in the array, but if any new HTML elements are needed, they will now need to be created as an `ElementType`, which could become cumbersome.

Screenshot 2023-08-18 at 2 32 06 PM

### Description of work
- Add animation style options to theme settings
  - Options: minimal|Minimal, artistic|Artistic
- Install by default with a minimal default setting
- Added as a select instead of a radio button
- Attempted to add CSS needed to style select in ys_theme

### Functional testing steps:
- [ ] Visit a page
- [ ] Go to theme settings
- [ ] Scroll toward the bottom of the page
- [ ] Verify that `Animation Style` exists with the following options:
  - [ ] Minimal (default selected)
  - [ ] Artistic
- [ ] Select whichever one is not currently selected
- [ ] Verify that the newly selected item is highlighted
- [ ] Save
- [ ] Go back into theme settings
- [ ] Verify that the animation style you previously selected is still selected


dblanken-yale:
Closing as we are using the radio button version.

PR #396, opened 2023-08-22
YALB-1242: Use Pantheon Secrets

vinmassaro:
### Description of work
- Adds [Pantheon Secrets](https://www.drupal.org/project/pantheon_secrets) and [Key](https://www.drupal.org/project/key) modules so that secrets can be sourced directly from Pantheon Secrets Manager
- Adds Key module config overrides to replace items with values sourced from Pantheon
- Removes `ys_secrets` module as a dependency for `ys_captcha` and `ys_mail` modules, setting form overrides directly in `ys_captcha` and `ys_mail`
- Removes logic from `settings.php`

### Functional testing steps:
- [ ] Log into multidev, verify that Key module (`admin/config/system/keys`) config overrides are in place
- [ ] On Mailchimp Transactional settings page, verify key is not visible
- [ ] On reCAPTCHA and reCAPTCHA v3 settings, verify key and secret are not visible

### Todo
- [YALB-1541: Remove ys_secrets](https://yaleits.atlassian.net/browse/YALB-1541)

### DIFF
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -64,6 +64,7 @@
     "drupal/node_revision_delete": "^1.0@RC",
     "drupal/override_node_options": "^2.6",
     "drupal/pantheon_advanced_page_cache": "^2.1",
+    "drupal/pantheon_secrets": "1.0.1",
     "drupal/paragraphs": "^1.12",
     "drupal/paragraphs_features": "^2.0@beta",
     "drupal/pathauto": "^1.8",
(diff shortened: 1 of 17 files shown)


dblanken-yale (approved):
Looks good.  Since ys_secrets isn't enabled now, should we delete the code as well?  We always have git history to bring it back if we need it.  Other than that question, it all seems to be well hidden and accounted for.  Nice work.

vinmassaro:
Thanks for looking. I opened up a future task to delete ys_secrets since it still has to exist to uninstall it.

nJim (approved):
Everything is working great. I'm glad to see this go in a semi-more-modern direction with centrally managed env variables. My only issue is that now we can not test these on our localdevs. This means that we can not test a webform submission for example.

Luckily folks in the community are talking about it. Maybe this guy will come up with an override tool. https://www.drupal.org/project/pantheon_secrets/issues/3380531 

In the mean time, I think we should move this work forward. Having less support for localdev is less of a risk than our current problem -- deployed sites not having keys. Ship it.

PR #398, opened 2023-08-23
CI: remove tmate debug

vinmassaro:
### Description of work
- Remove tmate debug step in Github Actions deploy job. It was useful when developing but now prevents action runs from failing fast.


nJim (approved):
Approved.

PR #403, opened 2023-08-25
YALB-1446: Feedback: Views UI Matching Labels

dblanken-yale:
### Description of work
- Changes the any/all labels in the view block form to:
  - Any: 'Can have any term listed in tags and categories'
  - All: 'Must have all terms listed in tags and categories'
- Move the any/all match section to be next to the tag inclusion/exclusion fields
- Ensured that javascript selection still functioned
- Modified the CSS to give these items a little more breathing room on the sidebar as the text was bleeding out of the boundaries of the label

### Functional testing steps:
- [x] Attempt to add a view block
- [x] Note that the labels for "Match content that has" match the above descriptions of work
- [ ] Verify that the matches fields appear below the include/exclude tag fields
- [ ] Verify that upon click, it visually shows that it is selected
- [ ] Verify that upon edit, there is plenty of room for the text of these items for the labels they are in
- [ ] Verify that other labels/radios on this page are unchanged

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/assets/css/views-basic.css
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/assets/css/views-basic.css
@@ -147,6 +147,14 @@
   height: 150px;
 }
 
+/* while other items look great, the term-operators text was getting cut off
+ * so we need to adjust the label width and height for the term operators
+ */
+#drupal-off-canvas .grouped-items .fieldset--group .fieldset__wrapper--group .glb-form-type--radio input.term-operator-item + label {
+  width: 130px !important;
+  height: 150px !important;
+}
+
 /* second column */
 #drupal-off-canvas .grouped-items .fieldset--group.views-basic--view-mode .fieldset__wrapper--group .glb-form-type--radio label {
   width: 175px;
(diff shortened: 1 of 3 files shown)


dblanken-yale:
Adding don't merge until PR fix for ys_node_access goes into develop.

codechefmarc (approved):
Looks good to me! Approved!

Unrelated to this ticket, but something that caught my eye while I was testing and thought we could include it in this PR: we should probably move the "sorting by" below the "Match content that has" So that the "Match content" is next to the terms. I looked at the code briefly and it would require moving the `term_operator` form element to another group (filter and sort) and then in `massageFormValues` to make sure we're getting the correct form value.

If you'd feel comfortable with that here, cool, if not, create a ticket and we can get to it at some point.

dblanken-yale:
@codechefmarc I don't mind giving it a try here.  Let me see what I can do and I'll have you re-approve the changes.  :D

dblanken-yale:
@codechefmarc Ok take a look; I moved it into the other array like you mentioned, which seemed to break the js wire, so I modified it to be added.  Then I noticed when opening the sidebar to edit the view block, the text on these were bleeding outside of the boundaries of the label, so I targeted it via a class to extend the width a little, keeping the other radio buttons in tact.  Let me know if I missed anything, but when I tested, it saved the values fine and was able to filter without issue.

codechefmarc (approved):
Very nice! Thank you for that! (Was bugging me that those fields weren't "together"!) Approved!

PR #404, opened 2023-08-26
Feat motion

nJim:
### Description of work

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...


web/profiles/custom/yalesites_profile/modules/custom/ys_themes/src/ThemeSettingsManager.php
```diff
@@ -161,8 +161,36 @@ class ThemeSettingsManager {
...
+      ],
+    ],
   ];
 
+  /**
+   * A string to represent a false value in an HTML attribute.
+   *
+   * @var string
+   */
+  const ATTRIBUTE_FALSE = "['false']";
```
nJim, line 185: Asking Joe a question about this on team. Leaving a comment so that I can link to it.

dblanken-yale: Sure.  I was able to look at the tracing of the == operator parsing, and it looks like it is indeed checking against "true":

Screenshot 2023-09-01 at 10 40 09 AM

In the last one, false is the overall == operator result that's being pushed onto the stack.  The first value is the lefthand value, followed by the righthand value.

Keep in mind this data is from Storybook/CL, not directly from Atomic, which would be better.  I haven't yet figured out how to get the logs out of that yet to see what we're passing from the Drupal side.

I'm going to attempt to write a quick test for this if I can between Drupal and CL to see if it keeps this when json_encoding.  But that's what I've found so far.  Let me know what you two find out as well.  :D

joetower:
@nJim I pushed up a branch to the component library which: 
- Changes `animate_item` to `enabled/disabled/` instead of `true/false`
- Adds `animate_item` to card_collection to enable/disable artistic animation on collection of reference cards (this animation was not enabled for custom_card_collections, but we could)

https://github.com/yalesites-org/component-library-twig/pull/286

PR #361, opened 2023-07-21
YALB-956: Bug: Table Properties window doesn't work

dblanken-yale:
## [YALB-956: Bug: Table Properties window doesn't work](https://yaleits.atlassian.net/browse/YALB-956)

Using solution proposed from https://www.drupal.org/project/layout_builder_modal/issues/3063834

### Description of work
- Create jQuery to allow interaction of the modal dialog when CKEditor is active
- Load this script on ys_core

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...

PR #411, opened 2023-09-07
YALB-1291: Update to CKEditor5

codechefmarc:
### Description of work
- Upgrades CKEditor 4 to CKEditor 5 in preparation for Drupal 10 upgrade
- Changes the following text formats to use CKEditor 5: Basic HTML, Heading HTML, and Restricted HTML
- Uninstalls CKEditor 4
- Adds temporary patch to increase the height of the CKEditor text area. This is fixed in Drupal 10.1 and the patch should be removed prior to upgrading to Drupal 10.
- Requires [Atomic PR-171](https://github.com/yalesites-org/atomic/pull/171)

### Functional testing steps:
- [ ] Login to the site and visit any page
- [ ] Enter layout builder and add a block of type "text"
- [ ] Verify that the text editor has changed, see the comparison below:

Screenshot 2023-09-07 at 2 04 30 PM

Screenshot 2023-09-07 at 2 04 08 PM

- [ ] Add some text to the editor including styling like bold, italic, headings, tables, and lists.
- [ ] Verify that all of the items available in the toolbar work and you can read the text (i.e. no white on white)
- [ ] Add the block
- [ ] Verify that the markup is as expected in the layout preview
- [ ] Add a block of type "Accordion" or "Tabs"
- [ ] Repeat the same process as outlined above and verify that tables, specifically, are styled (in the editor) to allow for ease of editing. Add header rows and columns as well by using the pop-over tools:

Screenshot 2023-09-07 at 2 10 04 PM

- [ ] Add a block of type "Action Banner" and verify that the Action Banner Content is the new text editor as well but only has bold and italic as the allowed markup.
- [ ] Feel free to check other blocks and node edit forms for these changes as well.


codechefmarc:
Just noting that @joetower found a bug in this where pages on the site won't load at all now after saving nodes that aren't in a menu. I think I know the fix and will implement and notify when ready for testing.

codechefmarc:
@joetower - This PR is ready for testing. I fixed the issue you found - thank you for finding it. However, I've got a question - the multidev CSS for this one looks odd - see the top admin toolbars. Other PR's that I've created since (https://github.com/yalesites-org/yalesites-project/pull/414) the admin toolbars look fine. Is this from one of those odd disconnects where updates to Atomic / admin theme aren't in there somehow? It works fine locally for me.

joetower:
@codechefmarc That is _weird_. :thinking:

joetower:
@codechefmarc So, I disabled `Aggregate CSS files` in https://pr-411-yalesites-platform.pantheonsite.io/admin/config/development/performance and all appears well. 

This is something we'll need to look into. It's so odd that it's an issue now.

dblanken-yale (approved):
I was able to accomplish everything detailed, as well as created a page that wasn't in a menu and was able to load it.  I love CK5's look, especially when it comes to tables!  I remember having to make some CSS and JS changes so that old CK would do the table modal correctly; those might no longer be needed.  Looks like YALB-1442 just for referencing here.

codechefmarc:
@dblanken-yale - I noticed there was a `atomic/css/ckeditor-overrides.css` that I believe was part of that ticket. Do you think this is needed anymore? We can remove it as part of this PR if not.

dblanken-yale:
@codechefmarc Nope, we can delete that--it was specific to the table modal for CK4.  I believe the code in this isn't needed any longer as well since it addressed the modal not being able to be used for the table modal.

https://github.com/yalesites-org/yalesites-project/pull/361/files

PR #414, opened 2023-09-08
HOTFIX: Fix Nodes not able to be viewed if no menu item

codechefmarc:
### Description of work
- It was discovered that with some custom code, if a node didn't have a menu item attached and you saved the node, it would mess up the site and prevent nodes (and the menu form) from being viewed and instead give missing route errors.
- This PR removes the custom code and to edit menu item extras (the heading CTA), content editors need to go into the menu item itself and can't be done on the node form.

### Functional testing steps:
- [x] Edit a node that does not have a link in the main menu
- [x] Save the node
- [x] View the node
- [x] Verify no errors show


nJim (approved):
Approved. This fixes the issue.

PR #421, opened 2023-09-12
YALB-1303: Finding content tagged with a specific tag

codechefmarc:
### Description of work
- Filter "Content" view and "Manage Pages" by tags
- Filter "Manage Posts" by tags and/or post categories
- Filter "Manage Events" by tags and/or event categories
- Filter "Manage Profiles" by tags and/or affiliation

### Functional testing steps:
- [x] Login and visit the "Content" page (`/admin/content`)
- [x] Verify that there is now an exposed filter for "Tags", the "Tags" field is now in the table, and that filtering on this field works as expected.
- [x] Visit Content -> Manage Pages (`/admin/content/manage-pages`)
- [x] Verify that there is now an exposed filter for "Tags", the "Tags" field is now in the table, and that filtering on this field works as expected.
- [x] Visit Content -> Manage Posts (`/admin/content/manage-posts`)
- [x] Verify there are 2 exposed filters for "Tags" and "Categories" and that filtering works as expected for these filters (Categories may need to be added to some posts to test this)
- [x] Visit Content -> Manage Events (`/admin/content/manage-events`)
- [x] Verify there are 2 exposed filters for "Tags" and "Categories" and that filtering works as expected for these filters (Categories may need to be added to some events to test this)

Screenshot 2023-09-12 at 5 30 04 PM

- [x] Visit Content -> Manage Profiles (`/admin/content/manage-profiles`)
- [x] Verify there are 2 exposed filters for "Tags" and "Affiliation" and that filtering works as expected for these filters (Profiles may need to be added to test)

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/chosen.settings.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/chosen.settings.yml
@@ -1,17 +1,21 @@
 _core:
   default_config_hash: exRaKE0s6uZbfQrl_2WFEcTlc_1t81xEqCLuCCteUSk
-minimum_single: 20
+minimum_single: 15
 minimum_multiple: 20
 disable_search_threshold: 0
 minimum_width: 0
+max_shown_results: null
 use_relative_width: false
 jquery_selector: 'select:visible'
 search_contains: false
 disable_search: false
 allow_single_deselect: false
+disabled_themes:
+  claro: '0'
+  atomic: '0'
+  gin: '0'
+  ys_admin_theme: '0'
+chosen_include: 2
 placeholder_text_multiple: 'Choose some options'
 placeholder_text_single: 'Choose an option'
 no_results_text: 'No results match'
-max_shown_results: null
-disabled_themes: {  }
-chosen_include: 2
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.node.profile.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.node.profile.default.yml
@@ -24,6 +24,7 @@ dependencies:
     - node.type.profile
   module:
     - allowed_formats
+    - chosen_field
     - field_group
     - maxlength
     - media_library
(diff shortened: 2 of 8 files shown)


jvaughn-johnson:
I went through all functioning testing steps and everything works as expected for me. Thank you.

For the exposed filters on "Tags" and "Categories", is it possible to have the exposed filter to be typed into (i.e. typing "admin" into the search bar) and also be a dropdown? I'm not sure if that's worth it, but I'm thinking of the potential extreme case where someone might know what's available to them.

dblanken-yale:
@codechefmarc Unrelated to this PR, just wanted to throw it out there since you're in "Chosen" mode and I noticed it while testing this.  😁

I noticed that the create profile contact info, the tags field is not using chosen, but the drupal select.  Didn't know if you wanted to incorporate the fix in this or wait for a new ticket, but my worry is we'd run into a backward compatibility problem if we don't address it soon.

Screenshot 2023-09-13 at 2 13 09 PM

codechefmarc:
@dblanken-yale - good call, that is easy to add to profiles so I'll do that and let you know when finished.

dblanken-yale:
@codechefmarc Awesome; will be easy to test with this PR I think as I've already created the profile.  We can see what it will do and then test a new one.  Just ping me and I'll test.  :D

codechefmarc:
@dblanken-yale - OK, you're good to go - Profiles tagging is Chosen enabled!

dblanken-yale:
@codechefmarc Just verifying; did you happen to add "generic" as a tag to Davy Jones?  I ask because while admin is showing, which is what I had entered, I didn't remember entering "generic" as a term, and of course that's taxonomy id 1, which makes me suspicious.  If you did, that makes me feel better.

dblanken-yale:
The other profile works great, and doesn't have generic so I bet I either added it and forgot or someone else did.  I think that looks good now!  Good fix!

codechefmarc:
😆 Yes, that was me adding it to test. So, not an error! We should ban any taxonomy term with id 1. Hehe.

@jvaughn-johnson - would you like to take another look before I merge? Thanks!

jvaughn-johnson:
Hi @codechefmarc. Somehow I missed your comment from 5 days ago. The update looks good to me and I think the logic behind when Chosen decides to show both the dropdown and autocomplete makes sense. I think people will appreciate this new feature overall. Thank you.

PR #422, opened 2023-09-13
Fixes for semantic release

vinmassaro:
### Description of work
- Adds `YALESITES_BUILD_TOKEN` to semantic release action to fix 401 errors coming from npm
- Change version number in yalesites_profile.info to a comment that can be replaced by semantic-release-replace-plugin
- Remove `@semantic-release/git` plugin so changes only exist in the tagged release and are not committed back to the repo


nJim:
This seems like a simple change with little consequence if it does not work. Are you okay if I approve the code by skip the functional testing? Otherwise I may fork the repo to test things. But I want to save the time if this seems like overkill

PR #423, opened 2023-09-13
YALB-1472: Bug: Image generation failures

dblanken-yale:
### Description of work
- Base images uploaded greater than 3840x2160 (4k resolution) will be resized to that resolution
- ImageMagick restrictions have been set:
  - Use disk if the area of the image is greater than 20MP
  - Use disk if the memory used is over 96MB
  - Write to disk from memory upon hitting 196MB
- Use the main repo for search_api_html_element_filter (caused build errors for multidev)

### Functional testing steps:
- [ ] SFTP to the server
- [ ] Delete either all subfolders under /code/web/sites/default/files/styles or select generated files inside subdirectories.
- [ ] Visit either a page that contains an image style you deleted, or in the case of deleting all, any page containing images.
- [ ] Notice the time needed to generate the image
- [ ] Verify that on refreshes, caching and pre-generation takes over and the image loads very quickly

_Note: On pages with many large images, this still could cause a no-load, but on subsequent requests it should resolve and properly generate.  From then on, it should be a non-issue._


joetower:
@dblanken-yale I ran through testing this several times. 

On two occasions, the homepage took about 1 minute to load all images. One time, the image which took the longest to load was the homepage hero. The other time, the card images in the "Get to Know YaleSites" block took the longest. 

Also on two occasions, the images in the "Get to Know YaleSites" block failed to load. If I refreshed the page they loaded, but they still took a while.

Screenshot 2023-09-21 at 9 30 19 AM


I'd love to test this outside of a multidev. It's tough to test how viable this might be outside of multidev limitations. Wondering if we should add further image optimization/lossless compression, as well.

dblanken-yale:
@joetower Bummer that it's still not the best experience.  Great points.  I mean, the change is just flags on ImageMagick, so maybe I could try using a live environment and then revert the flags.

@vinmassaro Do we have any current options for testing this on a live environment that wouldn't put a lot of work on your side?

vinmassaro:
@dblanken-yale AFAIK there are no major differences between multidev and live environments and the PHP memory limit is going to be the same. There is CDN caching, but the origin server still has to generate the images.

I don't love this test because we're still using source images that are way too large, which may be setting this testing up for failure off the bat. On one page with multiple images being loaded, I picked one at random and the source image was 18MB. If I take that image and scale it to 5000px wide, then apply lossless JPEG compression, it ends up being 4MB.

dblanken-yale:
@vinmassaro Yes, I think this solution just isn't the right one.  I just spoke with @nJim and we think we're going to start with a limit of 10MB images and if feedback gets hot about it, we can up it.  Then comes updating any existing sites with these new versions, but sounds like we will span this to next week to figure that out.  I'd love to only have to deal with 4MB jpegs.

nJim (approved):
Approved. These are sensible limits and work for our environment.

PR #426, opened 2023-09-18
YALB-1554: Remove "Platform admin" from options for site admins when adding or managing site users

dblanken-yale:
### Description of work
- Adds and enables `role_delegation` module
- Sets up `role_delegation` to only allow `site admins` to set `site admin` permission levels
- Update CAS bulk update form to use the `role_deleagation` module's settings to determine options to select from

### Functional testing steps:
- [x] Log in as the admin
- [x] Open the same site in an Incognito window and log in as your own CAS user
- [x] On both sites, visit `People`
 - [x] Take note of your CAS user's access level--it should be only `Platform administrator`; if not, please set it on the admin user account's page

#### Test adding non-CAS users
- [x] Do the following as your CAS user:
  - [x] Click `Add User` in the upper right corner
  - [x] Add in bogus data and ensure that on your logged in user side
  - [x] Verify that you have two options available for roles, `Platform administrator` and `Site administrator`
  - [x] Click submit and verify that the user you added has that role

#### Test adding CAS users
- [ ] Do the following as your CAS user:
  - [x] Click `Add CAS user(s)` in the upper right corner
  - [x] Add gas users, but verify that currently you have two roles available to select from, `Platform administrator` and `Site administrator`
- [ ] As your admin user:
  - [x] Find your CAS username in the people list and edit
  - [x] Unselect `Platform administrator` and ensure that `Site administrator` is selected and save.
- [ ] As your CAS user:
  - [x] Refresh the page; you should now only see `Site administrator` as the only role you can assign
  - [ ] Add any CAS users you can think of and submit
  - [ ] Verify that those users were added with the `Site administrator` role

NOTE: I noticed that each time our users log into the site via CAS, it gets the `Platform administrator role`.  My assumption given the CAS group name used is that this does not happen for our users of the platform.  If this is incorrect and any logged in CAS user gets this outside of our group, please let me know.

### DIFF
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -75,6 +75,7 @@
     "drupal/recaptcha": "^3.1",
     "drupal/recaptcha_v3": "^1.7",
     "drupal/redirect": "^1.7",
+    "drupal/role_delegation": "^1.2",
     "drupal/search_api": "^1.25",
     "drupal/search_api_exclude": "^2.0",
     "drupal/search_api_html_element_filter": "^1.0",
--- a/web/profiles/custom/yalesites_profile/config/sync/core.extension.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.extension.yml
@@ -104,6 +104,7 @@ module:
   redirect: 0
   redirect_404: 0
   responsive_image: 0
+  role_delegation: 0
   search_api: 0
   search_api_db: 0
   search_api_exclude: 0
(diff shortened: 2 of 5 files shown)


codechefmarc (approved):
This all worked for me - but I didn't add a CAS user that wasn't already on the site because I don't know a CAS username to add - if you have any examples I'm happy to test. But otherwise, all looked great to me. I've used role delegation in the past and it's really a good one. Approved!

dblanken-yale:
@codechefmarc Ah true, you can delete and attempt to re-add me if you'd like.  I'm the db one.  :D  April was able to test also so I understand if you're moving on. :)

And thanks for checking with Vincent; I was in "image" mode during standup and forgot to ask.  LOL

PR #432, opened 2023-09-20
YALB-1373: Feedback: URLs with filler words

dblanken-yale:
## [YALB-1373: Feedback: URLs with filler words](https://yaleits.atlassian.net/browse/YALB-1373)

Before, we were removing filler words in the URL generation, such as 'a', 'the', 'this', 'that', etc.  We no longer wish to remove these, so urls will now include these filler words as well going forward.

### Description of work
- Stops Pathauto from removing strings when generating a URL alias for a node.

### Functional testing steps:
- [ ] Create a new page naming it something with a filler word or two, like "This is a page to remember"
- [ ] When you visit the view of this page, ensure that the url includes the full list of words in the name of the page (this-is-a-page-to-remember)


nJim (approved):
Approved. Ship it.

PR #439, opened 2023-09-27
YALB-1566: Fix deployment for Berkeley Divinity

nJim:
### Description of work
- Adds a condition to the layout builder update hook to ensure content exists before updating the content.

### Functional testing steps:
- I have tested this locally to ensure it does not have any negative impacts. Since this is an update hook, it will not run on other environments so the scope is limited to Berkeley Divinity. I'm happy to setup a sandbox for testing if we want a more comprehensive test plan.


codechefmarc (approved):
I think this is just fine. If this hook as already been run on Berkeley Divinity, I assume it'll simply be run manually via Terminus on just that one site then. I don't have Pantheon access to it, otherwise I would have tested locally myself as well, but I think we're good to go.

PR #440, opened 2023-09-27
YALB-1186 and YALB-1210: Upgrade to Drupal 10 and Pin versions in profile composer.json

codechefmarc:
### Description of work
- Upgrades to Drupal 10
- Updates custom code to be compatible with Drupal 10: `ys_views_basic` CSS, `ys_layouts` `UpdateExistingNodes.php`
- Adds a replacement fork MR for Search API HTML Element Filter to work with Symfony 6 (see https://www.drupal.org/project/search_api_html_element_filter/issues/3390283)
- Adds the composer package `laminas/laminas-escaper` for BugHerd compatibility with Drupal 10
- Pins versions of contrib modules in profile `composer.json` file
- See link to Drupal 10 notes Teams spreadsheet in [Jira ticket](https://yaleits.atlassian.net/browse/YALB-1186) for specific notes about each contrib module pinning and Drupal 10 compatibility

### Specific Module Testing
The following have not been tested as I wasn't aware how to test them. Perhaps someone from Yale is better able to test:
- [ ] Captcha
- [ ] Mailchimp Transactional

### General Testing Note
Drupal 10 operates almost entirely like Drupal 9.5, but there were some changes. I did my best to test as much as I could on the multidev but it would be best to get as many eyes on this as possible and test multiple areas of the site. Below are some specific areas that needed some changes in reaction to the upgrade to Drupal 10.

### Front-end testing steps:
- [x] Visit the multidev https://pr-440-yalesites-platform.pantheonsite.io/ as a logged out user
- [ ] Load various pages and verify that: the pages load, images load, and components all look appropriate

### Back-end testing steps:
- [x] Visit the multidev as a site admin role
- [x] Visit the content overview page https://pr-440-yalesites-platform.pantheonsite.io/admin/content
- [x] Filter content by Tags which uses Chosen and verify that Chosen works as expected and that the filters work as expected.
- [x] Visit any page on the site and enter layout builder
- [x] Click "Add block" at the bottom of the content region
- [x] In the modal to choose a block, start typing "vi" and verify that the list only shows a few blocks that match

Screenshot 2023-10-04 at 4 13 06 PM

- [x] Add a block of type "View" to the page
- [x] Save the block and verify that the view shows as expected
- [x] Configure the view block that was just added and verify that the sidebar for the view is styled correctly

Screenshot 2023-10-04 at 4 19 16 PM

- [x] Verify that the Chosen interface is styled and works the same as Drupal 9

Screenshot 2023-10-04 at 4 19 23 PM

- [x] Verify that the rest of the layout builder interface (outside of editing blocks) looks the same as Drupal 9

Screenshot 2023-10-04 at 4 18 53 PM

- [ ] Test other component blocks - verify that adding other component blocks to the page works as expected


patches/add-php-c-format.patch
vinmassaro, line 1: Should we still be maintaining this or propose it as a patch to the module?

codechefmarc: IIRC, there was a patch proposed, but didn't have the "c" format in their patch. I suppose it's easy we could submit our version to add "c" to it.

web/sites/default/services.yml
vinmassaro, line 1: Not sure if the permissions on this file were supposed to change or not but figured I would mention it.

vinmassaro (commented):
Looking good @codechefmarc! I tried to do a `drush site install` locally on this branch to test installing from scratch and it failed due to not being able to install google_analytics from default config. Luckily there was a patch available so I added it in to resolve the issue.

codechefmarc:
Thanks, @vinmassaro! I'm looking into the `smart_date` patch to see if we still need it an also the file permission on `services.yml` as well.

web/sites/default/services.yml
codechefmarc: Right, so I followed these instructions and at the end (step 5) it said to change file permissions for that file to 644.
https://www.drupal.org/docs/upgrading-drupal/upgrading-from-drupal-8-or-later/upgrading-a-composer-based-site
We can certainly change it back to 755 if we want as that is still not world/group writable.

patches/add-php-c-format.patch
codechefmarc: Ok, I tested out removing it as we did refactor the smart date stuff for events a while back. All tests looks fine after I removed the patch.

kara-franco (approved):
Fantastic work! I went through the public-facing pages and also built out a test page. All behaviors and displays looked similar to D9. 

One odd thing I noticed, that resolved after re-setting a setting: 
The Pre-built From submit button was failing contrast (pictured below). I went in and re-set the Theme settings for Global Theme and Button Theme and the issue resolved and I cannot for the life of me recreate it. I think it's a non-issue? Maybe a remnant of the palette updates? 

Besides that, looks amazing! I'm OOO next week, so if any tweaks or a11y things come up, Mike V can help out! 

Screenshot 2023-10-06 at 2 17 26 PM

miketullo95:
Going through use cases and noticed these two things:

1.  Theme settings accent buttons don't show the colors
Screen Shot 2023-10-12 at 10 55 41 AM

2.  Mega menu style is off - missing white accents/line under "explore ..." 
Screen Shot 2023-10-12 at 10 57 45 AM

codechefmarc:
@miketullo95 - The header accent colors are now showing up on the multidev, thanks for pointing that one out!

vinmassaro:
@codechefmarc it occured to me that we are pinning versions in the profile's composer.json, but not in the root composer.json where Drupal core packages are defined. The `drupal/core*` packages should probably be pinned as well so we have a consistent version of drupal core.

codechefmarc:
> @codechefmarc it occured to me that we are pinning versions in the profile's composer.json, but not in the root composer.json where Drupal core packages are defined. The `drupal/core*` packages should probably be pinned as well so we have a consistent version of drupal core.

Good call - yes, that should be included here. I don't have hours this week to tackle this, but @nJim may?

codechefmarc:
> @codechefmarc Is the recommended way to upgrade to remove `./vendor/*` and the lock file and then `npm run setup` on an existing D9 instance? I ask because I did get it working and it's fine now, but it did take a couple of tries from me to get there. Wanted to make sure I had the upgrade path right.

@dblanken-yale - Yah, locally, I had some small issues with composer so I had to delete the lock file and the vendor directory and then run `lando composer install`. I was a little worried about that, but, the CI and the multidev worked great, so I think things should be fine. My thought on this is that the local was simply built on the previous 9.5 version and the multidev the artifact is being created from scratch?

miketullo95 (approved):
This looked great! Went through all of my AX use cases and only tasks that failed were 

- CAS Restricted Nodes (Known issue, being worked on now)
- Adding a NetID user (did not have an NetID to test that was not already added to the environment)

PR #446, opened 2023-09-29
YALB-1564: Header: Mega Menu Description Link

codechefmarc:
### Description of work
- Adds the ability to override the header menu link CTA text from within the page add/edit form.

### Functional testing steps:
- [x] Login to the site and create a new page
- [x] Title the page and open the right-sidebar if not already open
- [x] Open the "Menu settings" in the sidebar
- [x] Click "Provide a menu link"
- [x] Verify that you see a "Mega Menu Top Level Link Text" field at the bottom of the menu link settings form.

Screenshot 2023-09-29 at 11 29 35 AM

- [x] Enter some text to override the link text
- [x] Place this new page at the top level of the main navigation and save the page
- [x] Save layout builder
- [x] Create another page
- [x] Title the page and Open the "Menu settings" in the sidebar
- [x] Click "Provide a menu link"
- [x] Place this new page under the previously created page in the menu
- [x] No need to fill out the Top Level Link Text override
- [x] Save the page, then also save layout builder
- [x] Open the menu for the first page and verify that the overridden link text is what you typed in above

Screenshot 2023-09-29 at 11 32 22 AM

- [x] Create a third page
- [x] Title it, but do not add a menu link to this page
- [x] Save the page and save layout builder
- [x] Verify no errors when saving or viewing the page
- [x] Edit the first page again, with the overridden menu link text
- [x] Remove the overridden link text and save the node
- [x] Verify that the fallback text "Explore [menu title]" is now the link text

Capture-2023-09-29-113704

- [x] Overriding of this link text can also be done via editing the menu item itself from `/admin/structure/menu/manage/main`

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
@@ -404,9 +404,10 @@ function ys_core_save_menu_link_fields(array $form, FormStateInterface $form_sta
     // Only save the menu item extras if there is a menu link with a route.
     if ($link->getUrlObject()->getRouteName()) {
       $form_display = EntityFormDisplay::load('menu_link_content.' . $link->getMenuName() . '.default');
-      assert($form_display instanceof EntityFormDisplay);
-      $form_display->extractFormValues($link, $form['menu']['link']['extra'], $form_state);
-      $link->save();
+      if ($form_display instanceof EntityFormDisplay) {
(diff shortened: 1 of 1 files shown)


dblanken-yale (approved):
LOVE that you can modify it in either area.  Great work!

Just checking, is an assert ok to have here?  Or should we somehow exit gracefully with a message?  If the assert fails for some reason, what happens to the request?

codechefmarc:
@dblanken-yale - Good catch. I copied most of that code from the menu item extras issue queue. I looked into it, and that would be better as an `if` statement rather than an `assert`. I'll change it.

PR #452, opened 2023-10-05
YALB-1579: Fixing theme form error messages

nJim:
### Description of work
- Passes properties from the theme settings form via #context array to prevent render array errors.
 
### Functional testing steps:
- [ ] Login the website as a site administrator.
- [ ] Navigate to any node
- [ ] Click the 'Theme Settings' link in the toolbar to expand the modal window
- [ ] Review the error logs and verify no new error messages appear (admin/reports/dblog). A screenshot of previous errors is included below for reference. These related to a malformed render array.

Screen Shot 2023-10-05 at 8 36 27 AM


dblanken-yale (approved):
No errors!  Yay!

I did notice that opening the theme settings and hitting save created an error; it's like the default is not set for the footer theme.  Not sure if this is addressed in another ticket, but thought I'd mention it here in case it wasn't.

PR #453, opened 2023-10-05
YALB-1577: Fix time format on condensed events

nJim:
### Description of work
- Update the fields formatter to use 'plain' text instead of the 'smartdate' format. This ensure the time is passed to the component library in the correct format.

### Functional testing steps:
- [ ] Login as a site administrator
- [ ] Create an event with a far future date. Save this event and make sure it is published.
- [ ] Add a basic 'views' component to a page-node. The view should show all future events in a 'condensed' layout.
- [ ] Save the page and view the content. Verify that the view displays events with the correct date for the event. Previously this was showing today's date for every event.

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.event.condensed.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.event.condensed.yml
@@ -54,15 +54,11 @@ bundle: event
 mode: condensed
 content:
   field_event_date:
-    type: smartdate_default
+    type: smartdate_plain
     label: hidden
     settings:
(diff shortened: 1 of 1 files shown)


dblanken-yale (approved):
This is awesome work.  The dates are working just fine.

I did find this visual issue, although not sure it's appropriate for this ticket, so still accepting.  Just documenting.  :D

Screenshot 2023-10-18 at 10 25 27 AM

nJim:
@dblanken-yale Thank you for the note. There are a few issues on the admin theme that we need to address. But we will come at it with a bigger plan later next month. This irks me, but is not a blocker. I will doucment it for when Joe returns.

PR #454, opened 2023-10-05
YALB-1576): Bug: CAS Restricted pages

codechefmarc:
### Description of work

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...


vinmassaro:
This issue is caused by a default entry in the node_access table that gets removed after rebuilding node access permissions, so this can now be closed out.

PR #461, opened 2023-10-13
YALB-1586: Add Dependabot to project repo

vinmassaro:
### Description of work
- Adds `/.github/dependabot.yml` to automate weekly updating of pinned dependencies in root composer.json and yalesites_profile composer.json files.

This work is blocked until #440 is merged.


nJim (approved):
Marking as approved. This looks consistant with other projects where we use dependabot. I believe this is now unblocked, but I will leave it to Vincent to design when we should merge this work.

vinmassaro:
I found some outstanding tasks needed for this PR.

- Pin Drupal core in composer.json
- Check for non-pinned versions of dependencies in CI so we don't end up with modules being added without pinning
- Change to monthly schedule

dblanken-yale (approved):
@vinmassaro Do we want to also check for the following?

```
"drupal/module": "*",
```

I think the regex if we do would be: `"\\^|\\~|\\>|\\<|\\*"`

Look good to me otherwise; I was able to run the pin check on a separate branch and worked as expected.

PR #462, opened 2023-10-18
YALB-1593: Bug: FE: Views pagination not showing on front end

dblanken-yale:
### Description of work
- Add patch to allow `dom_id` to dictate unique pager. (https://www.drupal.org/project/drupal/issues/3225987)
- Ensure ys_basic_views have a unique dom_id before building to generate a unique pager.
- Remove C date format patch for smart_date

### Functional testing steps:
- [ ] Visit https://pr-462-yalesites-platform.pantheonsite.io/test-page
  - [ ] Ensure that each view has a different pager with different max numbers.
- [ ] Feel free to create your own nodes and views, but ensure that:
  - The number of total nodes in each view are different so you can tell a difference in the pager
  - Each view is set for pagination
  - That the two views are on the same page


miketullo95 (approved):
tested and all looks good! 👍

nJim (approved):

PR #464, opened 2023-10-25
YALB-961: WYSIWYG: Text field size to reflect character limits

dblanken-yale:
### Description of work (Multidev)
- Adds patch to allow CKEditor to calculate `--ck-min-height` variable based on rows defined
- Adds min-height to CKEditor displayed "text area" spaces to respect `--ck-min-height`
- Changes a multitude of block content rows values to 3

### Description of work (Branch)
- Changes a multitude of block content rows values to 3

### Functional testing steps:
- [ ] Attempt to add the following block, noticing the change in textarea height on each compared to other blocks.
  - [ ] Action Banner
  - [ ] Grand Hero
  - [ ] Image
  - [ ] Quote
  - [ ] Quick Links
  - [ ] Video
  - [ ] Wrapped Image

This branch is now based on master with D10, not needing the patch that was used previously to create the multidev.  For the truest experience of the result, bring local with D10.

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.block_content.grand_hero.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.block_content.grand_hero.default.yml
@@ -88,7 +88,7 @@ content:
     weight: 6
     region: content
     settings:
-      rows: 5
+      rows: 3
       placeholder: ''
     third_party_settings:
       allowed_formats:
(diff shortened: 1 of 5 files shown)


dblanken-yale:
Multidev created based on the patched 9.5 version; the branch has since been moved to be based on master to take advantage of the D10 fix for this.  It should essentially function similarly, but be aware this isn't a 1-1 since it's using a patch.

If you can, bring locally to see the D10 version.

miketullo95 (approved):
This looks great! I'm not able (or am confident enough) to pull this down locally. But on the multidev this is working exactly as intended! 👍

codechefmarc (approved):
Was able to test this locally, finally - sorry for the delay. All looks good, I was able to see the different heights, though it was subtle in some cases. Approved!

PR #467, opened 2023-10-27
YALB-520: Links: Dynamically apply link treatments

dblanken-yale:
## [YALB-520: Links: Dynamically apply link treatments](https://yaleits.atlassian.net/browse/YALB-520)

MultiDev PR: https://github.com/yalesites-org/yalesites-project/pull/467
Component Library PR: https://github.com/yalesites-org/component-library-twig/pull/319
Atomic PR: https://github.com/yalesites-org/atomic/compare/YALB-520-links-dynamically-apply-link-treatments

### Description of work
- Shows off link treatment work completed now in component library twig
- Links decorated:
  - External
  - Downloads
  - Target Blank

### Pages to check link treatments
- [ ] [Testing all of the links](https://pr-467-yalesites-platform.pantheonsite.io/testing-all-of-the-links) - Used content import for this one.  🕺 
- [ ] [Mike V Testing](https://pr-467-yalesites-platform.pantheonsite.io/mike-v-testing)
- [ ] [YALB-520 (Marc's page)](https://pr-467-yalesites-platform.pantheonsite.io/yalb-520)
- [ ] [Testing all of the things](https://pr-467-yalesites-platform.pantheonsite.io/testing-all-of-the-things)

### Functional testing steps:
- [ ] Add a block that has a WYSIWYG editor (Text works great here)
- [ ] Attempt to add different types of links:
  - [ ] Internal link
  - [ ] External link
  - [ ] Download link
  - [ ] Link with a `target="_blank"` defined: Note that I don't think it'll allow you to do this out of the box, but as this is an upcoming feature, we are including the decoration here.
  - [ ] Any other link you can think of that the editor will take
- [ ] Save the block
- [ ] Ensure that the layout builder view decorates each of these with those in our component library styles: https://yalesites-org.github.io/component-library-twig/?path=/story/atoms-controls--text-link
  - [ ] Keep in mind some decisions have been made to create a better flow to these; feedback is appreciated:
    - [ ] links are inline-blocked instead of flex so that they appear inline with the rest of the text
    - [ ] underlines are extended to the icons attached to show that it is 'attached' to that link
    - [ ] in the mailto example in storybook, we copy the text to the left of the copy link, but in this one, we actually copy the URL that was given so that users can have different text, but copy the hidden URL.
    - [ ] for downloads, we append the file type to the link name.  i.e. links to PDFs will have `Name of PDF (PDF) <icon>`
    - [ ] although we do not have a telephone treatment in Storybook, I decided to mimic the email treatment; we can disable this until truly needed, but was simple to implement given the mailto implementation.
- [ ] Save and ensure that the page view decorates each of these with those in our component library styles (linked above)
- [ ] Attempt to add other blocks and ensure that the links in them are not taken over by these styles.  (i.e. buttons are not treated)

### Things to look for when verifying
- [ ] Padding around component links; these were adjusted since the icon is not considered when determining the overall width of the box surrounding them
- [ ] Padding to include the icon in the underline; these have been adjusted since the icon is not included in the link calculations, resulting in the underline not originally extending to the icon.

__Keep in mind scope was change to not include mailto and telephone decorations until more design and feedback is completed.  Anchors are no longer decorated as other decorations handle it (local vs external).__

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/CoreTwigExtension.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/CoreTwigExtension.php
@@ -188,6 +188,7 @@ class CoreTwigExtension extends AbstractExtension {
       'zip',
       'csv',
       'xml',
+      'rtf',
     ]);
     $extension = strtolower(pathinfo($url, PATHINFO_EXTENSION));
     return in_array($extension, $fileExtensions);
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/CoreTwigExtension.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/CoreTwigExtension.php
@@ -151,7 +151,7 @@ class CoreTwigExtension extends AbstractExtension {
    *   TRUE if the URL is internal.
    */
   private function isInternal($url) {
-    return $this->urlHasCurrentDomain($url) || $this->isAnchor($url) || $this->isRelative($url) || $this->isData($url);
+    return $this->urlHasCurrentDomain($url) || $this->isQueryString($url) || $this->isAnchor($url) || $this->isRelative($url) || $this->isData($url);
   }
 
   /**
@@ -235,6 +235,19 @@ class CoreTwigExtension extends AbstractExtension {
    */
   private function isAnchor($url) {
     return str_starts_with($url, '#');
+  }
+
+  /**
+   * Check if a URL is a query string.
+   *
+   * @param string $url
+   *   The URL to check.
+   *
+   * @return bool
+   *   TRUE if the URL is a query string.
+   */
+  private function isQueryString($url) {
+    return str_starts_with($url, '?');
   }
 
   /**
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/CoreTwigExtension.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/CoreTwigExtension.php
@@ -3,6 +3,7 @@
 namespace Drupal\ys_core;
 
 use Drupal\Core\Config\ConfigFactoryInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFunction;
 
@@ -33,17 +34,35 @@ class CoreTwigExtension extends AbstractExtension {
   protected $yaleMediaManager;
 
   /**
+   * The Request Stack.
+   *
+   * @var \Symfony\Component\HttpFoundation\RequestStack
+   */
+  protected $requestStack;
+
+  /**
+   * The current domain.
+   *
+   * @var string
+   */
+  protected $currentDomain;
+
+  /**
    * Constructs the object.
    *
    * @param \Drupal\Core\Config\ConfigFactoryInterface $configFactory
    *   The configuration interface.
    * @param \Drupal\ys_core\YaleSitesMediaManager $yale_media_manager
    *   The YaleSites Media Manager.
+   * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
+   *   The Request Stack to retrieve the domain.
    */
-  public function __construct(ConfigFactoryInterface $configFactory, YaleSitesMediaManager $yale_media_manager) {
+  public function __construct(ConfigFactoryInterface $configFactory, YaleSitesMediaManager $yale_media_manager, RequestStack $requestStack) {
     $this->yaleCoreSettings = $configFactory->getEditable('ys_core.site');
     $this->yaleHeaderSettings = $configFactory->getEditable('ys_core.header_settings');
     $this->yaleMediaManager = $yale_media_manager;
+    $this->requestStack = $requestStack;
+    $this->currentDomain = $this->getCurrentDomain();
   }
 
   /**
@@ -53,6 +72,7 @@ class CoreTwigExtension extends AbstractExtension {
     return [
       new TwigFunction('getCoreSetting', [$this, 'getCoreSetting']),
       new TwigFunction('getHeaderSetting', [$this, 'getHeaderSetting']),
+      new TwigFunction('getUrlType', [$this, 'getUrlType']),
     ];
   }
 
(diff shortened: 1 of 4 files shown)


codechefmarc:
@dblanken-yale - I did some testing with this today and here are some of my notes:
- First off great work on this! I would love a tour of the code at some point. I tried to follow some of the JS's and got a little bit lost in how it picks from the different formatters. But neat stuff!
- My tests: See https://pr-467-yalesites-platform.pantheonsite.io/yalb-520
- External - worked
- New Window - didn't work initially because the editor was stripping out the `target` attribute. I added that in (to the multidev only, this is a larger issue that is being addressed in another ticket). So, even with `target` attribute there, it is not getting the link treatment.
- Document - didn't work
- Mailto - worked
- Tel: worked
- Anchor - was this to add the chevron? It didn't work though

I'm happy to do some additional testing if you need. I didn't test other components at this time, but can in subsequent tests.

dblanken-yale:
@codechefmarc Thanks a lot for trying it out.  Ya, I was hoping I'd not have to set a weight to each evaluation, but it is looking like that will be the case.  ☹️ 

I'll implement new window implementation since at the time we weren't allowing it.  Thanks for pointing that out.

Ya I think the ones that didn't work are receiving the internal link treatment first.  Once I get the weights in, it'll be fine I bet.  I'll ping ya to look at your page when it's all done since it shouldn't affect the creation of those links.

I'd be happy to run through the evaluator/render logic whenever you have time.

dblanken-yale:
@codechefmarc Actually, I did implement the include_blank, but I remember now that I couldn't test it because CKE was removing the blank targets.  With your new module I saw you talking about, it will be included?  How did you get it included in the multidev before that module got in?  I try to do that locally using source and it removes it before my evaluator can get to it.

dblanken-yale:
@codechefmarc I was able to get target working by modifying the Basic_HTML config to allow target as both an editing and rendering option for the `a` tag.  But that seems more related to your change of adding the option in CKE to have targets, right?  Should I enable it here?  I did do it locally and it decorates fine now that we have weights.  I didn't know thought if it was appropriate to have here since it'd allow users to put targets in before we truly allowed them.

codechefmarc:
@dblanken-yale - The new module that I mentioned isn't going in until later whenever that ticket (YALB-1368) gets worked on. So I did what I think you did - temporarily modify the basic HTML format on the multidev to allow the `_blank` attribute on links. I just tested it again, and it looks like both all of them are working now!

Yah, since we have both standup and grooming today, perhaps either after grooming or tomorrow we can go over the code. Thanks!

codechefmarc (changes requested):
Ok, did a bunch more testing with more blocks, inline links, etc. All tested great with the exception of the following:

- All tests are here: https://pr-467-yalesites-platform.pantheonsite.io/yalb-520

- Not sure if this is intended or not (based on ordering of the processors) but an external document PDF works fine, but an external new window document only gets the new window treatment. I think that's fine, but wanted to point it out:

Screenshot 2023-11-14 at 10 49 13 AM

- Bug: When adding a telephone in the list, it worked with the copy button, but oddly, when inline with other text, it was copying the wrong thing. I.e. the phone number to copy is 123-456-6789, but it copied "test@example.com".

Screenshot 2023-11-14 at 10 50 50 AM

- Not sure if this a bug? Copy feature does not work within a media modal caption:

Screenshot 2023-11-14 at 10 52 14 AM

- Bug: Grand Hero adds a copy button for an email address and looks strange:

Screenshot 2023-11-14 at 10 52 44 AM

- Bug tabs: getting strange line through the tab, not sure if this is something based on your work or an existing bug, but probably talk with Joe about that one:

Screenshot 2023-11-14 at 10 54 07 AM

dblanken-yale:
@codechefmarc Nice work!  Thanks so much for testing it so much!  I suspect it is due to the order of the processors in the case you mentioned of the external link.  I think I should get some feedback on what "should" win out on really all of them.

The copy functionality is now working--I attempted to use an unmodified version of the copy from storybook, but it seems in these cases, there is no parent that would contain the two elements that are interacting (which that JS version was expecting), so I had to change it to target it's previous sibling.  In my tests, the p tag was the parent so I think that's why it worked before.  They should all now work.  (Well, once Pantheon is able to do a new deploy to the multidev--looks like quay.io is having issues currently)

Looking at the grand hero, it looks as if it has specific special classes to handle itself, so I am excluding that from being treated.  The same goes for the tabs--it is attempting to treat it when it doesn't need it.

All of those fixes have been pushed, although the build currently isn't able to do so.  I'll let everyone know once it is, and will try to test even more controls to make sure I've excluded all I can.

dblanken-yale:
I believe this is ready for a re-review.  It looks like the order of the evaluations in teams are sound, so I think we can go forward with it as is?

dblanken-yale:
The latest "revert" of the "revert" was to implement a change the accessibility team wanted to see as a way to determine whether the copy button's toggle-back event should be done, or if it should stay with a static message instead.

If this is accepted, I'll be creating a ticket for the component library change to mirror this.

joetower:
@dblanken-yale This is looking great! Awesome work.  

I noticed, on this page, `table` links aren't awesome: https://pr-467-yalesites-platform.pantheonsite.io/testing-all-of-the-things. We could add `table a.link--with-icon.ys_linked {display: inline;}` to fix this. I feel as though everything else is looking good. I'll keep testing a bit, though. 👍 

https://github.com/yalesites-org/yalesites-project/assets/366413/39e84b6c-6a74-4097-87da-23a3e9197bf4

dblanken-yale:
@joetower Ouch, outstanding find.  I'll try that inline since that's similar to what we did with links in general to see how it does.  Updates incoming soon.  Thanks for looking at it.

myvaughn (approved):
The changes look good from an accessibility standpoint.

codechefmarc (commented):
Hey David! Sorry it took me a while to get back to this one. Since I've been a little out of the loop on the discussions, I'm not sure which components are excluded from the treatments so I tested all of the components you can add links to.

Test page here: https://pr-467-yalesites-platform.pantheonsite.io/marcs-test-page

Text block - all worked as expected. I see that external documents it defaults to the PDF and document treatment, not external.

As for the rest of the components, all worked except for the following ones with screenshots attached:

* tabs
* accordion
* gallery (caption links were completely stripped out)
* spotlight - CTA worked, links in body did not. (both kinds of spotlights)
* wrapped image (both body text and caption)
* image (caption)
* video (caption)
tabs-external
accordion-new-window
gallery-links-stripped
spotlight-portrait-external
spotlight-landscape-new-window
wrapped-image-new-window
video-with-caption-new-window
image-with-caption-new-window

joetower (approved):
@dblanken-yale I tested this here, using the helpful page examples folks have made, and here https://github.com/yalesites-org/component-library-twig/pull/319. 

Everything is looking great 👍 💯

PR #469, opened 2023-10-30
YALB-1574: Branding: Header Variations Backend

codechefmarc:
### Description of work
- Adds a new header settings form
- Moves some YS Themes and YS Core settings into the new form: enable site search, navigation position, and navigation type (now called header variation)
- Wiring of the components for the new settings
- Refactors some older code used for the footer settings form so now both forms share similar CSS and JS and easier to update
- Adds an update hook to migrate old theme and YS Core settings to new header settings
- Adds an update hook to migrate footer theme `four` to footer theme `one`
- Site name as image is a new field allowing SVG's to be uploaded (platform admins only) and automatically rewrites the SVG title to be the site name
- Note: For the focus header to work, the "Front page" on the Site Settings form `/admin/yalesites/settings` needs to be set to `/node/XX` where XX is a node ID. This form does use Linkit so it will autocomplete the node ID by typing in a page title name.

### Functional testing steps:
- [x] Login to the site as a site administrator role
- [x] Visit the new Header Settings form at `/admin/yalesites/header`

Header Settings - site admin

- [x] Verify you can only see 3 details accordions - Navigation Position, Site Search, and Full Screen Homepage Image
- [x] Switch the header variation and verify that the different accordions will enable/disable based on your selection. I.e. Focus nav only gets Full Screen Homepage Image and the other two get the first two options but not the last.
- [x] Start with the basic and mega nav - set the navigation position and the site search.
- [x] Save the form and verify on the front-end that the settings match what was selected.

Header Settings - mega - center - search

Screenshot 2023-10-31 at 11 08 40 AM

- [x] Go back to the Header Settings form and select the new Focus Nav.
- [x] Attempt to save the form and verify that the Full Screen Homepage Image is required when Focus Nav is selected
- [x] Add a Homepage header image from the media library or upload a new photo
- [x] Save the form and then visit the front end, specifically the front page
- [x] Verify that the front page gets the new homepage image and that the header is greatly simplified. (See note above about front page needing to be a `/node/XX`)

Screenshot 2023-10-31 at 11 21 25 AM

- [x] Visit another page on the site that is not the homepage and verify that the header is still the simplified header, but no homepage image shows at the top

Screenshot 2023-10-31 at 11 22 34 AM

- [x] Open the theme settings drawer by clicking on Theme Settings at the top of the front-end site when viewing a page
- [x] Cange the color palette, header theme, and header accent and verify that the new focus header changes correctly.
- [x] Verify that the theme settings form no longer has the navigation position or navigation type as it once did.
- [x] Log out of the site and login as a platform administrator
- [x] Visit the Header Settings form and verify you see a new option for Site Name Image

Header Settings - platform admin

- [x] Upload an SVG image for this field - verify that no other images are allowed
- [x] Save the header settings and view the front end
- [x] Note: this new Site Name Image works for all three variations of the header
- [x] Verify that the uploaded SVG replaces the site title

Screenshot 2023-10-31 at 3 16 00 PM

- [x] Use the browser inspector and inspect the SVG.
- [x] Verify that the SVG has a `title` tag that matches the site name


dblanken-yale:
@codechefmarc On the header style changes, I'm only seeing change in the "Header Theme", not the color palette or accent.  Should I be seeing a difference?  My steps:

* select a header theme
* select a different color palette
* select a different accent

dblanken-yale:
Also, not sure if it's related to this PR, but I attempted to log out via the username in the upper right menu and am showing as still logged in.  Strange.

dblanken-yale:
@codechefmarc Is there any way to limit the size of the SVG for the site logo?  Or are we considering this a sharp knife?  LOL

Screenshot 2023-11-01 at 11 37 05 AM

SVG used: https://upload.wikimedia.org/wikipedia/commons/6/63/A_large_blank_world_map_with_oceans_marked_in_blue.svg

dblanken-yale:
I've not seen the title tag used in that way.  I thought title was only to be used within the head tag?

dblanken-yale (changes requested):
I made some comments before this one.  Overall great work as always!  Let me know and I'll check through it all again.  :D

codechefmarc:
> @codechefmarc Just making sure I understand the spec; if they choose the new header, they must use the basic right with no search for other pages? I noticed when I changed it to a centered header instead of the new one that other pages mimic'd this. Just making sure I understand what should be there.

Yup - the focus header doesn't do any alignment options. But the config is saved, so if you switch to focus and then switch back, you'll get the alignment on basic and mega.

codechefmarc:
> @codechefmarc On the header style changes, I'm only seeing change in the "Header Theme", not the color palette or accent. Should I be seeing a difference? My steps:
> 
> * select a header theme
> * select a different color palette
> * select a different accent

So, for focus header, the palette and the header theme works, but the accent only shows up on sub-pages, not the homepage because on the homepage, the image has that gradient on top and an accent doesn't really make sense. Happy to go over this together if you're not seeing the header theme or color palette changes as I am seeing them on the multidev.

codechefmarc:
> Also, not sure if it's related to this PR, but I attempted to log out via the username in the upper right menu and am showing as still logged in. Strange.

I ran into this one too here and there. I think you have to close the browser window to end the session? @vinmassaro - is that has CAS works with Drupal and logging out?

codechefmarc:
> I've not seen the title tag used in that way. I thought title was only to be used within the head tag?

Yup, it's mainly for accessibility and with this being the site title and all, it makes sense: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title

I kept thinking about this one because, in theory, someone could upload malicious code in this way - though we are limiting it to platform admins. So, I may want to add some additional processing to strip out javascript tags or something...

vinmassaro:
> > Also, not sure if it's related to this PR, but I attempted to log out via the username in the upper right menu and am showing as still logged in. Strange.
> 
> I ran into this one too here and there. I think you have to close the browser window to end the session? @vinmassaro - is that has CAS works with Drupal and logging out?

@codechefmarc The home page has "CAS Login Required" turned on, so when you click log out, it is ends the Drupal session, but the CAS session is still active. It is redirecting you back to the home page on logout, which checks for the CAS session, and then logs you back in. This redirection all happens very quickly so it looks like you never logged out.

dblanken-yale:
@vinmassaro That makes total sense.  Thanks!  Whew, glad it's not a bug.

dblanken-yale:
@codechefmarc On the SVG filter; ya that might be a good idea.  Looks like it could be low hanging fruit to get around things.    Good thought on it!  Thanks for the link.  That is neat that you can use it though.

vinmassaro:
> @vinmassaro That makes total sense. Thanks! Whew, glad it's not a bug.

I think this may lead to confusion so I'm going to open up a ticket to also end the CAS session

joetower:
@dblanken-yale I pushed up a fix for the SVG. That should be ready for another look. 

https://pr-469-yalesites-platform.pantheonsite.io/

For the color theme issues - the colors not changing in the offcanvas panel when changing global theme. I saw what I think you were describing once, but I'm not seeing it consistently. I'm mainly seeing this: 
https://github.com/yalesites-org/yalesites-project/assets/366413/1e912ff6-cb3c-4e1a-88a2-ea67abcba161

Were you seeing it not working all the time?

dblanken-yale:
@joetower Perfect on the SVG!  MUCH better!

Ya, today I can't recreate it at all.  I was having some strange issues yesterday overall with my browser so I'm chalking it up to that.  Works on Safari and Chrome/Edge.  Nice job!

dblanken-yale (approved):

miketullo95:
hello! taking a look at this. Functionally, the header form is perfect! so far my only comments are front end, and may have reasonings behind these decisions, so let me know! @joetower 

### focus header image cropped 
it looks like some heavy cropping and zooming in is being done on bigger images. see screenshots. 

Focus header image in use (compare with uploaded photo)
Screenshot 2023-11-07 at 10 22 13 AM

uploaded photo
Screenshot 2023-11-07 at 10 22 55 AM

### smaller screen gradient size
on a smaller screen size the gradient seems to be too large
Screenshot 2023-11-07 at 10 24 54 AM

joetower:
@miketullo95 Thank you! 

1. Cropping: this is happening because we're making sure the image is 100% of the viewport (in CSS) instead of actually cropping the image. If we do not want the image to always be 100% of the viewport height, we could specify a specific size to crop the image to or let the image scroll and be dependent on the image uploaded - scaling only the width.  Please let me know if you have any thoughts. I implemented it the way it is because I recall it being discussed during the design process.
2. I can take a look at reducing the gradient.

miketullo95:
@joetower, regarding your first point, I completely agree and believe it's the right path to take. However, I'm wondering if there's something else we could do, perhaps with image sizing, to enhance the appearance of our default images with the focus header. In my example, the image is only displaying about 20% of its content. Do you think if we downscaled the image to a smaller size, more of the picture would appear in the header? My thought with this is the images in storybook look better so it must be something to do with our provided images.

We'll definitely need to create some resources to guide users in selecting appropriate images for this focus header. However, our default images should be as close to perfect with the focus header.

joetower:
@miketullo95 This is ready for another look, please. I pushed up an update to: 
- make the background-images a ratio of 16:9 instead of 16:5 so they are taller
- adjusted the overlay gradient position at smaller resolutions

https://pr-469-yalesites-platform.pantheonsite.io/

Please let us know if this is good to go or if you have additional feedback. Thank you!

miketullo95 (approved):
@joetower great work! this looks awesome

PR #471, opened 2023-10-31
YALB-1588: Starterkit: Install ys_starterkit content on new sites

vinmassaro:
### Description of work
- Creates a github repo to store the exported files: https://github.com/yalesites-org/yalesites-starterkit
- Document process a README for updated the exported files.
- Write an installation script for new sites to pull this content in.
- Ensure the process cleans up after itself.
- Manually update the default values of the levers in the Drupal Upstream. https://github.com/yalesites-org/yalesites-project/blob/develop/web/profiles/custom/yalesites_profile/config/sync/ys_themes.theme_settings.yml

### Functional testing steps:
- [ ] Locally, run `npm run setup`, which should run the content importer while setting up a local environment
- [ ] Branch off of this and create a multidev to test that CI also does this.
- [ ] On the multidev, modify a block that can have a link (Like text), and verify that when you use an alias like `/welcome` it does not convert it to an internal /node/x URL.


composer.json
nJim, line 1: There are a few ways to do this. I tend to use the generated patch from the fork. You can find a link to this in the issue itself with the text 'plain diff'. Using patches is a good way to keep track of small changes that we want to get rid of. Typically I will only add a new repository to composer.json if I intend to used the forked version of a project for a long time. 

https://git.drupalcode.org/project/search_api_html_element_filter/-/merge_requests/1.diff

scripts/local/setup.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Define variable to check from other scripts if this script is being invoked.
+export YALESITES_INSTALL=1
```
nJim, line 4: Let's plan to either rename this something that describes the utility such as IS_LOCAL 
OR we can try another check in the content-import.sh script to see if lando is available. It would be best to programmatically check if we are on local if possible.

scripts/local/starterkit/create-export
```diff
@@ -0,0 +1,121 @@
...
+export-content() {
+    local zipRemoteLocation
+    local zipFilename
+    local destinationFolder="${1:-$(mktemp -d)}"
+
+    zipRemoteLocation="$(retrieveZipLocation "$(runExport)")"
+    zipFilename="$(basename "$zipRemoteLocation")"
+    retrieveFileFromSftp "$zipRemoteLocation"
+    if [[ -f "$zipFilename" ]]; then
+        unzipToFolder "$zipFilename" "$destinationFolder"
```
nJim, line 46: Let's remove all existing files before unzipping the new ones. (Per David's suggestion)

scripts/shared/content-import.sh
nJim, line 1: We can skip these steps. Instead the values can be set in config within the profiles. We should check the levers config, alerts config, and site settings config are updated to match what we have in the ys-starkerkit website.

scripts/shared/content-import.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+STARTERKIT_VERSION="v1.0.0"
```
nJim, line 3: Instead of specifying this version, let's just point this to the latest release programatically. If we want, we can make this an optional cmd line argument if someone wants to specify a version.

scripts/local/starterkit/create-export
```diff
@@ -0,0 +1,121 @@
...
+    # We don't have permission to create the scs-export directory under
+    # /code/web, so it errors out.  We redirect the error to stdouot so we can
+    # capture the file name that was attempted.  We then use the tmp directory
+    # location that it was when it was created, which still exists.
+    zipFilename=$(echo "$1" | head -5 | tail -1 | grep -o "temporary://[^']*" | xargs -I {} basename {})
+    echo "$tmpPath/$zipFilename"
+}
+
+runExport() {
+    terminus drush "$PANTHEON_NAME" -- content:export --all-content --assets --translate 2>&1
```
nJim, line 72: Consider renaming these to follow other naming conventions for other scripts. For exmaple 'machine name'.

dblanken-yale:
I was able to do a successful import locally with the 1.4.4 change.  I think we are ready unless @nJim or @vinmassaro can think of anything else?  If not, I'll merge today!

codechefmarc (commented):
Still reviewing. Notes in progress:

I tested this from an existing site, thinking I could simply run `npm run setup` and have it create a fresh site with content on install, but that didn't work. I got the following error:

```
Executing drush content:import ../starterkit.zip

> 
> In FileSystem.php line 405:
>                                                                                                                                      
>   The specified file 'temporary://import/zip/98f48787-a200-487b-bf8e-91a3a2890501/assets/' could not be moved to 'public://assets'.  
>                                                                                                                                      
> 
 [error]  Message: Output is empty.
```

I didn't dig too deep, but it looked like it may be file system permission issues?

But, I decided to wipe out my local completely and start over from scratch and that worked great - I got a new fresh site with the correct content on install from the starter kit site.

I'm not exactly sure how to test the CI job as if I branch off of this branch and create a new multidev, it will simply get the content from the `https://dev-yalesites-platform.pantheonsite.io/` no? How would I force a new site install / content import on a multidev?

scripts/shared/content-import.sh
```diff
@@ -0,0 +1,28 @@
...
+  echo "Failed to download starterkit file."
+  exit 1
+fi
+
+# Check if running under lando, otherwise assume CI.
+# Put starterkit file in place and import.
+# Clean up, and set front page in config.
+if [ "$YALESITES_IS_LOCAL" == "1" ]; then
+  lando drush content:import ../"$STARTERKIT_FILE"
+  lando drush cset system.site page.front '/homepage' -y
```
codechefmarc, line 20: We had talked a little bit about this one to set this as `/node/xx` instead of `/homepage`. We could either run some PHP code on the CLI via `php eval` or perhaps write a custom drush command to parse a path like `/homepage` and turn that into the internal path needed `/node/xx` to then use for the config we need to set here. I'm happy to look into creating a drush command if you'd like as I've never done that before and would be kind of neat to see how to get that to work.

vinmassaro: Is Linkit the problem here? A clean D10 install of the standard profile allows us to use an alias without a problem, so I think we should make this work.

Screenshot 2023-12-11 at 4 32 52 PM

codechefmarc: OK, I just tested Linkit and it correctly does choose the `/node/xx` when using autocomplete. The core Drupal site settings form (which we aren't using) will convert it to a `/node/xx` in config even though in the form, it shows the alias.

I think the Linkit change is new, as IIRC, it didn't do that before - it saved the alias before. So, the main issue is for any new sites and if we set it to `/homepage` in config via the script, it won't get the `/node/xx`. You have to go into the Linkit form to set it correctly.

I put this in Teams, but I just created a drush command (as part of YS Core) to get the internal path when you specify a path alias, so we could use that to get the correct internal path and set that in the script dynamically. `drush intpath /alias-to-page`

vinmassaro: @codechefmarc I found an active issue in the Linkit queue for this: https://www.drupal.org/project/linkit/issues/2877535

I tested patch # 64 locally and it seems to work fine setting `/homepage`. While the Drush command would work, I think it isn't needed if we use the patch, and this should eventually get fixed in that issue.

codechefmarc: So, I retested with the change of Linkit, and it still doesn't work for me when doing a clean site install using the starter kit scripts. This is because we're setting the config value of `system.site` `page.front` to a string `/homepage` and not specifically the internal node id which it needs to not have the `/homepage` in the URL when visiting the front page. So, we should be setting it to `/node/xx` like Drupal does internally.

Also, this change with Linkit causes a reversion where it, too, is saving the config as a string:

```
page:
  403: ''
  404: ''
  front: /homepage
```

Whereas if I set it via Drupal's form:

```
page:
  403: ''
  404: ''
  front: /node/4
```

So, I think we do need to either use the Drush command or find someway to save the homepage config when exporting content from the starter kit site?

codechefmarc: Oh, this affects not only the URL having the extra path on the front page, but also, with the new focus header, there is a core function to check to see if you are on the homepage and that only works if the config is set to `/node/xx` and won't work if it's set to an alias. It's annoying, hah.

codechefmarc:
@vinmassaro - Tested today. From clean, ran `npm run setup` and that worked great. From existing, ran `npm run build-with-install` and that also worked beautifully. Also, the home page is set correctly in the config from these setup scripts.

Linkit still doesn't save the homepage correctly as`/node/xx` but, we talked about that one and that is probably a future fix either from core or Linkit.

dblanken-yale:
This looks great!  Nice edits!  I have also ran locally without issue.

dblanken-yale:
@codechefmarc I know you already weighed in, just wanting to get the approval to merge it.  :D Unless you want Jim to do it.

nJim (approved):

PR #484, opened 2023-11-16
Add conditional logic to `drush deploy` quicksilver script

vinmassaro:
### Description of work
`drush deploy` that runs in CI was getting no output because it was first running on the platform from Quicksilver. This updated script uses Pantheon's workflow API endpoint to check if the sync_code hook was initiated by our machine user or not. This commit also renames this file and references in pantheon.yml, since it is now doing more than just a config import.

### DIFF
--- a/pantheon.yml
+++ b/pantheon.yml
@@ -5,15 +5,15 @@ workflows:
   clone_database:
     after:
       - type: webphp
-        description: Import configuration from .yml files
-        script: private/scripts/confim.php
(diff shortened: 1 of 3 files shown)


nJim (approved):
I'm so glad you found this path. This looks great. I wouldn't have know what is available at this level so I'm glad you added the extensive documentation to this script. Code looks solid. Marking as approved. We can keep an eye things like `cim` running as this gets out into the wild.

PR #486, opened 2023-11-17
YALB-1537 AX: Remove Items from multivalue fields

hectorlj:
### Description of work
- Installs and enables `multiple_fields_remove_button` module to provide a user friendly way to remove fields that were previously populated.

### Functional testing steps:
- [ ] Login as a site administrator
- [ ] Edit the content of a page in the layout builder interface
- [ ] Add a `Quick Links` block and add multiple links
- [ ] Save your changes
- [ ] Edit the page with your newly added `Quick Links` block in the layout  builder interface
- [ ] Configure the new `Quick Links` block
- [ ] Verify that there is a delete button on each link field 
image
- [ ] Verify that the delete button removes any link that is marked as required
- [ ] Save the `Quick Links` block without any errors
- [ ] Verify that the delete button does not appear on other multivalue fields, such as the Image Grid Items in the `Media Gallery` block


nJim (approved):
Works as intended. A small but important change. We may want to alter the styles of this button in the future, but we can handle that in a larger admin-theme epic. Shipping it.

PR #488, opened 2023-11-19
YALB-1584: Update dashboard content

nJim:
### Description of work
- Updates the content of the dashboard to replace the placeholders and remove the Service Now link.
- Moves block templates to the module. This should appears for anything so it should not appear in the admin theme.
- Removes any unnecessary blocks, templates, hooks as this has become simplified. Everything is now in a single template for a single controller.

### Functional testing steps:
- [ ] [Login as a site admin](https://pr-488-yalesites-platform.pantheonsite.io/cas)
- [ ] [Navigate to the YaleSites Dashboard](https://pr-488-yalesites-platform.pantheonsite.io/cas)
  - [ ] Make sure links are working
  - [ ] Mike T can verify that content matches expectations. This is basic right now because we have not invested much time due to competing priorities.

Screen Shot 2023-11-20 at 10 25 36 AM


codechefmarc (commented):
Tested this as well and it looks good to me. My only note is to make sure we don't lose the syntax for the ServiceNow link to submit a ticket as I would assume this will get used again sometime in the future.

PR #490, opened 2023-11-21
YALB-1634: PHP warning logged on admin pages post-D10

dblanken-yale:
### Description of work
- Create npm script to copy the tokens from `node_modules` into the theme's `css` directory for commit
- Update `npm run develop` to also copy the tokens so we always have a fresh copy.

### Question
Should we always use latest tokens or should we still limit it to the 1.17.x version?

### Functional testing steps:
- [ ] Visit an admin configuration page
- [ ] Verify the following error does not display itself:

`Warning: file_get_contents(themes/custom/ys_admin_theme/node_modules/@yalesites-org/tokens/build/tokens.css): Failed to open stream: No such file or directory in Drupal\Core\Asset\CssCollectionOptimizerLazy->generateHash() (line 43 of /code/web/core/lib/Drupal/Core/Asset/AssetGroupSetHashTrait.php)`


dblanken-yale:
Actually, talked with @codechefmarc more about it and we're not sure that this is the true issue; we've not had to do something like this before, and even in other PRs there's no loading of the node_modules version.  We're going to continue to poke around and maybe include @joetower to see what he thinks it could be also.

dblanken-yale:
Closing; the issue was actually due to a problem in atomic: https://github.com/yalesites-org/atomic/pull/186/commits/df5c83d8d5afdc5d4b69a94b28aef8eae0e133f4

Nice work @codechefmarc!!!!!!

PR #492, opened 2023-11-23
YALB-1169 Links: Enable WYSYWIG tools for anchor links

hectorlj:
### Description of work
- Installs and enables `anchor_link` module
- Adds the anchor link button to the `basic_html` and `restricted_html` text formats

### Functional testing steps:
- [x] Login as a site administrator
- [x] Edit the content of a page in the layout builder interface
- [x] Do the following instructions for each block component and its corresponding field
  - [x] Text -> Content
  - [x] Wrapped Image -> Content, Caption
  - [x] Content Spotlight -> Content
  - [x] Image -> Image Caption
  - [x] Quick Links -> Quick Link Content
  - [x] Video -> Video Description
- [x] Verify that the anchor link button is visible in the field's text editor
image
- [x] Add an anchor(s) along with content. Content will help us verify the anchor is functioning properly.
- [x] Verify that anchors added in the component are not visible when the content is published
- [x] Verify that accessing the anchors loads in the correct location
  - e.g. If anchor `anchor1` is set to the bottom of a page `Test Page`, then loading the url `/test-page#anchor1` will load at the bottom of `Test Page`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,6 +10,18 @@
     "drupal": {
       "type": "composer",
       "url": "https://packages.drupal.org/8"
+    },
+    "ckeditor5-anchor-drupal": {
+      "type": "package",
+      "package": {
+        "name": "northernco/ckeditor5-anchor-drupal",
+        "version": "0.4.0",
+        "type": "drupal-library",
+        "dist": {
+          "url": "https://registry.npmjs.org/@northernco/ckeditor5-anchor-drupal/-/ckeditor5-anchor-drupal-0.4.0.tgz",
+          "type": "tar"
+        }
+      }
     }
   },
   "require": {
--- a/composer.json
+++ b/composer.json
@@ -10,18 +10,6 @@
     "drupal": {
       "type": "composer",
       "url": "https://packages.drupal.org/8"
-    },
-    "ckeditor5-anchor-drupal": {
-      "type": "package",
-      "package": {
-        "name": "northernco/ckeditor5-anchor-drupal",
-        "version": "0.3.0",
-        "type": "drupal-library",
-        "dist": {
-          "url": "https://registry.npmjs.org/@northernco/ckeditor5-anchor-drupal/-/ckeditor5-anchor-drupal-0.3.0.tgz",
-          "type": "tar"
-        }
-      }
     }
   },
   "require": {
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -10,6 +10,18 @@
     "drupal": {
       "type": "composer",
       "url": "https://packages.drupal.org/8"
+    },
+    "ckeditor5-anchor-drupal": {
```
hectorlj, line 14: @nJim This question for after Thanksgiving 😄 And also to add some background incase someone else ends up picking this up. 

The anchor link module relies on this package to work properly, but I was getting errors adding it to the profile composer.json file. However adding it here allowed it to install without issue. Is it alright to leave this here, or would it be better to find a way move it to the profile composer.json file?

codechefmarc: Hi @hectorlj! Yup, since the profile `composer.json` is the correct place to add the `ckeditor-anchor-drupal` repo, you also have to add that `ckeditor-anchor-drupal` to the `repositories` section in the profile `composer.json` to get that one to load correctly. This is for security reasons to make sure you aren't loading other repos you didn't intend.

hectorlj: Ah, gotcha. Thanks @codechefmarc!

codechefmarc: One last thing, then you've got it - please move the `"northernco/ckeditor5-anchor-drupal": "^0.3.0",` from the root `composer.json` to the profile `composer.json`. Thank you!

hectorlj: @codechefmarc Just double checked the composer docs and it looks like the packages in the `repositories` section is only read from the root `composer.json`... I'll double check to see if there's another way to get that package installed
https://getcomposer.org/doc/04-schema.md#repositories

codechefmarc: Thanks for looking into it, @hectorlj! Let me know what you find. It's been a while since I did something like that so I forget exactly the placement, but I wonder if it needs to be in both places? I think I tested that and it worked? I hate to have redundant code though.

hectorlj: @codechefmarc Looks like it works when it's in both places 🤷🏼‍♂️ I did try using `wikimedia/composer-merge-plugin` to use the composer file in the anchor link module, but it had its own issues with the root composer file

codechefmarc (approved):
Works for me, approved!

PR #494, opened 2023-11-27
YALB-1648: Bug: empty footer logo instance causes PHP fatal error

vinmassaro:
### Description of work
A user reported an issue with their site throwing fatal errors:
```
Uncaught PHP Exception Error: "Call to a member function getFileUri() on null" at /code/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesFooterBlock.php line 87
```
They had one footer logo uploaded and the other instance of the field was returning as `null`. This checks to make sure an instance of the field is not null.


codechefmarc (approved):
Tested this on the multidev - added 1 and 2 images, tested, then deleted one of the media w/o removing it from the footer. Site still worked. Approved!

PR #506, opened 2023-12-13
YALB-864/YALB-1650: Global reusuable blocks

codechefmarc:
## Tickets
[YALB-864: Reusable Content: Global component blocks](https://yaleits.atlassian.net/browse/YALB-864)
[YALB-1650: Reusable Content: Browser UX](https://yaleits.atlassian.net/browse/YALB-1650)

### Description of work

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/ys_views_basic.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/ys_views_basic.module
@@ -87,31 +87,4 @@ function ys_views_basic_form_alter(&$form, $form_state, $form_id) {
     }
   }
 
-  /* Add messaging with link to edit global block and destination back to
-    layout builder.
-   */
-  if ($form_id == 'layout_builder_update_block') {
-
-    // block_form is not set on global blocks.
-    if (!isset($form['settings']['block_form'])) {
-      // We need the global block uuid to look up the block id.
-      $blockUUID = explode(":", $form_state->getBuildInfo()['callback_object']->getCurrentComponent()->get('configuration')['id']);
-      $blocks = \Drupal::entityTypeManager()->getStorage('block_content')->loadByProperties([
-        'uuid' => $blockUUID[1],
-      ]);
-      $blockId = array_key_first($blocks);
-
-      // We need the current path to extract the node id.
-      $current_path = \Drupal::service('path.current')->getPath();
-      preg_match('/node\.(\d+)/', $current_path, $matches);
-
-      $parentBlockTitle = $form['settings']['admin_label']['#plain_text'] . ' global block';
-      // Compiled global block message.
-      $form['settings']['global_info'] = [
-        '#markup' => "<p>Note: This is a global block. Any changes made to a global block will be reflected anywhere it is used.</p> <a class='glb-button glb-button--primary' href='/admin/content/block/{$blockId}?destination=/node/{$matches[1]}/layout'>Edit {$parentBlockTitle}</a>",
-      ];
-    }
-
-  }
-
 }
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/ys_views_basic.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/ys_views_basic.module
@@ -81,8 +81,14 @@ function ys_views_basic_views_pre_build($view) {
  */
 function ys_views_basic_form_alter(&$form, $form_state, $form_id) {
   if (str_starts_with($form_id, 'layout_builder_')) {
-    if ($form['settings']['block_form']['#block']->bundle() == 'view') {
+    if (isset($form['settings']['block_form']) && $form['settings']['block_form']['#block']->bundle() == 'view') {
       $form['#in_layout_builder'] = TRUE;
+    }
+  }
+  if ($form_id == 'layout_builder_update_block') {
+    if (!isset($form['settings']['block_form'])) {
+      // $form['']
+      // dpm($form_state);
     }
   }
 }
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/ys_views_basic.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/ys_views_basic.module
@@ -85,10 +85,10 @@ function ys_views_basic_form_alter(&$form, $form_state, $form_id) {
       $form['#in_layout_builder'] = TRUE;
     }
   }
+  // Add messaging to global blocks to inform content editors where to go to edit.
   if ($form_id == 'layout_builder_update_block') {
     if (!isset($form['settings']['block_form'])) {
-      // $form['']
-      // dpm($form_state);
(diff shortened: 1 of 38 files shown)


miketullo95:
some quick notes so I don't forget, want to look into this in more depth tomorrow.

- could we list all reusable blocks (in the block picker) in one section below advanced instead of separating them? I could see if an organization has like 20+ reusable blocks separating them by block types would be good, but it may be more common to have fewer. So having them in one section may be easier way finding
- is there an easy way to find how many instances of a global block are used? could be interesting to see how often its used. If this is possible, we could show it in the manage page and also in places like screenshot below. Something like "Any changes made to this global block will be reflected in all X instances used on your website".
Screenshot 2023-12-14 at 4 47 30 PM
- when adding a global block, could we put the blocks in the picker groups? (basic, media, advanced, etc) on this page 
Screenshot 2023-12-14 at 4 40 08 PM
- I'm wondering if we should do some visual treatment to global blocks in /layout. maybe having the dashed line around the block be a different color (light blue?) + an icon (globe if we stick with this name lol)
- some fun css here, only happening now because we can name the blocks pretty long
Screenshot 2023-12-14 at 4 46 08 PM

This looks amazing! such an exciting feature and will be used A LOT by our community

kara-franco (commented):
Wicked amazing work! I'm so excited for this feature. So far, accessibility testing is great, I'm going to test with a screen reader more on Monday. One thing I noticed that could improve the cognitive load on users is for the "Add content block" order of blocks to match out block picker (rather than be in alphabetical order). 

Screenshot 2023-12-15 at 1 10 16 PM

Also, I totally third Sebasitanna's idea of adding a toggle to make a block sharable in Layout Builder. The back and forth is tedious/not intuitive for users.

kara-franco (approved):

dblanken-yale:
I love this!  I especially like the globe to show that it's a global block.  A question though: Should we have some indicator in the non-content preview version?  Right now it does give the name of the global block, but doesn't specify anywhere that it's global?  Should it?
Screenshot 2024-01-03 at 10 23 40 AM

dblanken-yale (approved):
Views work great!  I love the indicators to show that it is a global block and the message to convey that changes made will be made globally.  REALLY great job!!!!!!  Feels like something that should go in core eventually.  I did have a question regarding non-content display and whether to display that it is a global block, but I don't feel it's blocking this from going through, so approving!

miketullo95:
@codechefmarc 

- global blocks configure form has some alignment things going on (see below, first is global, second is non-global). This happens with most block types that I tested.

Screenshot 2024-01-03 at 11 23 29 AM
Screenshot 2024-01-03 at 11 23 46 AM

- global views - display all options still show limit to # of items 
Screenshot 2024-01-03 at 11 26 08 AM

- I agree with @dblanken-yale, maybe something like a prefix of "Global [Block name] Block" for the admin label

codechefmarc:
@dblanken-yale - I've added the globe icon to the blocks when content is not in preview:

Screenshot 2024-01-03 at 3 51 41 PM

@miketullo95 - I've fixed the padding issues around the global blocks vs non global blocks so they should all look the same when editing.

miketullo95 (approved):
approved! so excited about this feature!

web/profiles/custom/yalesites_profile/config/sync/gin_lb.settings.yml
```diff
@@ -1,5 +1,8 @@
 _core:
   default_config_hash: 4DJ-ODpqIpFYPDmqzRIVHczK321I99zFqNgPzKwVibs
-toastify_loading: cdn
+toastify_loading: custom
```
nJim, line 3: What's going on here? I didn't see a custom packaged being loaded anywhere.

web/profiles/custom/yalesites_profile/config/sync/views.view.block_content.yml
```diff
@@ -21,7 +21,7 @@ display:
     display_plugin: default
     position: 0
     display_options:
-      title: 'Custom block library'
+      title: 'Global Blocks'
```
nJim, line 24: I think we are changing all of the 'globals' to 'reusable'. I'm guessing you are already working on this.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/ys_views_basic.module
```diff
@@ -70,3 +70,21 @@ function ys_views_basic_views_pre_build($view) {
...
+function ys_views_basic_form_alter(&$form, $form_state, $form_id) {
+
+  /* The views tool uses Ajax and the form elements and arrays are different
+  when loaded in layout builder vs the global Drupal block edit form. This
+  will add a key that we use in ViewsBasicDefaultWidget to determine the
+  correct arrays to use for Ajax calls.
+   */
+  if (str_starts_with($form_id, 'layout_builder_')) {
+    if (isset($form['settings']['block_form']) && $form['settings']['block_form']['#block']->bundle() == 'view') {
+      $form['#in_layout_builder'] = TRUE;
```
nJim, line 86: Can you please double check that this is required? Or can we find this information from within the widget itself and avoid setting this flag?

web/profiles/custom/yalesites_profile/config/sync/gin_lb.settings.yml
codechefmarc: This is just for Gin Layout Builder adding the toastify messages. With it being turned off like this, the status messages are left inline rather than disappearing. This is helpful for things like the reusable notice when editing a reusable block that notice will disappear. Also, it helps with Hector's work with the layout builder lock notification not disappearing and allowing someone to release the lock.

dblanken-yale:
@codechefmarc Are we wanting to address deletion of a global block in this?  This is what I got when I deleted the global divider I made when it was being used.  Or should we somehow enforce some sort of referential integrity to say you can't delete a used global block?

Screenshot 2024-01-22 at 8 41 05 AM

codechefmarc:
> @codechefmarc Are we wanting to address deletion of a global block in this? This is what I got when I deleted the global divider I made when it was being used. Or should we somehow enforce some sort of referential integrity to say you can't delete a used global block?

I see what you mean - we can change the message, I'm sure. But, what's nice is, the message only shows up for logged in users. If you're anonymous, you don't see it at all. I'll look into changing the message. @miketullo95, do you have a preference for the error message if a global block gets deleted and it's still be used someplace? See David's screenshot.

miketullo95:
> I see what you mean - we can change the message, I'm sure. But, what's nice is, the message only shows up for logged in users. If you're anonymous, you don't see it at all. I'll look into changing the message. @miketullo95, do you have a preference for the error message if a global block gets deleted and it's still be used someplace? See David's screenshot.

"This Reusable Block has been deleted. Remove this instance of the deleted block."

codechefmarc:
@dblanken-yale - Ok, the message for broken blocks has been updated.

codechefmarc:
@miketullo95 - I do like this idea, however, because the work is still coming from a patch to core, I would think adding this functionality now would make getting the patch through to core take a lot longer (The issue for this is already 6 years old). That said, I do think that it's definitely doable. Looks like some folks are working on that piece in a separate issue: https://www.drupal.org/project/drupal/issues/3303306

PR #509, opened 2023-12-18
YALB: 1429 Fix wysiwyg growing indefinitely

laura-johnson:
### Description of work
- Adds scss/css to the admin theme to set the missing drupal displace variables.

### Functional testing steps:
- [ ] Add a text block to a page, verify that it never exceeds the height of the viewport. 
- [ ] If you go here on the multidev and edit/configure the block in layout builder or add a new one, you should see that it does not expand indefinitely

https://pr-509-yalesites-platform.pantheonsite.io/node/92/layout


codechefmarc (approved):
Changes to Atomic did fix the overflow issue! Also tested on the general block add page too. Approved for this PR and then if you can create a PR for Atomic, I can go and review/approve that one as well.

PR #511, opened 2023-12-20
YALB-1645: Bug: Concurrent editing in Layout Builder

hectorlj:
### Description of work
- Adds layout builder locking to prevent concurrent edits from other users

#### As user 1
- [x] Login as a site administrator
- [x] Edit the content of a page in the layout builder interface
- [x] Verify that a pop-up appears notifying the user that the content is now locked by them image
- [x] Verify blocks can be added and saved, and that editing is otherwise unchanged
#### As user 2
- [x] In a separate browser, or incognito window, do the following
  - [x] Login as a different user
  - [x] Edit the same page content in the layout builder interface
- [x] Verify that a pop-up appears notifying users that the content is currently locked and the user image
- [x] Verify that any type of content editing is prevented.
#### As user 1
- [x] Unlock the page content using the unlock button in the layout builder settings  image
- [x] Confirm lock break
- [x] Message informs user that lock has been removed.
#### As user 2
- [x] Refresh page containing the page's layout builder interface
- [x] Content is now locked by user 2
- [x] User 2 can now make edits


codechefmarc (approved):
Tested and this looks good! Approved! I brought up in stand up today that I think we should disable tastily messages for this because the message that talks about the lock and how to unlock is only there for a brief moment and then disappears. Not part of this ticket, so no changes needed from you.

One other thing, I didn't see `content_lock` in the profile `composer.json` but saw it was enabled and some patches applied. I think this is fine for now, but we may want to specify the specific version of it in our `composer.json` to pin it like we do the others.

kara-franco:
Heya! @nJim Is this work complete? Can the fix be added to the release?

kara-franco:
So, @dblanken-yale , @miketullo95  and I did some concurrent testing and were going through the user flow. A user will need to click unlock after saving. Can we have the unlocking happen when Save is selected? 

Or can the lock end after a user session has ended? Like if someone closes a tab? Or use a chronjob?

I'd love to hear any other options to lessen the amount of steps for users.

dblanken-yale:
Hi @hectorlj!  Nice to virtually meet you.  :D  Heard great things about you and your work!

Yes, we saw that `content_lock` comes with coupled with `content_lock_timeout`.  Maybe that could be the ticket to clear out any stale sessions that have locks?  I haven't had a chance to dive into that much to see what it can do yet.

nJim:
Closing this work. There is still a need to have this kind of feature, but the current module does not work well with layout builder. The autosaving features of this interface are making it difficult to lock and break the lock in a predictable way. We can customize this module in the future if we want to prioritize this work.

PR #514, opened 2024-01-02
feat(YALB-1674): add new content spotlight portrait component

joetower:
### Description of work

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/block_content.type.content_spotlight_portrait.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/block_content.type.content_spotlight_portrait.yml
@@ -1,0 +1,33 @@
+uuid: b6c1e225-d1e1-4b06-a972-d6223b25c923
+langcode: en
+status: true
+dependencies:
+  module:
+    - entity_redirect
+third_party_settings:
+  entity_redirect:
+    redirect:
+      add:
+        active: 0
+        destination: default
+        url: ''
+        external: ''
+      edit:
+        active: 0
+        destination: default
+        url: ''
+        external: ''
+      delete:
+        active: 0
+        destination: default
+        url: ''
+        external: ''
+      anonymous:
+        active: 0
+        destination: default
+        url: ''
+        external: ''
+id: content_spotlight_portrait
+label: 'Content Spotlight (Portrait)'
+revision: 0
+description: 'Content spotlight - Portrait allows you to display a portrait 2x3 image adjacent to text. This component has optional content, an image offset option, and different orientations.'
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.block_content.content_spotlight_portrait.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.block_content.content_spotlight_portrait.default.yml
@@ -1,0 +1,133 @@
+uuid: 7336cef2-ae23-49a3-969a-e32b5bc98708
+langcode: en
+status: true
+dependencies:
+  config:
+    - block_content.type.content_spotlight_portrait
+    - field.field.block_content.content_spotlight_portrait.field_heading
+    - field.field.block_content.content_spotlight_portrait.field_instructions
+    - field.field.block_content.content_spotlight_portrait.field_link
+    - field.field.block_content.content_spotlight_portrait.field_media
+    - field.field.block_content.content_spotlight_portrait.field_style_position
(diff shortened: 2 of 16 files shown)


miketullo95:
this looks awesome! here are my comments @joetower 

- move theme color drop down to below link field group and above image position alignment 
- are these dial options names different from content spotlight because they can't be the same? (image position vs image position alignment)
- default option for theme color is just color 1 (remove default)
- I saw the discussion about making this a separate block from content spotlight last month, and I agree but I think the original content spotlight could also benefit from the background color dial as well.

joetower:
@miketullo95 Thank you! I'll make those adjustments. 

Follow up questions: 

> are these dial options names different from content spotlight because they can't be the same? (image position vs image position alignment)

Because we have a position style and a position alignment, I wanted to make them clearer. What words would you suggest we use to communicate that image-left/image-right are positions and inline/offset are a style on the image?

---

> default option for theme color is just color 1 (remove default)

It was my understanding that this component could still be used within the flow of content - which would mean that it needs to not use a color theme (for example, no background color and it does not have top/bottom padding and different margins). Do you see this only being used as a standalone element? I left the default option so the component could be used in the regular flow of content, like the Content Spotlight. 

---

> I saw the discussion about making this a separate block from content spotlight last month, and I agree but I think the original content spotlight could also benefit from the background color dial as well.

I can add the background color dial to the regular Content Spotlight, no problem. I'd use the same settings we decide upon for the themes - `default`, `one`, `two`, `three`. That is, again, unless you feel this, too, should always be used as a standalone element on the page. 

---
Please let me know what you think.

miketullo95:
@joetower 
> Because we have a position style and a position alignment, I wanted to make them clearer. What words would you suggest we use to communicate that image-left/image-right are positions and inline/offset are a style on the image?
> 
I'm wondering since they are the same function for other blocks (wrapped image, content spotlight, etc), it may be clearer to the user what this action is if they have the same name. Maybe Sebastianna has opinions on this 

> It was my understanding that this component could still be used within the flow of content - which would mean that it needs to not use a color theme (for example, no background color and it does not have top/bottom padding and different margins). Do you see this only being used as a standalone element? I left the default option so the component could be used in the regular flow of content, like the Content Spotlight.
> 
Ohhh you are completely right. So default should be the "default" which is no color. Maybe we can suggest the option would read "Default - No Color".

> I can add the background color dial to the regular Content Spotlight, no problem. I'd use the same settings we decide upon for the themes - `default`, `one`, `two`, `three`. That is, again, unless you feel this, too, should always be used as a standalone element on the page.
>
Yeah! that would be awesome

joetower:
@miketullo95 This is ready for you to take another look, please. 

I pushed fixes to: 
- Make the dial label for `Image position` consistent between Content Spotlight and Content Spotlight (Portrait)
- Make the dial label for `Image style` consistent with Wrapped Image and Content Spotlight (Portrait)
- Move the theme color dial to appear below the link field
- Changed the label for the default color theme to read `Default - No Color`

codechefmarc (approved):
This looks really good - I understand the need for the different component now after having played with both of them on the multidev. I also like the addition of the colors to the regular content spotlight as well. Approved!

miketullo95:
amazing work! final request via the teams chat we had this morning:

- [ ] change new block name to Portrait Spotlight
- [ ] change old block (content spotlight) to Landscape Spotlight
- [ ] add new block picker preview images to both "spotlights"
portrait spotlight
landscape spotlight

miketullo95:
@joetower SVGS 
portrait spotlight
landscape spotlight

joetower:
@miketullo95 This should be ready for another look. I updated the preview icons and labels. 
https://pr-514-yalesites-platform.pantheonsite.io/

Screenshot 2024-01-10 at 1 52 13 PM

dblanken-yale (approved):
This is so cool to see and how much work goes into wiring a component in here.  Wonderful work.  I'm going to look a bit closer to learn more, but it looks like everything is wired up correctly, and that the overrides are set.

Making sure I understand--because the UUID was not changed, all existing content spotlights will just take the new settings and work with it since we only changed the label and description.  Love it.

Great work as always!

miketullo95 (approved):
happy to approve this! one quick question, im demo'ing this to Yale publishing either tomorrow or Tuesday. Is there a way to have a multidev link of this after it's merged in to show it off? @joetower 

or if this PR will live until then, that works too. haha

PR #515, opened 2024-01-02
YALB-959 min max referenced items

laura-johnson:
### Description of work
- Adds JS validation to block content forms and layout builder forms to validate number of items added is within the desired range for each block type.

Quick Links: min 3, max 9 
Tabs: min 2, max 5 
Media Gallery: min 2, unlimited max
Image Grid: min 2, unlimited max

### Functional testing steps:
- [ ] Add each type of block via layout builder, verify if number of items does not fall within the parameters, you see JS message and form will not submit
- [ ] Edit each type of block via layout builder, verify if number of items does not fall within the parameters, you see JS message and form will not submit
- [ ] Add each type of block via add block type, verify if number of items does not fall within the parameters, you see JS message and form will not submit
- [ ] Edit each type of block, verify if number of items does not fall within the parameters, you see JS message and form will not submit


codechefmarc (approved):
This looks really good and works well. I like how you added the current number of items to the error message as well. In reviewing this ticket, I researched some other modules like field validation and Clientside Validation. Field validation works to replicate this, but it's server side only and the errors will show at the top of the (possibly scrolled) block edit form so content folks can't see the error. So, your work improves on this. I was thinking for the future, perhaps some sort of JSON-based file to allow us to update different validators for multiple kinds of fields. But that is future us problem!

tldr: approved!

laura-johnson:
Thanks @codechefmarc! Yeah I tried clientside validation and it seemed like it would not work for this. I didn't know about field validation! I'm going to take a look at that one to see if that might work for a similar case next time... thanks for mentioning those.

PR #522, opened 2024-01-08
YALB-1588: Media links redirect to files

nJim:
### Description of work
- Requires and enabled the media_entity_download module to create a download route for each media item.
- Creates a new route for media entities where folks linking to these will get a file download link instead of viewing the entity. Download links are set to inline meaning they will try to open in the same tab.
- Side quest: Changed setting to allow unpublished notes to be linked to with Linkit.

### Functional testing steps:
- [x] [Login as a site admin](https://pr-522-yalesites-platform.pantheonsite.io/)
- [x] Navigate to the [edit-layout page for a page](https://pr-522-yalesites-platform.pantheonsite.io/node/1/layout)
- [x] Edit a text block or any other block with a WYSWIYG text area
- [x] Highlight some text and click on the link tool to create a link with the link tool.
- [x] Enter the word 'Reference' into the autocomplete.
- [x] In the 'Media' section, select the item "Yalesites-Blocks-Reference-Sheet.pdf" to make this link.
- [x] In another part of the WYSIWYG, highlight another word to create a second link. 
- [x] Again, enter the word 'Reference' into the autocomplete.
- [x] In the 'File' section, select the item "Yalesites-Blocks-Reference-Sheet.pdf" to make this link.
- [x] Save this block and this node.
- [x] View the page as an anonymous user. Verify that both links (the one created with a file reference and the one created with a media reference) both link to the PDF.
- [x] Verify that for both cases the PDF opens in the same browser window (it does not try to force download the file.)

Screen Shot 2024-01-10 at 9 07 24 AM


miketullo95:
testing this I ran into an error because I don't think the text block has a default text style variation. this doesn't happen with new text blocks but with already existing blocks. 
Screenshot 2024-01-10 at 10 09 51 AM

after getting passed that error everything else works as intended

codechefmarc (approved):
Tested these out and all looks good to me. I didn't run into the issue that Mike Tullo had, though - my default text (new block) had a variation of "Default" set.

PR #534, opened 2024-01-17
YALB-1687: Add content from template

dblanken-yale:
### Description of work
- Add the ability to add a new node based on available content templates
- Add support for zip files
- Add backup placeholder image in the event the image doesn't exist on the platform. (YAML import only)
- Create local menu for each content type
- Add 'blank' content type creation to templates drop down
- Reuse taxonomy if it was found
- Unrelated to content from template, but requested
  - Re-organize "Manage" content types inside of a "Manage content" submenu
  - Re-oragnize the "Manage menus" into their own submenu
  - Created landing page for "Add content" to template versions
  - Created landing page for "Menus"

### TODO
- [x] Verify which showcase items should be available in the templates -- subset determined
- [x] Verify that groups of pages should or shouldn't be included in a template -- determined 1 page per template
- [x] Replace fallback image of Dave's avatar with a more appropriate one -- replaced with handsome dan image
- [x] Create descriptions for all templates
- [x] Come up with any other text that should be displayed to a user upon creation
  - Currently set to: "Content generated successfully. Please make any edits now as this has already been created for you. Don't forget to change the URL alias."
- [ ] Find better location to house files -- currently at [content-templates](https://github.com/dblanken-yale/content-templates)
  - Note: not the end of the world to keep this for this version and update with the next

### Possible future enhancements
- [ ] A Lorem Ipsum sitewide checker (out of scope for this)
- [x] Create YAML manifest file in content-template repo so we don't have to hard code items inside of project
  - [x] Cache once retrieved
  - [x] Would deserialize into array resembling [current code](https://github.com/yalesites-org/yalesites-project/pull/534/files#diff-0b9e785ac71b6e46f3821c7fe6ff61d20304d7cfa80e94ae4cc5a28809cc4872R45-R144)
  - [ ] Could start the storage of these to later allow user contributions:

| id | content_type | title | description | yaml | file_location | platform_template |
| --- | --- | --- | --- | --- | --- | --- |
| 'blog1' | 'post' | 'Blog' | 'A template for a blog.' | NULL | 'https://....file.zip' | 1 |
| 'user1' | 'page' | 'Custom landing page' | 'Custom landing page for special events' | 'yaml data here' | NULL | 0 |

### Functional testing steps:
- [x] Visit `Content->Add Content`
- [x] Verify that "Create a page" is the default and that the default template for it is "Blank"
- [x] Click "Event", "Post", or "Profile" in the local menu above; ensure it takes you to the right place and that the right templates are loaded for them
- [x] Ensure that on submit, it takes you to the creation of a new node of that content type
- [x] Go back to the "\`Content->Add content` link
- [x] Select the type of content you'd like with a template that is NOT labelled `blank`
- [x] Ensure that it takes you to the edit node with data repopulated for edit
- [x] Make appropriate changes, Save, and then verify that content exists!
- Note: Cannot do this any longer due to us only using zip files.
  -  Delete an image that is used in one of the templates you imported already from media library if you know how to remove the file itself.  Keep in mind deleting from the media library might not delete the actual file, so it may come back.
  - Attempt to use that template again, verifying that a placeholder image (currently my avatar) is in its place
- [x] Validate that when creating the same template multiple times that the alias is properly updated to a unique alias, currently sequentially numbered (x-1, x-2, x-3)..
- [x] Import the zip file for a Page
- [x] Ensure new images are created, new tags are created if they didn't exist, and that no existing tags or images are duplicated in taxonomies or media gallery.
- [ ] Attempt to add an event (either Online or In-Person)
- [ ] Verify that it did not duplicate the event type after creation
- [ ] Add a hybrid event
- [ ] Verify that it reused both Online and In-Person event types

#### Menu testing
- [x] Ensure when mousing over "Content" in the menu you see:
  - [x] Each content type under "Add content"
  - [x] Each content type under "Manage content"
  - [x] You see "Manage Main Menu" and "Manage Utility Menu" under "Menus"
- [x] Ensure that in both "Add content" and "Manage content" the content types are sorted in the same way


codechefmarc (commented):
This is so great @dblanken-yale! Wonderful work here. In general, everything worked for me. A few notes:

1. I assume that this is still early in testing / approval as there isn't content for some of the items created like a blog post.
2. What is the best way to update existing / add new templates? I assume you'd have to add just the config file to this module as it reads it dynamically, it looks like? No other code to update?
3. I tried to test the fallback placeholder image, but it didn't work for me - I removed the tower image used as part of the banner image on the landing page template. Then, recreated that template and that tower image came back. I see that in the config for the single content sync, that image is referencing the `https://dev-ys-starterkit.pantheonsite.io/sites/default/files/starterkit/harkness-tower-dusk.jpg` image directly from the dev-ys-starterkit site. Is that correct? Or should these images come along for the ride and be inside that `config/templates` folder?
4. (I just read the comments in Jira about your issues on the multidev where I'm testing too. I can test locally, if you'd like. There are also ways to run an "install" command again on the multidev via Terminus if that helps, happy to help there)
5. I assume that your "headshot" is temporary and that they will provide a more official fallback?

dblanken-yale:
@codechefmarc Thanks for looking!

1. Oh yes, it is very much still a big work in progress.  Just wanted feedback on how I've started to implement to make sure I'm not going down the wrong path.
2. Yes, originally my idea was you'd only have to place a file with the proper naming convention to add a new template, however, in feedback already I'm seeing the need for metadata, which will make more modifications needed.  I might scrap the file placing idea all together and just have the YAML in a string to have it all together--jury still out there.
3. Yes, I noticed when you delete an image from media library, it doesn't actually delete the image itself, just the reference in media library--maybe for this specific case, where it's being used elsewhere in the site as to not create a broken image.  If you delete the image from the file system and then do the import, it should show my avatar (for now) as a fallback image.
4. Ah yes; I think I figured out the issue with the permissions--it was actually that I was using too broad of a stroke replacing 'uri' nodes in the YAML, so I was replacing things that weren't images.  LOL. Now that I am targeting those that are 'uri' with a value that has 'public://' it seems to be fine.  Thanks for offering!
5. Yes, it was just the image I had at the moment to place there.  LOL Hopefully we'll have a much better fallback image, like something that says "replace this image".  LOL

Any tips or ideas you can think of looking at the implementation that you either find hard to read or a bad Drupal practice?

codechefmarc:
To you point on #2, it may be beneficial to look at your work with the starterkit - https://github.com/yalesites-org/yalesites-starterkit - for example, maybe there is another directory in that repo that contains config/assets for templates. And then you can dynamically read from that repo so anytime you want to add a new template, it comes from there rather than the main YaleSites repo? Just thinking of ways to make it easier to update / manage these things is all.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Controller/TemplatedContentController.php
```diff
@@ -0,0 +1,38 @@
...
+    return new static($container->get('current_user'));
+  }
+
+  /**
+   * Builds the content selection form.
+   */
+  public function build() {
+    $formBuilder = $this->formBuilder();
+
+    return $formBuilder->getForm('Drupal\ys_templated_content\Form\TemplatedContentForm');
```
codechefmarc, line 35: Curious why the extra call to get a form - you could put a form in this controller itself - i.e. the way that `ys_alert` works?

codechefmarc:
And another thought and then I'll stop talking! :) You mentioned a yml file for overall config to read data. Could be still that way on the `starterkit` repo and then you could add fields in there like an image preview of the template?

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Controller/TemplatedContentController.php
dblanken-yale: @codechefmarc I was going off of documentation I read on Drupal.org.  Are you talking about on the config side of ys_alert where it specifies the _form in the routing?  I'd love to consolidate since the Controller isn't doing much at all.

dblanken-yale: I basically removed the controller and am directly using the form.  It seems to be fine and eliminated a file for us to manage.

dblanken-yale:
Hmmmm let me build atomic and CL in the event some things went into develop here that need their develop repos.

dblanken-yale:
@kara-franco Looks like it needed develop branches from atomic and CL.  Now it's looking good.  Feel free to visit now to double check, but here is a screenshot from when I went in just now.

Screenshot 2024-02-07 at 11 35 58 AM

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/config/templates/event__in_person.yml
```diff
@@ -0,0 +1,72 @@
...
+      base_fields:
+        name: yale-brick-in-wall-detail.jpg
+        created: '1691755627'
+        status: true
+        langcode: en
+      custom_fields:
+        field_media_image:
+          -
+            uri: 'public://2023-08/yale-brick-in-wall-detail.jpg'
+            url: 'https://yalesites-platform.lndo.site/sites/default/files/2023-08/yale-brick-in-wall-detail.jpg'
```
codechefmarc, line 61: Just noticing the URL's for these images for the templates are pointing to `.lndo.site` and not a relative URL or external. I know we tested these on the multidev and they worked so not sure if they need to be changed, but just making a note about it.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Managers/TemplateManager.php
```diff
@@ -0,0 +1,209 @@
...
+   * @param string $content_type
+   *   The content type.
+   * @param string $template_name
+   *   The template name.
+   *
+   * @return array
+   *   The template options.
+   */
+  public function getTemplateDescription($content_type, $template_name) {
+    return $this->templates[$content_type][$template_name]['description'] ?? "Hi";
```
codechefmarc, line 199: Heh. I like the fallback, but perhaps empty string or default message?

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Managers/TemplateManager.php
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  /**
+   * Get the templates.
+   */
+  public function getTemplates() {
```
codechefmarc, line 205: This is picky so feel free to ignore me. But this function and the `getCurrentTemplates` are similar enough - you could make the `$content_type` argument optional and then if one is passed, juts get the templates with the specified content type, or if none is passed, get all of them.

codechefmarc (approved):
This looks really good! I've got a couple of notes in here, but nothing that is blocking it. Obv you mentioned the fallback image which would need to be replaced (I dunno, I like your headshot there as fallback - gives credit to your work as an easter egg!). Keep up the great work!

miketullo95:
This is just an amazing feature, my only comments would really be to get a temp check on if we want to give the content type selection the same design treatment as footer/header/alert settings. See my mockup below 
tempalte

My other comment is, im not sure if this was kept for testing purposes, but can we remove empty as an option and just have template options available to the user?

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/config/templates/event__in_person.yml
dblanken-yale: In my testing, it attempts the uri first and then attempts the URL as the fallback.  Before import, I check for the existence of the uri and replace it with the fallback image if it doesn't in the uri field.  This was my way of trying to ensure that the url was never used, as you're right--when you import and it uses the localhost URL, it'll timeout.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Managers/TemplateManager.php
dblanken-yale: LOL I can't believe I left that one in there.  Thank you so much for looking at this!  Will fix immediately!

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Managers/TemplateManager.php
dblanken-yale: Ah not a bad idea.  I can try that.  :D

dblanken-yale:
@miketullo95 Yes, we can have images and the like--this was a first pass to make sure we have functionality in place.  But yes, we can add metadata to the arrays to reference images and the like for view.

For the empty, we can remove it but it was part of the spec when I was tasked with this.  But we can easily remove this so we can take a look at it.  I'll do that when I make @codechefmarc 's changes above.

dblanken-yale:
@codechefmarc @miketullo95 Take a look; made the changes Mark requested as well as removing the empty versions.

codechefmarc (approved):
Another pass - looks great! Thanks for the info about the uri/url.

dblanken-yale:
@nJim I'm going to see how easy/hard it would be to add an option for remote files as an option.  I was thinking of targeting GitHub to start.  We'll see how it goes.

dblanken-yale:
@nJim Take a look when you can.  I have local and remote examples now.

miketullo95:
Template Content Options in ticket - https://yaleits.atlassian.net/browse/YSP-364

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/ys_templated_content.routing.yml
```diff
@@ -0,0 +1,7 @@
+ys_templated_content.selection:
+  path: '/admin/node/add/from_template'
+  defaults:
+    _form: '\Drupal\ys_templated_content\Form\TemplatedContentForm'
+    _title: 'Create content from a template'
+  requirements:
+    _permission: 'create page content'
```
nJim, line 7: This works, but may not be true in the future. I can foresee a time where we have additional roles and permissions to create different types of contributors. For example, someone who can contribute to the blog but is not allowed to create new landing pages. With this in mind, I would either:

Pick a more generic permission, such as 'access content overview'
OR create a new permission for 'create content from template' and give this permission to the appropriate roles.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Importers/ZipFileImporter.php
```diff
@@ -0,0 +1,245 @@
...
+
+      return new RedirectResponse(
+      Url::fromRoute(
+      'entity.node.edit_form',
+      ['node' => $nid],
+      )->toString()
+      );
+    }
+    else {
+      drupal_set_message(t('An error occurred and the content could not be imported.'), 'error');
```
nJim, line 194: We should replace this with the Messenger service. Not sure when this function is going away, but I feel like it's been on the chopping block for a long time but never actually gets depricated.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Modifiers/TemplateModifier.php
```diff
@@ -0,0 +1,228 @@
...
+  /**
+   * Generate a unique alias using the current date/time.
+   *
+   * @param string $alias
+   *   The alias.
+   *
+   * @return string
+   *   The unique alias (original alias with date/time).
+   */
+  protected function generateUniqueAliasWithDate($alias) {
```
nJim, line 145: This method is not currently in use.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Modifiers/TemplateModifier.php
```diff
@@ -0,0 +1,228 @@
...
+  /**
+   * Generate a unique alias.
+   *
+   * @param string $alias
+   *   The alias.
+   *
+   * @return string
+   *   The unique alias.
+   */
+  protected function generateUniqueAlias($alias) {
```
nJim, line 182: Food for thought. You are recreating logic that Drupal should do for you. If you remove the `url` line from the template, then you can delete all of these alias methods as the pathauto module will just fill in the details for you.

Of course relying on manual edits is never a good long term strategy. Is it better to sanitize the template to remove the 'url' setting? I like the idea of removing all of these alias function if possible.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Modifiers/TemplateModifier.php
```diff
@@ -0,0 +1,228 @@
...
+    return $this->uuidService->generate();
+  }
+
+  /**
+   * Get the current unix timestamp.
+   *
+   * @return int
+   *   The current unix timestamp.
+   */
+  public function getUnixTimestamp() {
```
nJim, line 224: This method feels a little silly to have. I love small methods, but I can't see a case where simply using `time()` isn't just as good. To each their own though.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Modifiers/TemplateModifier.php
```diff
@@ -0,0 +1,228 @@
...
+use Drupal\Component\Uuid\UuidInterface;
+use Drupal\path_alias\AliasRepositoryInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Modifies a content import for a unique insertion.
+ */
+class TemplateModifier {
+
+  const PLACEHOLDER = 'public://templated-content-images/placeholder.png';
```
nJim, line 14: I like your approach. Let's get the design/UX team to provide a real Yale branded image.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Importers/ZipFileImporter.php
```diff
@@ -0,0 +1,245 @@
...
+  /**
+   * ZipFileImporter constructor.
+   *
+   * @param \Drupal\ys_templated_content\Managers\ImportManager $importManager
+   *   The import manager.
+   */
+  public function __construct($importManager) {
+    $this->importManager = $importManager;
+    $this->contentSyncHelper = $importManager->getContentSyncHelper();
+    $this->templateModifier = $importManager->getTemplateModifier();
```
nJim, line 46: This is not something I normally see where you import a service via another service. I see that these different importers are used within the ImportManager. I think most of the time I see this setup as a Factory rather than calling these classes with the `new` keyword.

Everything you have is working. There is no need to refactor. Just sharing what is top of mind.

nJim (changes requested):
Small changes requested. This is lot of code so there is a bunch of things we could debate. But I tested it through and through and everything is solid. 

Some of this may be improved by using Drupal/Symfony conventions. I could see the importers being turning into a custom plugin type. This would given them a factory for creating them and better organize how they are loaded.

The templates themselves could be packed as config entities if we are feeling ambitious. But this does not buy us anything new at this moment.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/ys_templated_content.routing.yml
dblanken-yale: Nice; ya I'll work on just creating a specific permission for this so we can be explicit about that functionality.  Thanks a lot for looking.  I'll work on these updates this week so we can be ready.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Importers/ZipFileImporter.php
dblanken-yale: Can do; I think I originally used it as there was some issue that I can't remember with it being inside of the batch job queue.   But I'll figure it out--sounds like we'd have to make the change anyway in the future, so might as well get it now.

Is it that way with a lot of those druapl_<function> methods or is this one special?  If it's most all of them at this point I'll try to steer clear of them all in the future.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Modifiers/TemplateModifier.php
dblanken-yale: That's for reminding me about that one--that was an alternative to the generation of an alias that I knew wouldn't be taken.  Now it goes sequentially as this one seemed like it could puzzle users more than the 1, 2, 3 aspect.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Modifiers/TemplateModifier.php
dblanken-yale: I don't mind swapping it.  HAHAH  I wish I could even tell you why it was split out now that I see it again.  hahah

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Modifiers/TemplateModifier.php
dblanken-yale: Good idea.  Ya I worried a lot about the idea that an image might not exist on the page, but we'd still want it to load something so that they knew they would need to change it.

web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Importers/ZipFileImporter.php
dblanken-yale: No, these are the kind of things I'd like to know.  Thanks for sharing it.  I really do want to make this where someone more experienced in Drupal and PHP would recognize the correct logical pattern, so I'll keep this on my list of things to maybe address after some of the other ones here and see where it can get us.

dblanken-yale:
@nJim Ok, now that you mentioned the Plugin idea, I love it.  I was pretty heads down just trying to get a functioning version, but I agree a plugin for these kind of importers might be a great way to allow it to be extended.  I'm going to work on this as well.  Great thought.

Think I really need to find some time to deep dive into Symfony to see what other ways we can make this even better and leverage it.

miketullo95:
@dblanken-yale this is looking so good! the menu is sooooo much cleaner... 
My notes going through some things:

- Super super small, but how come the page section of adding content doesnt have the " | from template" in the breadcrumbs? its next to "back to site".
Screenshot 2024-04-16 at 3 07 44 PM
Screenshot 2024-04-16 at 3 07 40 PM
- I noticed clicking "Menus" doesn't have a target. Do we need to make a view similar to the manage taxonomy view in the authoring experience? 
- Trying to add content from a template thats not blank gives me a site error
Screenshot 2024-04-16 at 3 14 08 PM
error message - "Error: Call to undefined method Drupal\ys_templated_content\Managers\TemplateManager::getTemplateTitle() in Drupal\ys_templated_content\Managers\ImportManager->createImport() (line 96 of /code/web/profiles/custom/yalesites_profile/modules/custom/ys_templated_content/src/Managers/ImportManager.php)."

dblanken-yale:
@miketullo95 Thanks for looking at it and my apologies for the error.

- I'm sure I have some reference to "From Template" somewhere.  Let me find where that is and I'll update it.
- Yes, I want to do similar to Taxonomy but was having some trouble getting it to do the same thing.  I'll look more into it to see what we can do.
- I removed that function accidentally thinking it wasn't being used LOL I'll re-add it hahaha

dblanken-yale:
@miketullo95 Ah the "From template" is deriving from the URL; here is the path to them:  `/admin/node/add/from_template/{content_type}`.

So, for instance, if I removed "from_template" that above would read `Add`.  I think it's not doing this in the event of a page because the default for that if you were to go to `/admin/node/add/from_template` is page.  But for the others, it thinks it's nested.

If we had to have it, is there a name you'd like to be there?  Keep in mind it'd be part of the URL generated for it.

dblanken-yale:
The bug error fix should be re-building now.  I removed a method I shouldn't have.  :D

dblanken-yale:
Worked on a way to have a landing page for both Menus (in ys_links) and Add content (ys_templated_content).  It's looking good so far.  Thanks @codechefmarc for directing me to that twig file.  I think that's all of the requested changes.  I also removed the default for page creation such that the "breadcrumb"-looking thing stays consistent for all content types, but this meant that the default link does not have a default value, which should be fine now that we have the landing page.

codechefmarc (dismissed):
This is looking really good! All worked great as far as I can see. I like the change to make the "Add content" page feel like a Drupal core page but it takes you to the template page when you click on a content type. Perfect. A few notes that I'm sure you are aware of (hence the request changes):

1. Placeholder image - I see Jim brought this one up too.
2. I tried deleting the images used, but no luck - it must be a really niche use case? Glad you build in the placeholder though! This one isn't a blocker, just a comment.
3. For page, I see "Remote Zip", "Local Zip", "Remote Zip Does Not Exist", and "Landing Page Does Not Exist" - I assume these are tests for these various scenarios? I assume these will be removed before this gets merged? Possible future enhancement would be to, when selected "Remote Zip" to be able to specify a remote Zip URL and have it pull that in from someplace else. 'Course this means verifying that it actually a correctly formatted Zip, etc!

Great work!

dblanken-yale:
Found an issue in my own testing that it's duplicating the taxonomy terms.  I'm going to try to modify this to search for the taxonomy term and if it can't find it, it'll add it.  Otherwise, it'll attempt to re-use it.

Otherwise we get multiple "Online" or "In-Person" entries per import.

dblanken-yale:
I think I addressed the change requests so going to dismiss them for now.

dblanken-yale:
@kara-franco @miketullo95 Check me to make sure the right Handsome Dan image is used by looking at that last commit adding him (can't test it in the PR since we're now only using zips), but provided we are ok with the descriptions and success text, I believe this to be ready for this release.  We can always change where the zips are housed next release.

miketullo95 (approved):
looks great! 😄 with the new copy this is ready to go through final testing

PR #535, opened 2024-01-17
rel-100-quicklink-format-fix: Re-introduce heading_html format

dblanken-yale:
## rel-100: Quicklink format fix

In updating develop from master, the allowed_formats were overwritten from before.  This re-establishes this for new quicklink blocks.

### Description of work
- Ensure that new Quicklinks blocks do not allow links

### Functional testing steps:
- [ ] Create a new quick link block
- [ ] Verify that the content section will not allow a link


kara-franco (approved):
Looks great, link option is now gone.

PR #537, opened 2024-01-18
rel-100: Linkit alias patch removal

dblanken-yale:
## rel-100: Linkit alias patch removal

Since I believe we are using the node id to set the home page for content on install, I don't think this patch was needed.

This is causing the home page alias to be used over the node id and somehow resulting in too many redirects in some cases.

### Description of work
- Removes the alias patch to allow node id to be used in Linkit

### Functional testing steps:
- [ ] Attempt to set the home page to a node
- [ ] On save, ensure that it still uses the node id
- [ ] Try Linkit in other cases, like inside CKE blocks to verify all is well.
- [ ] Verify no redirection error when logged out


dblanken-yale:
Due to being so close to release, and finding that the patch was not the reason for the multiple redirects occurring on the front page that eventually produces an error, we decided to let this one go until we can properly test more.  The fix for the original issue was to clear caches, then redirect errors don't happen.  So in the future if you change the front page and find that logged out users get redirect issues, clear the cache and all should be good.

PR #543, opened 2024-01-22
REL-100 Hotfix: add guard clauses to bypass <nolink>

dblanken-yale:
# REL-100 Hotfix: add guard clauses to bypass

There were some sites who had link references in their top level menus, resulting in errors attempting to get a valid node. From what I can tell, having a still results in a menu item showing as "routable", even though a node does not exist on it. This patches this by checking each variable exists and to check that a node param exists on the routable parameters. It should also guard against any other cases where it thinks it's routable but has no node.

### Description of work

- Adds guard checks to node detection

### Functional testing steps:

- [ ] Create a top level menu item with a link to
- [ ] Create a submenu item for that menu
- [ ] Verify that you can load the front page without errors
- [ ] Verify that the first submenu item has the same name as the top level item.


vinmassaro (approved):
LGTM

PR #544, opened 2024-01-23
fix: YALB(1963): Hotfix: Text Style Variation not set on existing blocks

dblanken-yale:
## Pull requests
- https://github.com/yalesites-org/yalesites-project/pull/544

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.install
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.install
@@ -77,23 +77,29 @@ function ys_core_update_9002() {
  * This update will set a default field_style_variation for text blocks.
  */
 function ys_core_update_9003() {
-  // Get all blocks that have field_style_variation that is empty.
-  $query = \Drupal::entityQuery('block_content')
-    ->accessCheck(FALSE)
-    ->condition('type', 'text')
-    ->allRevisions();
+  $block_storage = \Drupal::entityTypeManager()->getStorage('block_content');
+  $query = $block_storage->getQuery();
+  $query->accessCheck(FALSE)
+    ->condition('type', 'text');
 
   $ids = $query->execute();
-  $block_content_storage = \Drupal::entityTypeManager()->getStorage('block_content');
+
   foreach ($ids as $id) {
-    $block = $block_content_storage->load($id);
-    if ($block && $block->id() == $id) {
-      $style_variation = $block->get('field_style_variation')->getValue();
-      if (empty($style_variation)) {
-        $block->set('field_style_variation', 'default');
-        $block->save();
-      }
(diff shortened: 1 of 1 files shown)


dblanken-yale:
Doh, it overwrote my instructions.  haha

Basically, I thought a good test would be to attempt to bring down that URL in question in the Jira ticket and verify that before the update hook, text_style_variation is not set, and after running the hook update, it is set.

To run the update hook: `lando drush -d php-eval '\Drupal::moduleHandler()->loadInclude("ys_core", "install"); echo ys_core_update_9003();'`

dblanken-yale:
I added both @codechefmarc and @nJim to just double check that I am, indeed, targeting the latest revisions since `loadRevision` has been deprecated.  Using `$block_content_storage->load` seems to work in my tests, but I can't find [specific documentation](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityBase.php/function/EntityBase%3A%3Aload/10) that states it always brings back the latest revision?  Didn't know if you knew off the top of your head at all?  I assume it is since it should return the entity.

Also brings up that our first update hook is referencing `loadRevision` still.  Didn't know if we wanted to include that change in this as well or make it a separate ticket since it's not really needed for this hotfix.

codechefmarc (approved):
This worked for me. It took me a bit longer to test because of one node from dev (https://dev-yalesites-platform.pantheonsite.io/welcome) that has a text block right at the top under the two buttons was not changing! All of the rest of them were. So, I halfway suspect it has something to do with that field getting partially set and so your code isn't touching that one block. But, this one (https://dev-yalesites-platform.pantheonsite.io/testing-all-of-the-things) worked great - all of the text blocks got updated there no problem.

dblanken-yale:
@codechefmarc Thanks a lot for looking.  I'll check out that outlier to see what could be up.  I want to make sure we get them all.  So it still made you select the default?

dblanken-yale:
It makes me think it's not targeting correctly.  I attempted to look for that UUID and I am not seeing it at all. hmmm

dblanken-yale:
So strange.  It is set after the update; I found the block id (329) and was able to see it updated in `block_content__field_style_variation` referencing revision 620 in the db, which is referenced also in the `block_content` table.

If I manually set it in the page, it sticks.  Nothing changes, and the revision table is pointing to the same values!  So strange.

dblanken-yale:
@codechefmarc Mind trying it one more time?  I think I finally got it right.  It should take even the case you found.  Let me know if it works for you.  So far it has for me locally as well as the site that reported this.

codechefmarc (approved):
Yup, that did it! I pulled down this branch locally, then ran `nom run build-with-assets` which grabbed the version from `dev` that had the non-set text variation and then it ran the update hook and now it's fixed locally. Approved!

PR #547, opened 2024-01-24
YALB-1696: Hotfix: Can't configure Page title as Site admin

dblanken-yale:
### Description of work
- Add method to have a better attempt to get the block type
- Modified block form JS to revert to providing the error message to the submit button if the input targeted is not visible.

### Functional testing steps:
- [ ] Verify you can modify the page title of a page using the configure button
- [ ] Find or create the following items, ensuring that the functionality in [YALB-959](https://github.com/yalesites-org/yalesites-project/pull/515) is in tact:
  - [ ] Tabs
  - [ ] Quick Links
  - [ ] Image Grid
  - [ ] Media Gallery
- [ ] Attempt to add new elements outside of those to ensure no errors happen

On develop for now so we can test, then will move to master for hotfix inclusion.

[YALB-959]: https://yaleits.atlassian.net/browse/YALB-959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/js/block-form.js
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/js/block-form.js
@@ -1,59 +1,78 @@
 ((Drupal) => {
   Drupal.behaviors.ysCoreBlockForm = {
-    attach: () => { // eslint-disable-line
+    attach: () => {
+      // eslint-disable-line
 
       /*
-      * Function to validate that all media items contain media.
-      *
-      * @param object blockType
-      *   blockType object.
-      *
-      * @return string
-      *   The error message or empty string.
-      * */
+       * Function to validate that all media items contain media.
+       *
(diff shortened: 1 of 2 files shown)


nJim (approved):
do not see any red flags in this code. I walked through the logic and it is sound. It may have some problems within other block management tools, but we are not using them right now so it is okay. You called this part a little 'messy' and I think that's a good description of the type of tech debt. I would love to find a different vehicle for making these changes as this is a bit of code to maintain. Still a fine fix for now. I'm making as code review complete.

myvaughn (approved):
Functional testing looked fine. I didn't really review the code, but did notice a lot of formatting changes. Are all of the devs on this project using the same code formatting plug-in/settings?

PR #558, opened 2024-02-06
YALB-1568: FE Bug: Profile page breadcrumb issue

codechefmarc:
… content type styling

### Description of work
- Adds body class with content type for specific content type styling

### Functional testing steps:
- [x] Create two different content types with one of them being a profile
- [x] No need to place anything in layout builder for these tests
- [x] View these pages on the front-end and inspect the page
- [x] Verify that the body class has a class of `ys-content-type-[content-type]`.

Screenshot 2024-02-06 at 3 08 43 PM

- [x] Visit a non-node page such as the search page or an admin page
- [x] Inspect the page and verify that no specific content type body class has been added
- [x] Add the profile page to a menu as a third level deep item (this will ensure breadcrumbs will appear) - do this via `/admin/structure/menu/manage/main`
- [x] Visit the profile page and verify that the breadcrumbs have enough space between the profile meta block and the header.

Screenshot 2024-02-06 at 3 11 16 PM

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
@@ -457,3 +457,11 @@ function ys_core_get_block_type(&$form, &$form_state) {
 
   return $block_type_name;
 }
+
+/**
+ * Implements hook_preprocess_html().
+ */
+function ys_core_preprocess_html(&$variables) {
+  // Adds the content type in a class to support specific content type styling.
+  $variables['attributes']['class'][] = isset($variables['node_type']) ? "ys-content-type-{$variables['node_type']}" : NULL;
+}


dblanken-yale (approved):
Nice job.  I like the addition of this body class so we can better target differences.  Going to go look at the other repos for this, but this looks great and fixes the breadcrumbs so that they're visible.

dblanken-yale:
Just wanted to demystify this awesome code change; I thought there would be more in other repos regarding this fix, but the rest was there!  Very nice job, @codechefmarc!

[CSS with this class](https://github.com/yalesites-org/atomic/blame/62a7dfb72498f6f48f15c41e44645f61ce907c62/css/layout-builder.css#L79)

joetower (approved):
@codechefmarc Works wonderfully, again 😆 

Thank you! I'll remember what you did here. It looks straight-forward.

nJim (approved):

PR #559, opened 2024-02-07
YALB-1703: Bug: Updating site settings changes home page path from page node path to URL alias

dblanken-yale:
### Description of work
- Replaces front page input from a Linkit input to a textfield path with autocomplete input

### Functional testing steps:
- [x] [Log in](https://pr-559-yalesites-platform.pantheonsite.io/cas)
- [x] Visit the [site settings page](https://pr-559-yalesites-platform.pantheonsite.io/admin/yalesites/settings)
- [x] Start entering text that would match the name of a node
- [x] Verify that results do not show until you have typed at least 3 characters
- [x] Verify that results matching your input displays in the autocomplete section
- [ ] Verify that if you type an invalid input, it complains to you.
- [ ] Verify that if you have no input data, it still displays that it is required.

### To test against alias case (locally is best?)
- [ ] In the database using the editor of your choice, update the `system.site` `data` blob's inside of front to have a length of 9 (s:7 to s:9) and the value for it from `/node/4` to `/homepage`.
- [ ] Clear your cache on the Drupal side
- [ ] Visit the site settings page and ensure it correctly selected the `Homepage (4)`
- [ ] Save the page and go back into the database and ensure that blob data is back to `s:7` and `/node/4`


codechefmarc (approved):
Nice! This worked well! Approved.

myvaughn (approved):
Works as expected. The only thing I noticed is that the width of the changed input field doesn't match the width of the others on the form. But not a big deal as far as I'm concerned. 
image

dblanken-yale:
@myvaughn I'll create a ticket on that and address it next since it's really a CSS change in another repo.

nJim (changes requested):
I'm requesting changes to try and keep the work contained within the settings form. But I don't know if I'm actually keeping the code simpler. So feel free to push baack if you disagree as your approach is 100% valid when a custom solution is needed.

In short, we can keep the inout widget displayed in the form separate from the value stored in config. In the form:

```php
$form['site_page_front'] = [
      '#type' => 'entity_autocomplete',
      '#title' => $this->t('Front page'),
      '#description' => $this->t("..."),
      '#default_value' => /* get Node object from $siteConfig->get('page')['front'] */,
      '#required' => TRUE,
      '#target_type' => 'node',
    ];
```

Next we need to update the `validateForm` method as this is storing a node object and not a string.

Finally we update the `submitForm` method. Here we can convert the form value into what is expected in the configuration. Is it just this path?

```php
->set('page.front', '/node/' . $form_state->getValue('site_page_front'))
```

Let me know what you think.

dblanken-yale:
@nJim Let me see what I can do.  I was hesitant to modify data inside of the validate method; is that common practice in Drupal?  I'd think there'd be no mutation of data inside of a validate method.

But yes, I have code somewhere that was attempting to convert back and forth path to entity and back.  Let me see how it goes.

dblanken-yale:
@nJim Building now; took a crack at keeping it all in the `SiteSettingsForm.php`.  Also handled the slight case where any site would have an alias already set, so it could change it back to a node notation the next save.

The `gin-custom` work is just making the input the same width as the others like @myvaughn noticed.

nJim (approved):
Approved

PR #565, opened 2024-02-15
YSP-320: Admin secondary toolbar showing white foreground on white background in light mode

dblanken-yale:
## [YSP-320: Admin secondary toolbar showing white foreground on white background in light mode](https://yaleits.atlassian.net/browse/YSP-320)

We found that on light mode, the secondary toolbar was presenting itself as white on white, making it hard to navigate.  This removes that decoration so that it displays the same in both dark and light mode.

### Description of work
- Removed CSS that forces background color on secondary toolbar

### Functional testing steps:
- [ ] Ensure you are in light mode on your computer
- [ ] Log into the site
- [ ] Visit any page
- [ ] Ensure that the secondary toolbar is visible
- [ ] Change your computer to use dark mode
- [ ] Visit any page
- [ ] Ensure that the secondary toolbar is visible

#### Light Mode
LightMode

#### Dark Mode
DarkMode

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_toolbar/css/ys_toolbar.theme.css
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_toolbar/css/ys_toolbar.theme.css
@@ -74,22 +74,3 @@ behind the a.toolbar-icon element. */
 .gin--dark-mode :not(.glb-form-checkboxes):not(td):not(.media-library-item__click-to-select-checkbox):not(.field-content) > .glb-form-type--checkbox input:checked ~ .glb-checkbox-toggle .glb-checkbox-toggle__inner {
   background-color: var(--gin-color-primary) !important;
 }
-
-/* Make the secondary toolbar the same color as the primary toolbar */
-.gin-secondary-toolbar--frontend {
-  background: rgba(255, 255, 255, 1) !important;
-}
-
-.gin--dark-mode .gin-secondary-toolbar--frontend {
-  background: rgba(42, 42, 45, 1) !important;
-}
-
-@supports ((-webkit-backdrop-filter: blur()) or (backdrop-filter: blur())) {
-  .gin-secondary-toolbar--frontend {
-    background: rgba(255, 255, 255, 1) !important;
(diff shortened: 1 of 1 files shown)


joetower (changes requested):
@dblanken-yale The changes you removed were to set the toolbar to match the admin bar color and not have transparency at all in that secondary toolbar. 

I think the issue you're seeing in this ticket (white on white in toolbar) is the result of seeing the release branch of `atomic` - a mismatch of not having the changes from `develop` branches all in the same place. 

If you have `develop` checked out locally for all repos, you should not see this issue. 

In the video below, I have your branch here checked out, with the `gin-secondary-toolbar--frontend` code block removed (as in your PR here). 

**With this branch and `develop` checked out on `atomic`**

https://github.com/yalesites-org/yalesites-project/assets/366413/886c5dcd-490d-4119-8b71-fa22e69d7379


Note that as you scroll the toolbar is not only a slightly darker color than the admin bar, it has some transparency. Mike Tullo wanted the two toolbars to match and the secondary toolbar to _not_ have transparency, so it needed to be overridden for both light and dark themes. 

---
Now, here is what the admin theme looks like with the lines you removed in place, and `develop` on `atomic` - so really `develop` on yalesites-project.


https://github.com/yalesites-org/yalesites-project/assets/366413/085fb537-b5cd-455a-8d97-e06e3cabcc51

---

Basically, I don't think there is an issue. The issue is that we need the admin updates released. 

Please test locally and confirm what I'm seeing. I was confused seeing this issue because I remember it not being an issue with the admin theme overhaul. 👍

dblanken-yale:
@joetower Ok great.  I'll double check the Min/Max PR as well as I thought I had made branches off of develop that are up to date that helped build it, but maybe it didn't do it successfully.  Thanks a lot. I'll give it another try locally to verify that it's working.

dblanken-yale:
@joetower Yep, I see the missing commit now in develop of atomic.  Thank you so much for chiming in.  I'm rebuilding Min/Max now to verify but will close this as it looks like atomic is the missing piece here for the toolbar.

PR #567, opened 2024-02-16
YSP-142 Update media to use focal_point rather than image_widget_crop

laura-johnson:
### Description of work
- Disable image_widget_crop, enable and configure focal_point

### Functional testing steps:
- [ ] Edit an image in layout builder, eg. a media grid image or action banner image. You should see a preview that includes a focal point 
- [ ] Move the focal point and verify that the image is cropped accordingly. Note that it uses a scale and crop, so you need to use a media item that has different dimensions than the image style in order to see the focal point shift
- [ ] This applies to all of the image styles that used to have the crop. That is, the 1:1, 3.1, 3.2, 16.5 responsive image styles. If we want it for all media it will need more work - the 'focal point scale and crop' needs to be added for each image style.
- [ ] The existing media should look the same as it did before. The default focal point is the centre of the image (50,50) so hopefully nothing has changed.

Demo here: https://pr-567-yalesites-platform.pantheonsite.io/welcome

I added a profile listing block. The Alyssa DeChiaro profile image I used focal point on to focus on the left side (normally the center of the nave is in the center of the image).

focal-point-image

focal-point-test

### Notes:
- If you apply a focal point to a media item, that focal point applies to that image everywhere.
- I don't believe that there are any config conflicts caused by the removal of image_widget_crop and addition of focal_point, it seems like we can deploy the changes all at once


codechefmarc (approved):
I think this looks good! I tested out a bunch of the different image styles and all appeared to work just fine for me. I used a very tall image and set the focal point to different areas. I would like to get some Yale folks to test as well, so I'll tag @miketullo95 and @dblanken-yale for review as well.

miketullo95:
this is great! functions perfectly. I had a quick question regarding the preview function. could we only have one example of the previews for each ratio? I noticed there is a number after the ratio, i'm assuming the width resolution? or is there a use case to have them all, they all seem the same to me
Screenshot 2024-03-11 at 1 17 09 PM

sskalisky-4k:
@laura-johnson, I did some [initial testing](https://pr-567-yalesites-platform.pantheonsite.io/new-testing) with an image that is a good candidate for cropping woes. Despite setting the focal point to the top of the woman's head (screenshot below), she's excluded from the crop used on the Portrait Spotlight. 

YSP-142-focal-point-ignored

YSP-142-focal-point-ignored-in-portrait-spotlight

Also, we may have to manage expectations around the focal crop & Blocks like Grand Hero. A natural expectation would be that you can set a focal point to prevent the people in the photos from being excluded or decapitated at certain screen sizes. I assume the way we handle the responsive background image  prevents this from being an option.

laura-johnson:
@miketullo95 Good question about one preview for each ratio. I'll look into and gather thoughts on this. 

@sskalisky-4k I'll take a look at this... I only set focal point on image styles where we had cropping set before. It's not on all of them. It doesn't look like it's set on this one. Quite possibly we want it on all image styles.

sskalisky-4k:
@laura-johnson, ahh, that makes sense! The Portrait version of the Spotlight is pretty new, so I suspect a crop was never set for it, since we knew cropping was on the way out of the YaleSites party. I suspect the expectations for users it that the focal point will work wherever images are in use, so if there are other image style outliers we might want to add them to the mix.

laura-johnson:
@sskalisky-4k I switched the rest of the crop image styles to focal point and I think your example is looking better now.

miketullo95:
@laura-johnson could we also add some help text here? 
Screenshot 2024-03-14 at 9 21 50 AM


@sskalisky-4k thinking something like this; what do you think?

"(heading): Image Focal Point
(paragraph): Choose the area of your image that you want to keep in focus for all image ratios supported by the YaleSites platform. Once the plus icon is positioned on your desired location, all blocks will try to keep that part of the image within the image area."

sskalisky-4k:
Agreed, @laura-johnson, I think it looks much better now! Thanks!

@miketullo95 here's a slight rewrite: "Select the area of your image to keep in focus for all image ratios supported by the YaleSites platform. After positioning the plus icon on your desired focal point, Blocks will adjust to display that part within the image area."

miketullo95:
> Agreed, @laura-johnson, I think it looks much better now! Thanks!
> 
> @miketullo95 here's a slight rewrite: "Select the area of your image to keep in focus for all image ratios supported by the YaleSites platform. After positioning the plus icon on your desired focal point, Blocks will adjust to display that part within the image area."

perfect as always! lets go with that copy :)

laura-johnson:
Hi @miketullo95 and @sskalisky-4k ,

A couple of updates: 

1. I added the patch to limit image styles and I think it works really well. Give it a try!
2. I added the help text to the image, but there is a lot of stuff in the image widget that appears before the help text. It's not going to be that simple to get that text to appear within the widget. I could hard-code it in the template there, but that is not ideal. I believe I can use preprocess to grab that text and insert it into the template, but it will take some more work, and I wanted to get a sense from you if you think it might be okay appearing where it does now, or if I should go ahead and move it into the space right underneath the preview. After chatting with Jim in standup, we are thinking that perhaps this PR could go as-is, and we'll address the help text issue in a different ticket - we might consider contributing back to focal_point with an improvement for their widget.

focal-point-widget-help-text

sskalisky-4k:
Hi @laura-johnson! Should I use another environment to see the patch in action? When I revisited my test content to add a new image and brought up the Preview window, I still received multiple iterations for each of the images. For example, there are about 14 options for 16:5. Let me know if a video would be useful, happy to record a Loom!

laura-johnson:
@sskalisky-4k I checked and you were right, the styles were all appearing. I think that was a bit of multidev weirdness. I redeployed and now the code/config is there and it looks correct. Please check again when you have a chance!

sskalisky-4k:
I'm happy to report everything looks amazing on my end now! This looks great, @laura-johnson!

PR #585, opened 2024-02-27
YSP-291: Create Editorial Workflow

codechefmarc:
### Description of work
- Adds two new roles to the site: Author and Editor
- Adds content moderation workflow for pages, posts, events, and profiles
- Moderation has three states: Draft (default for newly created content and new revisions), Published, and Archived (Not published)
- Adds contrib modules: Moderation Sidebar, Gin Moderation Sidebar, and Workflow Buttons
- Removes Publish Content module ( to be removed from Composer in a separate PR)

### Functional testing steps:
- [x] Login to the site as both an Author and Site Administrator roles (Since this is not yet part of CAS, login as your CAS user first, then login as an admin or platform admin in a separate browser instance and edit the test user to add the new role).
- [x] Create a new page, post, event, or profile.
- [x] Verify that on the edit page, there is a new button labeled "Create New Draft" at the top

Create Post _ YaleSites · 5 45pm · 02-26

- [x] Fill out the edit form, then click "Create New Draft"
- [x] Add any layout builder blocks you wish, then save the layout.
- [x] Verify that on the top right side in the toolbar, there is a new button for the moderation state that should say "Tasks DRAFT" in red.

My First Post _ YaleSites

- [x] Copy the URL and then visit the page in a browser that is not logged in
- [x] Verify that the page cannot be viewed as an anonymous user as it is still in a "draft" state
- [x] Login to the site as both an Editor and Site Administrator roles
- [x] Click on the "Content" menu link in the toolbar and verify there is a new tab at the top called "Moderated content"

Moderated content _ YaleSites

- [x] Click on "Moderated content" and verify that there the previously created content is in a draft state.
- [x] Click on the title of the draft content
- [x] You can change the state in one of two ways: One is at the top of the page there is a new "Moderation state" block that has a dropdown to change to a new state (for example "Published"). The second way is via the toolbar menu item on the far right that says "Tasks DRAFT" in red.

Screenshot 2024-02-26 at 5 51 27 PM

- [x] Change the content to "published" status via either method.
- [x] Verify in a logged out browser that the page is now accessible to anonymous users.
- [x] Authors or Editors can create drafts from existing published pages as well. To do this, simply edit the page in either layout builder or in the "Manage settings" node edit page.
- [x] Once the page is saved in layout builder, there will be a draft version available and logged out users will not see the new change until it is published.
- [x] Verify that the original content is still visible and the new content is not in a logged out browser
- [x] As an editor role, visit the page. Use the Moderation Sidebar via the top menu to select "View existing draft"
- [x] Then, while viewing the draft, you can make additional edits by clicking "Edit draft" in the Moderation sidebar, or publish changes by clicking on "Publish". There is also the ability for an author or editor to discard the draft. Only Editors can publish the changes.

Screenshot 2024-02-26 at 5 59 05 PM

Screenshot of editing roles for a user:

Screenshot 2024-02-26 at 6 11 41 PM


dblanken-yale (approved):
Very nice work!  Pretty cool to see revisions in action here.  I did want to make sure that being a combination site admin and one of the new roles won't present any permission issues, right?

Was it acceptable to let authors modify taxonomy, etc?  I saw there is no drafts for that sort of thing.  I'd think it'd be a pretty hard thing to implement as well.  Just thinking of a malicious author I guess.  ;)

Screenshot 2024-02-27 at 8 54 20 AM

But as far as the scope of this work, it looks great.  Now to get some styling done on the sidebar.  :D

miketullo95:
@codechefmarc @dblanken-yale 
Sorry for the super delay here, just getting to this PR. Im trying to add a user with one of the new roles and I don't seem to see them as an option. Apologies if im looking in the wrong area. 
Screenshot 2024-03-14 at 4 44 33 PM

dblanken-yale:
@miketullo95 I elevated you to Site Admin and Author; I think when you log in it resets to Platform Admin for us so hopefully you are still logged in.  If you find you can't see it again, let us know and we can find a time to be there with you so we can revert you back to that combination.  Or you can make a local account and I can elevate it for you, which wouldn't reset it.

codechefmarc:
@miketullo95 - Good find! What roles are allowed to set these new roles? I.e. if you are a site admin role, can you set other people as editors/authors or should that only be the job of the platform admin? I can change the permissions to allow this and push up those changes. Thanks!

miketullo95:
@codechefmarc my thinking is that site admins (ideally on the platform, those are the "site owners") are the ones promoting or demoting users roles. So I think that is a good change to have site admins set people as authors or editors

codechefmarc:
@miketullo95 - I just made the change on the multi dev so now site admins can set the editor and author roles for other users.

codechefmarc:
@miketullo95, @kara-franco, @dblanken-yale  - I've updated the site to now have the Contributor role take the place of the Author role. And I set the permissions of the Contributor and Editor to the same as the Site Admin with the exception that Contributors can only deal with drafts and can't publish items, where Editors can. Site admins are the only ones able to set the roles of other users to editor and contributor. Ready for a re-test, thanks!

dblanken-yale:
Are there any repercussions to disallow more for contributors.  I'm just thinking again of someone malicious that couldn't delete a published node, but could delete the media for all nodes, or update them to something inappropriate, or delete all documents that might be linked on the site.

What if they were only allowed to create new media, but not edit/delete?  Would that ensure those cases wouldn't happen or would that cause more headaches for other areas?  Because the draft is only in the context of nodes, correct?  It would be up to the editor/site admin to clean up any unused or accidental media uploads in this case.

miketullo95 (approved):

miketullo95:
> Even deleting/edit own media could be problematic unless we removed it from their possession after a publish from the editor.

I was thinking about not having contributors not being allowed to delete content too... what do you think @kara-franco, is that too far? or maybe we could get feedback if people want that

kara-franco (approved):
This is AWESOME!! I tested the different roles and validated what we talked about Tuesday. 

I'm so excited to get this out to users!

kara-franco:
re: deleting- i could see it being really frustrating, as a Contributor, not being able to delete blocks or pages that I'm working on. 

I think deleting is part of contributing and keeping a tidy work flow. It would be a pain if a Contributor made a playground page to try out some content, and then needs to contact the Editor to delete the playground page. 

I understand the worry tho, that they can go on a delete crusade.

dblanken-yale:
@kara-franco I could be wrong, but I think they'll be able to do what they want with content on a node.  Like removing blocks, their own revision, etc., since that's part of the workflow at the node level.  What seems to not be under workflow is the addition of assets supporting these.  That's the scenario I'm concerned with.

Scenario: contributor makes a new page that is approved that has an image on it.  Down the line, another contributor chooses to delete that image or change that image, not realizing it was used on that previous node--because there is no workflow on the image itself, the change is made immediately, now displaying on a page that was not under publishing review that is published.  I don't know that it'd hinder them in the sense that they couldn't make their end resulting page, but it would require adding a new image version if they made a mistake of uploading an image that wasn't to be used.

Of course, the other side of this is there is no way for editors to see unused media in order to make a good decision that it can be taken down, correct?  So I could see that backfiring as well.  Seeing unused assets might be a longer term win?

I do agree it does limit--just don't want users getting frustrated because content not under draft could be changed.  We could also say this is good enough for now and see what the feedback is?  Are image changes tracked under revision?  Then it might not be as big of a deal?

dblanken-yale:
Follow up to my own question, it doesn't look like media item update changes aren't tracked with revision history like nodes are that I could easily see.

dblanken-yale (approved):
Even with the discussion of limiting permission at hand, what was requested of this is wonderfully implemented and does what we set out to do.  If it ends up that we do want to limit permissions for contributor roles, this sets up an easy path to do so given there are permissions for media deletion/edit.

Nice job!

PR #598, opened 2024-03-08
YSP-414: Deleting used content.school_logo media results in site wide page errors

codechefmarc:
### Description of work
- Fixes site wide error if any footer logo media was deleted in media library before being removed from the footer config.

Note: What this code does is actually delete out the missing config ID of the deleted footer media if any was set. Therefore, there is an admin message that gets output at the top of the page that tells the content editor that:

"Note, one of the footer logo images has been deleted. You may set a new one in the footer settings form" with a link to the form.

### Functional testing steps:
- [x] Login to the site and visit the footer settings form here: `/admin/yalesites/footer`
- [x] Switch the footer variation to "Mega"
- [x] Add some images from the media library (or uploaded) to the footer logos. Also add an image into the school logo.
- [x] Save the form.
- [x] Visit the front-end and verify that all logos show appropriately in the footer.
- [x] Visit the media library here: `/admin/content/media-grid`
- [x] Delete one of the images that was used for the footer logos.
- [x] Clear the cache - by default the site will keep the footer logo in place until cache is cleared.
- [x] Visit the front-end and refresh. Verify that the site does not crash and there is an admin message on the top informing that a logo has been deleted.

Screenshot 2024-03-08 at 12 23 07 PM

- [x] Also verify that the other logos load just fine.
- [x] Visit the footer settings form again, add an image from the media library (or uploaded) to the one that was deleted previously.
- [x] Save the form
- [x] Verify on the front-end that the logos all load.

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesFooterBlock.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Plugin/Block/YaleSitesFooterBlock.php
@@ -5,7 +5,9 @@ namespace Drupal\ys_core\Plugin\Block;
 use Drupal\Core\Block\BlockBase;
 use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\Entity\EntityTypeManager;
+use Drupal\Core\Messenger\Messenger;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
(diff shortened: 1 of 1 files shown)


dblanken-yale (approved):
Nicely done; I like the method passing in the setting name so it's explicit which setting we're targeting.  Nice work!

kara-franco (approved):
Awesome! Worked as expected.

PR #599, opened 2024-03-09
YSP-388: Localist migration

codechefmarc:
### Description of work
- Adds a new module `ys_localist` that handles migrations (importing) of events, places, event types, experiences, groups, and status from Localist.
- The sync is off by default, and when turned on, also needs to have a group set to import events from that specific group.
- Settings form to configure this module.
- Manual sync and hourly cron-based sync.
- Events can still be used without importing from Localist
- New fields were added to support all of the data coming from Localist
- New layout for Events edit page to account for all of the new fields
- Event Meta block template (the one at the top of Layout Builder for events) modified to showcase new fields (to be styled in a separate PR)

## Testing Migrations Locally
- [ ] Checkout this branch locally and run `npm run build-with-install` to get a fresh install.
- [ ] Login to the site as an admin and go to `/admin/yalesites/localist` (Also available under the "Settings" menu)

Localist Settings _ Your Site Title

- [ ] Check the box "Enable Localist sync"
- [ ] Click "Save"
- [ ] Once the form as reloaded, click "Save" again to create the groups (this only needs to be done once)
- [ ] Verify that there is now a group autocomplete box in the form
- [ ] Start typing the name of a group. A good one to use for testing at the moment is "Yale Health"
- [ ] Save the form
- [ ] Open a new tab with the content overview page `/admin/content`
- [ ] Verify that there are no new events in the list. Note: The "Revitalize past events" and "Promote your Events" were created already from the starter kit.
- [ ] Go back to the Localist settings form and click "Sync now"
- [ ] Note: The first sync takes a little bit of time (approx 30 seconds)
- [ ] Verify that a message on the top of the form says that it synchronized X events

Screenshot 2024-04-04 at 5 19 51 PM

- [ ] Return to the content overview page and refresh
- [ ] Verify that there are some new events listed

Screenshot 2024-04-04 at 5 20 02 PM

- [ ] Visit one of the events and verify that it loads the detail page with data from Localist.

Screenshot 2024-04-04 at 5 20 15 PM

- [ ] Note: This PR does not include styling

## Testing Migrations on Multidev (Requires a Localist admin login)
- [ ] Visit https://pr-599-yalesites-platform.pantheonsite.io/ and login as an admin
- [ ] Also open the Localist admin in a separate tab
- [ ] The multidev is currently set to sync events from the Yale Health group
- [ ] On Localist, edit an existing event that is part of the Yale Health group or create a new event and add it to the Yale Health group.
- [ ] Make sure the event dates are in the future and that the event is published.
- [ ] Save the event on Localist.
- [ ] Return to Drupal, and visit the Localist settings page here: `/admin/yalesites/localist`
- [ ] Click "Sync now"
- [ ] Verify that a message on the top of the form says that it synchronized X events
- [ ] Wait for the sync to finish and then visit the content overview page `/admin/content`
- [ ] Visit one of the events that you created/edited in Localist and verify that the data comes across into Drupal.
- [ ] Note: Sometimes it does take a second sync as the data from Localist into the API can take a few minutes to update.
- [ ] Cron will also run every hour and will refresh the data in Drupal with new data from Localist. Testing this basically just takes time. Also note that we're on a multidev here, so if the multidev spins down due to inactivity, cron may not run on the regular hourly interval. If you are platform admin, you can run cron manually from the Yale icon in the top left corner in the Drupal site.

Screenshot 2024-04-04 at 5 30 11 PM

## Testing Event Edit Form
- [ ] Edit one of the events on Drupal (Manage settings link). It doesn't matter if it was one created from Localist or not.
- [ ] Verify the new layout of fields in the edit form.
- [ ] Verify that the tabs, field groupings, fields, and field descriptions make sense.

Screenshot 2024-04-04 at 5 31 20 PM


dblanken-yale (approved):
Wow this is nice.  Great job!  Definitely makes me want to learn more about migrate!

I did have a future enhancement possibly--instead of hitting save again to populate groups, I wonder if groups haven't been populated we have a button in its place that tells the user to "Sync groups", even though it really saves again?  Saving twice might be confusing to someone that didn't know any better.  Once the sync happens, then we would show the groups input like we do now.  Just a thought, but this is great!

codechefmarc:
> Wow this is nice. Great job! Definitely makes me want to learn more about migrate!
> 
> I did have a future enhancement possibly--instead of hitting save again to populate groups, I wonder if groups haven't been populated we have a button in its place that tells the user to "Sync groups", even though it really saves again? Saving twice might be confusing to someone that didn't know any better. Once the sync happens, then we would show the groups input like we do now. Just a thought, but this is great!

I like this idea. I had originally planned a button in the same spot as the message informing about groups needing to be created. I can add that! :)

dblanken-yale:
@codechefmarc Just re-checked this and the button make this so much better.  Nice addition!  I think this one is good to go.

PR #604, opened 2024-03-14
YSP-405: Review design options for left and right columns -- DO NOT MERGE

dblanken-yale:
This is a multidev for Em and Kara to test some things.


miketullo95:
@dblanken-yale same thing here, we can probably delete this. We have a lot of section work coming for IYY

dblanken-yale:
A bit old and we believe research has been done for future action.

PR #605, opened 2024-03-14
YSP-302 Hide empty exposed filters

laura-johnson:
### Description of work
- Hides exposed filters if there are no options.

### Functional testing steps:
- [x] Add the following types of blocks: Post list, Calendar list, Profile directory
- [x] Set it up so that there are no filter options (remove taxonomies from posts, eg)
- [x] Verify the exposed filters do not display if there are no options.

### Notes:
I originally wanted to add a class to the form if there are no options and then hide it using CSS but it's Jim's preference that the component library not be involved on this one. So, in-line it is.

I also tried unsetting the options and submit button but that leaves a gap visually because the form is still there, adding its height.

### DIFF
--- a/patches/selective_better_exposed_filters/selective_better_exposed_filters-3432551.patch
+++ b/patches/selective_better_exposed_filters/selective_better_exposed_filters-3432551.patch
@@ -1,0 +1,39 @@
+From 30bf569df2f6b17d5ea64da83bf3372165bfa98a Mon Sep 17 00:00:00 2001
+From: Marc Berger <marc@fourkitchens.com>
+Date: Wed, 20 Mar 2024 18:41:32 -0700
+Subject: [PATCH] fix(3432551): If all exposed form elements are set to hide,
+ also hide the form.
+
+---
+ .../filter/SelectiveFilterBase.php               | 16 ++++++++++++++++
+ 1 file changed, 16 insertions(+)
+
+diff --git a/src/Plugin/better_exposed_filters/filter/SelectiveFilterBase.php b/src/Plugin/better_exposed_filters/filter/SelectiveFilterBase.php
+index 9d064a2..c4440b9 100644
+--- a/src/Plugin/better_exposed_filters/filter/SelectiveFilterBase.php
++++ b/src/Plugin/better_exposed_filters/filter/SelectiveFilterBase.php
+@@ -269,6 +269,22 @@ abstract class SelectiveFilterBase {
+               )
+             ) {
+               $element['#access'] = FALSE;
++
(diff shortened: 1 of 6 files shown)


codechefmarc (dismissed):
Hi Laura, I looked into this one and it looks like the selective better exposed filters has an update from the dev version we're currently using to a stable release 3.0.2 (https://www.drupal.org/project/selective_better_exposed_filters). I removed your code, updated this module, then changed the setting in the view for these guys under the better exposed filter settings:

Screenshot 2024-03-15 at 3 58 06 PM

And that appeared to work. If you want to give that a go for those 3 views, see if that works.

codechefmarc:
@laura-johnson - How strange! When I tested that locally, I wasn't getting the apply button, or at least I didn't think I did? I'll double check it later and see.

codechefmarc:
@laura-johnson - This was bothering me 😆, so I did some research and added an MR to the issue queue for that module. Apparently, there was another issue that was reported last year but the maintainer couldn't reproduce and closed the issue. So, I created a new one: https://www.drupal.org/project/selective_better_exposed_filters/issues/3432551

laura-johnson:
I add the patch to the /patches directory and applied it, works perfectly.

dblanken-yale:
Just making sure I'm testing the correct functionality.  Is it to test only if all taxonomy is removed or is it to test if there are no taxonomies assigned to a content type?  For instance, not providing an affiliation to any profile, but still having affiliations to choose from?  Or removing all affiliations?

I was just now testing removing all tags from nodes, and am only seeing profile directory display with the affiliation set to only `- Any -`.  Is that expected?

https://pr-605-yalesites-platform.pantheonsite.io/daves-ysp-302-test

laura-johnson:
@dblanken-yale It should be hiding it in that case. I'll take a look.

laura-johnson:
@dblanken-yale I was missing the config that goes with the new patch! Rookie move. It's fixed now. I think you mentioned you tried to add the other two types of block and they didn't appear... there were no published posts or events on that multidev. I published the posts and events, so you will see the blocks now. Maybe I should add a message if there are no results to prevent future confusion...

miketullo95 (approved):

dblanken-yale:
@laura-johnson haha Thanks for double checking me; I did forget to check that they were published, just deleted all that had tags as it was easier than disassociating at the time.  LOL

Nice work!

laura-johnson:
@dblanken-yale Is this approved then?

dblanken-yale:
Sorry for the late response; yes I believe it passes for me.  :D

nJim (approved):
Approved.

PR #607, opened 2024-03-20
YSP-425: Remove link from Reusable Blocks view

dblanken-yale:
## [YSP-425: Remove link from Reusable Blocks view](https://yaleits.atlassian.net/browse/YSP-425)

Remove the export drop down in operations based on whether they have access to export content.  Not just from the Reusable block view, but any view that has operations.

Currently Single Content Sync are ensuring they have access to the node to display this option, but we don't want to just base it on this, but also whether they have the permission access to export content (which only admins have at the moment).  This ensures that we can intercept this before it gets to the user and remove the option to export.

### Description of work
- Adds a hook to remove the export operation if they do not have access to export content.

### Functional testing steps:
- [ ] [Log in](https://pr-607-yalesites-platform.pantheonsite.io/cas)
- [ ] Visit the [reusable blocks view](https://pr-607-yalesites-platform.pantheonsite.io/admin/content/block?menu=main)
- [ ] Verify that the drop downs do not have "Export"

Screenshot 2024-03-20 at 2 07 20 PM

#### As an admin
As an admin you should see it--if you can log in as user 1, give it a try:
Screenshot 2024-03-20 at 2 04 42 PM

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
@@ -471,6 +471,18 @@ function ys_core_get_block_type(&$form, &$form_state) {
 }
 
 /**
+ * Implements hook_preprocess_links__dropbutton__operations().
+ */
+function ys_core_preprocess_links__dropbutton__operations(&$variables) {
+  // Remove export link if they don't have permission to export single content.
+  if (isset($variables['links']['export'])) {
+    if (!\Drupal::currentUser()->hasPermission('export single content')) {
+      unset($variables['links']['export']);
+    }
+  }
+}
+
+/**
  * Implements hook_preprocess_html().
  */
 function ys_core_preprocess_html(&$variables) {


codechefmarc (changes requested):
I'm not sure this is necessary? I checked the permissions here: `admin/people/permissions/module/single_content_sync` and there is a permission setting for "Content block: Export entity". I tested this and as a site admin, not platform admin, I didn't see the export unless the permission was set:

Screenshot 2024-03-20 at 8 05 24 PM

Screenshot 2024-03-20 at 8 07 41 PM


If I'm missing something, let me know, thanks!

dblanken-yale:
@codechefmarc That’s what was puzzling me as well.  While the permission shows there, no one has the permission set, yet it still shows in the reusable blocks view.  It doesn’t show in the Content view, and I didn’t see a way to pick and choose operations like the bulk actions field let you do.  Am I missing something easy?  I’d love to not have to have this code there if permissions could do the trick.

dblanken-yale:
What if you are a platform admin?  Do you see it then?  Because given the permissions I’d think you shouldn’t.

dblanken-yale:
It looks to me that they're checking access here, but that the access check isn't really looking at the permission we see there:

https://git.drupalcode.org/project/single_content_sync/-/blob/1.4.x/single_content_sync.module?ref_type=heads#L134-145

https://git.drupalcode.org/project/single_content_sync/-/blob/1.4.x/src/ContentSyncHelper.php?ref_type=heads#L228-236

dblanken-yale:
Aha; it's because we have "Administer block content".  Ok maybe this isn't needed if only we are even Platform admins.  So while it is forbidden on Export check, it passes because of that permission.  I didn't know it would check up the chain--pretty cool!  I'll ask in our YS standup today to see if it's needed then.

dblanken-yale:
@codechefmarc Going to close this.  Talked to the group and they agree it's probably a non-issue since it's only platform admins.  Thanks for bringing this up--I felt a little bad having this specific code in there, but it did make it work.

PR #612, opened 2024-03-25
YSP-418: Fix double-encoded document links

laura-johnson:
### Description of work
- Links are double-encoded which breaks the link if there is a space in the file name.

### Functional testing steps:
- [ ] Add a file that contains a space (or more than one) in the file name
- [ ] Add a link to that file in a linkit field, like Spotlight - landscape
- [ ] Verify that the link address does not contain %2025 where the space was and that when you click on it, the document opens

### Notes:
There is an issue for this in the Linkit queue: https://www.drupal.org/project/linkit/issues/3348154. There was a patch added to Linkit 6.0.x and they closed the issue. The patch does not apply to 6.1.x. Also, the patch seems to rely on using the Linkit field formatter, which we're not, we're using a core link formatter: Separate link text and URL. I suggest in the queue that the widget should take care of urldecoding because the core link formatters expect a decoded URL and it would be great if we can use the Linkit widget with any of them. I provided a simple patch that I've added to Linkit here, which decodes the uri value in massageFormValues() in the widget.


codechefmarc (approved):
Tested on the multidev and compared to dev and works great! Approved!

vinmassaro:
@laura-johnson The patch looks good but may not get seen by the maintainer since it’s posted to a closed issue. Can you open a new bug issue in the Linkit queue for it? Thanks!

laura-johnson:
Thanks @vinmassaro you're right, I opened a new issue [here](https://www.drupal.org/project/linkit/issues/3436733). I updated the patch in our composer.json to the one from the active issue (named with the new issue number, same patch).

dblanken-yale (approved):
This is working great.  Tested with /media reference, pasted URL with encodings already entered, and URL without encodings (just spaces) and all work as intended.

Nice job!  Let's hope they accept that patch!

miketullo95:
heard this PR wanted a UX look, am I interpreting this correct? 

I uploaded a png named "testing this link.png" and when I uploaded and linked to that image via a button here is the URL I got "https://pr-612-yalesites-platform.pantheonsite.io/sites/default/files/2024-03/testing%20this%20link.png". It **does** have %20 for spaces but **does not** %2025, is this intended? if so, it passes for me :)

laura-johnson:
@miketullo95 Yes normal URL encoding substitutes %20 for a space. It was double-encoding that was breaking the URLs.

PR #624, opened 2024-04-05
Semantic release backmerge updates

vinmassaro:
### Description of work
- Semantic Release Backmerge was added in v1.2.0 to keep develop up to date with master after a release, but it failed. This PR switches the merge strategy to `merge` instead of the default `rebase`.
- `develop` is also a protected branch so I'm setting actions/checkout step [per the docs](https://github.com/saitho/semantic-release-backmerge?tab=readme-ov-file#backmerging-into-protected-branches)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


dblanken-yale (approved):
Looks sane to me.  :D

PR #627, opened 2024-04-10
YSP-419: Rebrand Twitter to X

dblanken-yale:
## [YSP-419: Rebrand Twitter to X](https://yaleits.atlassian.net/browse/YSP-419)

Changes also in [Component Library](https://github.com/yalesites-org/component-library-twig/pull/353)

### Description of work
- Change embed option to use X text/image
- Update footer to use X-twitter option
- Update existing media to use new png
- Modify each embed's image color to be black like the new X image
- Copy all images to the location expected for retrieval

### Functional testing steps:
- [x] Add an X embed via block addition or in the [media library](https://pr-627-yalesites-platform.pantheonsite.io/media/add/embed)
- [x] Verify when attempting to add an embed that `Learn about supported formats and options` talks about X over Twitter.
- [x] Verify that X is the image that shows for the added item
- [x] Go to [Footer Settings](https://pr-627-yalesites-platform.pantheonsite.io/admin/yalesites/footer)->Social Links and verify that `X (formally Twitter) URL` is displayed
- [ ] Verify that the visually-hidden span on the social media icon when visiting a page is `X (formally Twitter)`
- [ ] Add another embed, ensuring that the image describing it is now a dark black color over the tanish color previously present.  (Note, due to caching, you'll see the old one until a new instance is added)

#### Locally
- [x] Bring down develop and setup
- [x] Add one or more Twitter embeds
- [x] Change to this branch
- [x] `npm run confim`, allowing it to also do the embed update
- [x] Verify afterward that existing twitter embed media is now displaying with the X logo


codechefmarc (approved):
This all looks good to me! I was curious about leaving in the old config (`twitter`) in the `ys_core.social_links.yml` file as that is only for new installs to get default values, so you could probably remove the reference there and it shouldn't be a problem.

As to naming things, a lot of tech blogs started using the phrase "X (formally Twitter)" to help with confusion, so we could do something similar here?

Minor changes, so I've approved for now as all of it worked well for me on both multidev and locally.

dblanken-yale:
@codechefmarc Ah good call.  My worry there was that it'd remove the config and I wanted to make sure we copied the value; good to know it wouldn't affect currently set values.

Ya I'm trying to remember what was ultimately decided for the reference text we want to use.  I like them all; feels like since we're making the transition it should be all or nothing, but I am pretty stubborn when it comes to that, so don't 100% trust myself on that.  haha

Thanks for the feedback!  I'll remove that twitter reference in the config.  Glad it looks ok otherwise.

dblanken-yale:
Incorporated the "X (formally Twitter)" into this.  Think it's at its final stages.

dblanken-yale:
@miketullo95 I've updated the images to match the dark color that X was at.  This raised an issue though where there are images outside of our embedding that have that other grayish color that seem to be part of core.  I believe we can try to overwrite them, but feels like out of scope of this ticket since it's related to embeds.  What do you think?

@codechefmarc Don't you agree that the Drupal-specific png files used for things like "audio.png", "video.png", "generic.png" should be a part of `ys_core` since they seem to come from Drupal core?

miketullo95 (approved):
approved! 👍

PR #629, opened 2024-04-11
YSP-470: Media library images slow to load

vinmassaro:
### Description of work
Pam reported the media library images are very slow to load after a cache flush. After debugging, the media library display mode was not enabled for media library images, so full size images were loading and being scaled down.

This updates the media library thumbnail style to use focal point and increase to 350x350 for clearer previews, and
enables the Media Library display mode for the Image media type.

### Functional testing steps:
- [ ] View the media library on an existing site and open a thumbnail in a new tab and observe that it is the full original image
- [ ] On the PR-generated multidev, open a media library thumbnail image and see that it is now using a scaled down thumbnail


dblanken-yale (approved):
Wow that's so much faster!  Nice find and nice fix.  I tried to grep through and didn't notice anywhere that isn't a thumbnail that's using that, so I think this is great.

PR #631, opened 2024-04-12
Ysp 461

joetower:
### Description of work

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...


miketullo95 (approved):
this looks great! so exciting

codechefmarc (approved):
This looks great! I have a few comments, but no blockers:

1. We should check with Katy, but in looking at the grid, it may look better if the CTA's are all aligned on the bottom?

Screenshot 2024-05-13 at 3 19 03 PM

with some flex applied:

Screenshot 2024-05-13 at 3 20 13 PM

2. If I didn't fill out the "caption" field on Localist, the image doesn't get an alt text value, so Editoria11y flags it. Since this is a Drupal field, content folks can simply add one in Drupal (or Localist and then re-sync) and that will work. Or we can provide a default value? I don't like the default value as generally that won't work for good alt text. I'll bring it up in a check in meeting.

dblanken-yale:
When visiting the test event, on the multidev, it doesn't seem to be bringing in the times of Katy's Event, making it look a little odd:
Screenshot 2024-05-18 at 9 08 01 AM

The data seems to be present when going to edit

dblanken-yale:
Nothing that it must be just the wiring since her date shows here in the cards:
Screenshot 2024-05-18 at 9 11 40 AM

codechefmarc:
> When visiting the test event, on the multidev, it doesn't seem to be bringing in the times of Katy's Event, making it look a little odd: Screenshot 2024-05-18 at 9 08 01 AM
> 
> The data seems to be present when going to edit

@dblanken-yale - This one is fixed on the multdev. Something happened with the build and it didn't get the new templates. I just had to rebuild it.

PR #634, opened 2024-04-15
YSP-398: Localist: Drupal views

codechefmarc:
### Description of work
- Updates existing calendar list view to support new event types filter
- Updates existing calendar list to show only one event per date, not multiple if event has multiple dates
- Updates views tool to work with multiple dates, same as above, only show one event per date
- Adds a new view for taxonomy terms. For the moment, this is used for event types.
- Adds event type URL path auto pattern

### Functional testing steps:
- [x] Login to the multidev here: https://pr-634-yalesites-platform.pantheonsite.io
- [x] Add a page, title it, publish, and enter layout builder.
- [x] Add a block of type "Calendar List"
- [x] Save Layout Builder
- [x] Verify that the event list shows events and that there are no repeating events. Note that the two events already in the system (non-Localist events) have end dates that are in 2025, which is why they are showing up first.
- [x] Verify a new filter at the top of the view that allows to filter by event type.

Screenshot 2024-04-16 at 3 16 34 PM

- [x] Click through to one of the Localist events (Year of the Dragon, for example)
- [x] Scroll down to the "Event Types" section (these are all temporary sections ahead of styling which is in another ticket)
- [x] Click on the "Exhibitions" link
- [x] Verify that the link has a real path, not an internal Drupal path
- [x] Verify that this page contains a list view of nodes tagged with the event type specified

Screenshot 2024-04-16 at 3 20 24 PM

- [x] Go back to the new page created in step 2, and enter layout builder
- [x] Add a block of type view
- [x] Select "Events" as the type, and then feel free to a view mode, filters, etc.
- [x] Add the block, save layout builder, and verify that the view block works as intended.


dblanken-yale:
This is really good.  Works as expected and code looks easy to follow.  Nice work! @codechefmarc

PR #636, opened 2024-04-22
YSP-485: Card external link styling adds spacing with the right amount of characters

dblanken-yale:
## [YSP-485: Card external link styling adds spacing with the right amount of characters](https://yaleits.atlassian.net/browse/YSP-485)

This is an exploration into [LinkPurpose](https://github.com/itmaybejj/linkpurpose) to see if we could utilize this and combine efforts with Princeton since we're working on the same feature.  It would end up replacing [ys_links](https://github.com/yalesites-org/component-library-twig/tree/develop/lib/ys_link).


dblanken-yale:
Closing for now since we have so much work that needs multidevs.  Once I get feedback from Joe on some of the issues with using linkpurpose as an alternative to ys_links and some of the component interactions, I'll open a new PR.  Note these issues would arise also if I were to implement our own fix, since link purpose's fix is the proper fix for the issue at hand.

PR #647, opened 2024-05-14
fix(YSP-525): v1.3.0: Bug: Bulk publish/unpublish operations do nothing

dblanken-yale:
## [v1.3.0: Bug: Bulk publish/unpublish operations do nothing](https://yaleits.atlassian.net/browse/YSP-525)

Since we are using content moderation, using the existing bulk operations of publish and unpublish result in a warning notifying the user of the use of content moderation, followed by error messages reinforcing the same thing.  Since this is now not able to be done, we remove these options from being selected in the bulk operations drop down.

### Description of work
- Remove `publish` and `unpublish` from possible bulk operations in the following views:
  - Content
  - Manage Events
  - Manage Pages
  - Manage Posts
  - Manage Profiles
    - This originally allowed all bulk operations instead of a subset as the other did.  This also makes Profiles behave like the others by limiting the bulk operations to a subset.  They now all match.

### Functional testing steps:
- [x] Visit each view above
- [x] Select multiple items in bulk
- [x] In the bulk operations dropdown that appears in the bottom, ensure that `publish` and `unpublish` are not available options.

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/views.view.content.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/views.view.content.yml
@@ -81,8 +81,6 @@ display:
           include_exclude: include
           selected_actions:
             - node_delete_action
-            - node_publish_action
-            - node_unpublish_action
             - pathauto_update_alias_node
         title:
           id: title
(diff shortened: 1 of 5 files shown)


miketullo95 (approved):
👍

kara-franco (approved):

kara-franco:
Pub/Unpub actions are no longer present!

codechefmarc (approved):
Works for me too. Approved!

nJim (approved):

PR #650, opened 2024-05-14
YSP-529: V1.3.0: remove the configure button on moderation placeholder in /layout

codechefmarc:
### Description of work
- Content editors can no longer remove default blocks placed in the banner section.

### Functional testing steps:
- [ ] Login to the site as a site administrator, editor, or contributor role
- [ ] Add a new page to the site, title it, and save (publish or draft is fine)
- [ ] In Layout Builder, verify that if you click the "Configure" pencil next to the "Placeholder for the Moderation control field", the remove option no longer exists. Note that this only works for newly added content which is fine as the moderation control field is the same - only newly added content.

Screenshot 2024-05-14 at 11 55 10 AM


dblanken-yale (approved):
So surprised I missed that the lock would happen on "default blocks".  This is working beautifully.  Thanks a lot for taking this one.  Ship it!

PR #653, opened 2024-05-16
YSP-517: Tech debt consistent field settings

edwtorba:
### Description of work
- Check the option 'Create new revision' for all blocks.
- Use the label 'Theme' in the field field_style_color.

### Functional testing steps:
- [x] Create some custom blocks: https://pr-653-yalesites-platform.pantheonsite.io/block/add
- [x] Check that the following fields are required and have a default value:
  -  Alignment
  - Columns
  - Position
  - Style
  - Theme
- [x] Check that all block has revision.


nJim (commented):
Everything looks good. Thank you for the attention to detail. These kind of changes are not glamorous but it helps keep the site healthy.

PR #660, opened 2024-05-20
YSP-281: Localist Project

codechefmarc:
### Description of work
All finalized and approved work from the Localist Project which includes the following:
- Many new fields added to the Event content type. Fields are all described in the Data Model spreadsheet. Many of the fields can be used without Localist integration. Some fields are for use with Localist only and are greyed out on the event edit form.
- Many new taxonomy terms that are used with the Event content type to support things like event topics, places, etc.
- A new module `ys_localist` that contains a settings form, migrations, and custom migration plugins to support importing events from the Localist platform via their APIs.
- Migrations are run every hour on cron and can also be run manually via the settings form.
- Styling is included in the corresponding [Atomic PR 237](https://github.com/yalesites-org/atomic/pull/237) and [Component Library PR 375](https://github.com/yalesites-org/component-library-twig/pull/375) PRs.
- Testing instructions and screenshots listed below, but most of this has already been tested and approved.
- Note the [README.md](https://github.com/yalesites-org/yalesites-project/pull/660/files#diff-cf6bbc8b063ba847735cb5e2005a186caa99c3478222d9820a63963fcc704c24) file in the `ys_localist` module contains more information about the migrations and how to add additional filters, if needed.

### Functional testing steps:
- [ ] New fields and field layout in Event content type:
Create Event _ Your Site Title
Screenshot 2024-06-11 at 4 00 52 PM
Screenshot 2024-06-11 at 4 00 58 PM
Screenshot 2024-06-11 at 4 01 05 PM

- [ ] Settings form at `/admin/yalesites/localist` (Note, only available to Platform Admins)
Screenshot 2024-06-11 at 4 03 44 PM

- [ ] New styling for the top of Event pages
Marc_s Test Event _ Your Site Title · 4 05pm · 06-11

- [ ] New styling for list and card views
Screenshot 2024-06-11 at 4 06 53 PM

- [ ] New taxonomy views at `/events/event-type/[event-type]` and `/events/event-topic/[event-topic]`
Conferences Meetings _ Seminars _ Your Site Title


codechefmarc:
The latest changes include:
1. Localist settings - if you are not a platform admin, users are prevented from changing the Localist endpoint, groups, and the enable checkbox.
2. Place name added to the location on the front-end
3. Deprecation label added to old event type field
4. Edit event form - disabled Localist-only fields as they only should contain data from a sync
5. Added a function to remove old "experience" terms (Online and In-person) to be replaced with Localist terms (Hybrid, Virtual, and In-Person). This function only loads when installing Localist for the first time.
6. Added stream URL and stream embed code to the event. It is not on the front-end yet (awaiting design)

miketullo95:
@codechefmarc 
### AX Notes

- can we put the stream URL and stream embed code in a group like event and place? 
- once in a group, maybe move stream down to last group since I feel like that may be lower priority than place, website, or dates
- can the place group be renamed to "location" or is that not changeable due to localist?
- if place group can be renamed to "location", rename the place dropdown to "Yale Location"
- thoughts on renaming the "Classifications" tab to "Taxonomies"? It would be nice to keep that language accurate since I think these are all just taxonomy vocabs right? If a user wants to add a native event, they might be lost trying to find Event Category.
- Are users assigning `event category` in localist too? if not, maybe we can order these with native taxonomies first (Tags, Event Category), then the localist ones... maybe even group them.

dblanken-yale:
Found this while taking a look at the localist fields.  ☹️ Are all CL changes in this too?  That was just creating a new event manually and going to the localist data tab.

Screenshot 2024-06-17 at 4 23 51 PM

dblanken-yale:
Are there any implications on cloning a localist event?  I noticed when I did so I could modify the localist data tab (even though it says it's disabled), but after saving I could no longer.  If we allow this, do we have documentation to instruct the users how to modify the clone to not be updated on cron?  For instance, is it just deleting the Localist ID to make it stay?

dblanken-yale:
I guess along those lines, if a user forgets and does not modify the Localist data, does the feed update all instances matching that localist ID, or the first it encounters?  Or is this actually accomplished with that intermediate table the feed creates to match up?

codechefmarc:
> Found this while taking a look at the localist fields. ☹️ Are all CL changes in this too? That was just creating a new event manually and going to the localist data tab in Safari.
> 
> Screenshot 2024-06-17 at 4 23 51 PM

Wow, that is a good find there. I think what's happening is that Chosen JS is taking over the click event so it is bypassing the disabled control. I'll just turn that field into a normal select box instead of a Chosen one.

codechefmarc:
> Are there any implications on cloning a localist event? I noticed when I did so I could modify the localist data tab (even though it says it's disabled), but after saving I could no longer. If we allow this, do we have documentation to instruct the users how to modify the clone to not be updated on cron? For instance, is it just deleting the Localist ID to make it stay?

Good find on the disabled fields for cloned form. I fixed that. As to cloned events in general, nothing prevents anyone from cloning an event on the Drupal side, but the clone is then no longer connected to the migration. The original node is (based on node ID) so you can edit the clone as you see fit. Any new data coming from Localist will only affect the original node.

dblanken-yale (approved):
There is a lot of great work here.  Glad it all came together well!  Even realized something new with the scope of module functions.  :D

Nice work!  Let it go on in so we can get it in v140!

PR #642, opened 2024-04-30
Ysp 321  profile vertical images

joetower:
### Description of work
- Adds dial option for portrait oriented images in the Profile Banner, - `portrait` or `landscape`
- Adds dial option for alignment - `left` or `right`
- Adds dial option for image style - `inline` or `outdent`

### Testing Link(s)
- [x] Test in component library here: https://github.com/yalesites-org/component-library-twig/pull/362

### Functional Review Steps
- [ ] Navigate to: https://pr-642-yalesites-platform.pantheonsite.io/profile/alyssa-dechiaro
- [ ] Verify the image in the banner is `portrait` in orientation and styled with the `outdent` option
Screenshot 2024-05-07 at 14 31 31

- [ ] If you login and edit the profile, verify that you can: 
  - [ ] Change the image orientation - `portrait` or `landscape`
  - [ ] Change the image alignment - `left` or `right`
  - [ ] Change the image style - `inline` or `outdent`

Screenshot 2024-05-07 at 14 31 02

- [ ] Test each option - `landscape` orientation, `inline` style or `right` alignment

PR #661, opened 2024-05-21
YSP-306: Spotlight: Content area allow order/unordered lists

codechefmarc:
### Description of work
- Adds an interface to allow updating of text formats for text fields on existing placed blocks in layout builder
- Uses the new function in a deploy hook to change all content spotlights (landscape and portrait) to use the new text format of Basic HTML to allow lists and tables.
- NOTE: This PR includes Jim's lock work from [PR 642](https://github.com/yalesites-org/yalesites-project/pull/642) so that PR should probably be merged in before this one.

### Functional testing steps (Spotlights):
- [x] Enter layout builder for a node that has a spotlight of either kind already placed
- [x] Verify that the spotlight content field now has the ability to add lists and tables (this is the basic HTML formatter)

Screenshot 2024-05-21 at 12 43 22 PM

### Functional testing steps (New text format changer UI):
- [ ] Since one of the ways this PR works is via a deploy hook, most of the blocks that had the old text format have already been converted to the new text format for spotlights. Therefore, one of the tests of this PR will be to change a text format temporarily to test the UI portion of the code (since config will not be exported, this is ok to do on a multidev)
- [x] Login to the site as an admin
- [x] Visit a page with an Action Banner block (for example, https://pr-661-yalesites-platform.pantheonsite.io/generic-homepage)
- [ ] (Note, for some reason the "Welcome" page does not get these updates? All other nodes appear to.)
- [x] Enter Layout Builder and configure the Actions Banner block
- [x] Verify that the Action Banner Content field has the Heading HTML text format - i.e. only bold and italic are available.

Screenshot 2024-05-20 at 8 25 40 PM

- [x] Close the configure and save the layout. NOTE: It is important to actually save the layout here, not just leave the page.
- [x] Visit the manage fields for the Action Banner: https://pr-661-yalesites-platform.pantheonsite.io/admin/structure/block-content/manage/cta_banner/fields
- [x] Edit the "Action Banner Content" field
- [x] Change the "Allowed text formats" from Heading HTML to Basic HTML and save the settings.

Screenshot 2024-05-20 at 8 29 42 PM

- [x] Return to the layout builder interface of the Generic Homepage node and configure the action banner.
- [x] Verify that the Action Banner now is in an odd state - As an admin it is editable, but there is no CKEditor toolbar. As a site admin or less, this text field would not be available at all to edit.

Screenshot 2024-05-20 at 8 31 38 PM

- [x] Close the block and save the layout. NOTE: It is important to save the layout here and now just leave the page.
- [x] Visit https://pr-661-yalesites-platform.pantheonsite.io/admin/config/ys-layouts/settings
- [x] Verify that there is a new area in this settings page called "Update text formats"

Screenshot 2024-05-20 at 8 33 53 PM

- [x] This area will let you update all existing placed blocks with the new text format. A few notes: Any NEW blocks added will get the new text format set correctly. This interface simply updates existing placed blocks. Also note: If any layout is still in the middle of being edited, it will not get the update. That is what the bottom section "Nodes that have not received layout updates" is for. If a node is listed there, it will not even get updates from the update mechanisms on this page. This is why you have to go in to save the layout for these nodes before they will receive an update.
- [x] Select "Action Banner" from the list of blocks in the "Block type" dropdown
- [x] Select "Action Banner Content" for the field name.
- [x] Click "Update text format"
- [x] Verify there is a success message at the top of the page

Screenshot 2024-05-20 at 8 37 25 PM

- [x] Return to the Generic Homepage node and enter layout builder
- [x] Configure the action banner and verify that it now has the correct text format


dblanken-yale (approved):
Really cool functionality.  I'll reset the Action Banner Content back so others can test this in the same way, but everything looks great.  And the code looks like we can extend it in the future with non-text fields by using your work as a template.

web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.profile.default.yml
```diff
@@ -97,6 +97,7 @@ third_party_settings:
       - 'Content fields'
       - 'Custom block types'
       - 'Custom blocks'
+      - Devel
```
nJim, line 100: @codechefmarc I'm not sure ow this keeps creeping in. Can you remove this requirements and look for instances that crept into the codebase?

nJim (changes requested):
Just one small change the on 'devel' dependency sneaking in. Everything else works as described. Great job. A couple of thoughts for future work:
-We may want to create an update all button
-We may want to include paragraphs, nodes, and other entities

nJim (approved):

PR #664, opened 2024-05-24
YSP-479 Wrong alignment for Reusable Blocks in Posts, Profiles and Events

laura-johnson:
### Description of work
- Adds proper alignment for reusable accordion blocks in posts and events

### Functional testing steps:
- [ ] Add a reusable and non-reusable accordion block to an event or post. Verify that the alignment is the same for both.


codechefmarc (approved):
Works great! Thanks for finding this. Approved!

PR #665, opened 2024-05-24
YSP-431 Wire/Add/Tweak Call-to-Action button utility nav

laura-johnson:
### Description of work
- Adds the CTA button to the header if the CTA fields are populated in the header form

### Functional testing steps:
- [ ] Go to /admin/yalesites/header, verify there is a new Call To Action section with Link Text (text) and Link Target (linkit) fields
- [ ] Add values for either or both fields
- [ ] Verify if both fields are populated, the Call to Action button appears in the upper right of the navigation as per the designs

Requires https://github.com/yalesites-org/component-library-twig/pull/370

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Form/HeaderSettingsForm.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Form/HeaderSettingsForm.php
@@ -165,6 +165,18 @@ class HeaderSettingsForm extends ConfigFormBase {
       ],
     ];
 
+    $form['call_to_action_container'] = [
+      '#type' => 'details',
+      '#title' => $this->t('Call to Action'),
+      '#states' => [
+        'disabled' => [
+          ':input[name="header_variation"]' => [
+            'value' => 'focus',
+          ],
+        ],
+      ],
+    ];
+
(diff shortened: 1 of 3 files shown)


laura-johnson:
It works well and I think it looks good except that if it's on the blue header background the button gets a bit lost... I'm not sure how to handle that, it might be a task for @joetower 

apply-now-button

joetower:
@laura-johnson I pushed up a fix for this.

If you test it now, the utility nav CTA button will use the same background color as the menu item's `:hover` color. This is the default behavior, as the color variables cascade. The CTA elements themselves can use a color theme, but in this instance, I've set the theme to `none` which made the color variables point at the other `action` colors in the header - which is the hover states for the menu items. 

I'm wondering if we should set/hardcode these button bg colors to `white` if the theme is set to `two` or `three` (when the header has a darker background color) instead of letting these other colors apply. 🤔 

I'd like folks from Yale to give their opinions. @miketullo95 or @kara-franco , what are your thoughts?

https://github.com/yalesites-org/yalesites-project/assets/366413/6ec4bac4-6d16-4aed-b40f-63b771526082

nJim (approved):
Approved in weekly meeting with Kara and Mike T.

PR #666, opened 2024-05-28
YSP-546 - Subnavigation demo

edwtorba:
### Description of work
- Install and configure book module.
- Add the 'Book Navigation' in the layout builder.
- Add the 'Main Navigation' block to the layout builder.
- Install the contrib module 'Group' and its dependencies.
- Activate the 'Group' module and its dependencies.
- Basic configuration of a 'Site Group'.
- Apply a patch to avoid error on Group configurarion panel.
- Create and config 'Group Navigation'block.
- Switch to pinned version of modules of the new modules.
- Update quick_node_clone module to fix issue with group module, update the paragraphs patch too

### Functional testing steps:
- [ ] Create a new "Page" content-type: https://pr-666-yalesites-platform.pantheonsite.io/node/add/page
- [ ] Navigate to the layout builder option and add a new section of type One Column.
- [ ] Now add a new block in the previous section.
- [ ] Check that the category Advanced has the following blocks:
  - [ ] Main Navigation
  - [ ] Book Navigation
  - [ ] Group Navigation


nJim:
Thank you @edwtorba This is exactly what we needed to put together a demo. We will do a lot more with the Group module soon. I am closing this PR as we need to complete a planning cycle before picking this work back up.

PR #667, opened 2024-05-29
YSP-532: Increase grand hero video size maximum

dblanken-yale:
### Description of work
- Increases the max file size of a background video from 12 MB to 25 MB

### Functional testing steps:
- [ ] Attempt to upload a background video
- [ ] Verify that the max allowed says 25 MB
- [ ] Verify that you can upload a MP4 that is within that range

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/field.field.media.background_video.field_media_video_file.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/field.field.media.background_video.field_media_video_file.yml
@@ -15,13 +15,13 @@ label: 'Video File'
 description: ''
 required: true
 translatable: true
-default_value: {  }
+default_value: {}
 default_value_callback: ''
 settings:
   handler: 'default:file'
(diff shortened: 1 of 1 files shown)


miketullo95:
will look into this!  Did we find the preferred method of compressing video? would be nice to have maybe a tips and tricks with a recommended service to users if they have a larger file

dblanken-yale:
@miketullo95 I posted some tests in the Jira ticket; all are great in order to get to below 25MB I think.  Adobe Media Encoder is in our license so maybe that should be it?  But they'd have to play with the presets which might not be ideal.

miketullo95 (approved):
added a 17MB video to grand hero here: https://pr-667-yalesites-platform.pantheonsite.io/

looks great!

codechefmarc (approved):
Tested with a 21 MB video, worked great. Approved!

nJim (approved):
Updates are as described.

PR #669, opened 2024-05-31
YSP-556: Media library selection selects wrong item

dblanken-yale:
### Description of work
- Adds hook_media_insert and hook_media_delete to clear the media_view and media_list caches

#### Adding a new image
- [ ] Add a media image from the media library
- [ ] Go to a page and add an image block
- [ ] When selecting an image, go to a different page than the first, and select an item; make note of the image you clicked on
- [ ] When inserting that image, make sure the image showing up in the block configuration is the same image you selected.

#### Deleting an existing image
- [ ] Delete a media image from the media library
- [ ] Go to a page and add an image block
- [ ] When selecting an image, go to a different page than the first, and select an item; make note of the image you clicked on
- [ ] When inserting that image, make sure the image showing up in the block configuration is the same image you selected.


dblanken-yale:
@nJim @codechefmarc Including you as this fix feels very heavy handed, but works.  I attempted to pinpoint the real cachetag, but couldn't seem to target it.  If you have any insight or recommendations to make it not do a total render cache clear, I'm down to try it.  So far I tried `media`, `media_view`, and `media_list`.

dblanken-yale:
Thanks to @codechefmarc who was able to find two patches to fix this from the 4k team!

miketullo95:
testing this, it worked with the existing images. However, i added a new image in the media library and now its out of sync again.  @dblanken-yale 

https://github.com/yalesites-org/yalesites-project/assets/98550395/27a60d9d-18b6-450c-becc-9dd49db05def

dblanken-yale:
Grrrr, ya, that isn't what I experienced there.  I'll revert back to my clear cache solution since it was repeatably fixing it until we can figure out more.  It should now be building with the cache clear solution.

miketullo95 (approved):
tested and this works for me!

PR #671, opened 2024-06-01
YSP-299 - Update "term" description in Views block]

edwtorba:
### Description of work
-  Replace the word 'terms' with 'tags or categories'.

### Functional testing steps:
- [ ] Create or edit a Page or Event content type. https://pr-670-yalesites-platform.pantheonsite.io/
- [ ] Navigate to layout settings.
- [ ] Add a new block of type "View".
- [ ] Scroll down and check you can see the following fields:
  - [ ] "Include content that uses the following tags or categories" instead of "Include content that uses the following terms".
  - [ ] Exclude content that uses the following tags or categories" instead of "Exclude content that uses the following terms".


nJim (approved):
Simple change. Thank you for the update.

PR #673, opened 2024-06-04
YSP-548 Views: Offset results

laura-johnson:
### Description of work
- Adds offset field to viewsDefaultWidget
- Also fixes an error "undefined array key operator"

### Functional testing steps:
- [x] Add or edit a view using the views tool
- [x] Verify you see the Offset field and it defaults to zero
- [x] Verify setting an offset offsets the results correctly
- [x] Verify there are no errors with new or existing views


codechefmarc (approved):
Tested this with events and it worked great! I had forgotten that are also sorting with "sticky at top of lists" for all views, so it threw me off with the offset too. But I double checked with the core views UI and it worked exactly the same. Long way of saying, approved! Thanks for adding this.

nJim:
@miketullo95 Will post to this PR with suggested label changes and help text for the offset field.

dblanken-yale (approved):
Works as described.  Nice work!

miketullo95:
Header: Ignore Number of Results
Help Text: Specify the number of results you want to ignore. If you enter "2", your view will omit the first two results that match the overall parameters you've set in the view interface.

laura-johnson:
@miketullo95 I added the new field title and description as per your comment.

miketullo95 (approved):

PR #676, opened 2024-06-12
YSP-487: Add JWPlayer Embed

dblanken-yale:
### Other work related to this
- [Atomic: PR #238](https://github.com/yalesites-org/atomic/pull/238)
- [Component Library Twig: PR #376](https://github.com/yalesites-org/component-library-twig/pull/376)

### Description of work
- Adds jwplayer embed
- Adds jwplayer image

### Functional testing steps:
- [ ] Attempt to create a new embed block
  - [ ] Add the example embed or one that was from the embed->iframe on the jwplayer dashboard
  - [ ] Paste the copied markup
  - [ ] Observe that it successfully is recognized and usable


miketullo95:
i was able to embed the example jw player! it couldn't play, but i'm assuming thats because its not live. If thats the case, then this looks great! 

Do we know how orgs like the music department get these embed codes and do we know that there is one "correct" way to get this? we may want some resources on the website around this if other orgs want to get into live streaming

miketullo95:
also maybe we could get a live livestream to see how the player looks in action too

dblanken-yale:
@miketullo95 Ya Matt offered to turn on a livestream for us so maybe we can.  He had it on on Friday but I couldn't get to test it with it on in time.  ☹️

Currently they use the jwplayer dashboard to get a code that's in the table shown and paste it into their content type, however, there is an embed button on the rows they look at which has the iframe copy/paste markup.  That's the way we decided to go since a unique identifier can't be regex evaluated to determine it's jwplayer, but the iframe paste can be evaluated to know.

I also received some feedback from Matt on this, such as allowing fullscreen to be passed to the iframe, so that is building now, as well as proper height formatting of the player.

codechefmarc (approved):
Also looks good to me, approved!

dblanken-yale:
@miketullo95 @kara-franco Building now; I think I squelched the width issues with it while allowing it to expand the correct height of the player.

PR #677, opened 2024-06-12
YSP-487: Add publishcontent back

dblanken-yale:
## [YSP-487: Add publishcontent back](https://yaleits.atlassian.net/browse/YSP-487)

Builds could not finish due to `publishcontent` being enabled, but the module was removed.  This re-adds it back to the composer file.

### Description of work
- Adds the `publishcontent` module back to the composer.json file

### Functional testing steps:
- [ ] If this build succeeded and you can visit the site, all is well

### DIFF
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -83,6 +83,7 @@
     "drupal/paragraphs": "1.16",
     "drupal/paragraphs_features": "2.0-beta3",
     "drupal/pathauto": "1.12",
+    "drupal/publishcontent": "1.6",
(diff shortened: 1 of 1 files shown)


codechefmarc (approved):
A ha! So, I was really confused by this, but I see why now. It's because the dev site on the Yalesites Platform hasn't received 1.3.0 update yet. So, when running `npm run build-with-assets` this fails. Your update fixes, but wanted to let you know I did put in ticket YSP-291 a while back that is for removing this when the time is right. Long way of saying, approved! :)

dblanken-yale:
Closing this as it's included in the JW Player PR that went in

PR #678, opened 2024-06-17
Localist: Add conference support and prevent cloning

codechefmarc:
### Description of work
- Updates migration to sync "event kind" to a new taxonomy term for event kind.
- Event kind is passed into the event meta block and if it is a conference, a check in the component library template will print out a marketing alert that it is a conference and provide a link to the page on Localist.
- This is a workaround as individual sessions from a conference are not currently able to sync via the Localist API
- Also prevents the ability to clone events - this is because cloned events on Drupal where the original came from Localist, the clone will not receive any updated data from Localist.

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...


dblanken-yale:
Closing for now, but keeping the branch for when we might need it in the future.

PR #682, opened 2024-06-20
fix: ysp-445 set css in the head to fix sitewide alert flicker

laura-johnson:
### Description of work
- Adds JS in the head which in turn adds CSS in the head if there is an alert which hides the alert if it is dismissed.

### Functional testing steps:
- [ ] Create an alert
- [ ] Dismiss the alert
- [ ] Verify there is no flicker

Notes:

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_alert/js/anti_flicker.js
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_alert/js/anti_flicker.js
@@ -1,0 +1,30 @@
+/**
+ * @file
+ * Prevents flicker of the toolbar on page load.
+ */
+
+(() => {
+
+    // Function to get all alerts from localstorage.
+    const setHtmlAlertStyle = () => {
+        const alertStyle = document.createElement('style');
+        let styleContent = '';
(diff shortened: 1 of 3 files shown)


joetower (approved):
@laura-johnson From a code perspective, I think adding inline styles might be the best solution right now.

We could experiment with adding variables in the twig template, however, I'm not sure what that would look like. Almost as if we need a check higher in the DOM than the alert element itself to prevent rendering the alert component at all if it has been dismissed. Currently, the alert element is what gets the `data-alert-state` property, so outside of inline styles, anything in CSS would need to load on the page to apply to the alert element CSS - reading that property and applying the styles. We also have animation contributing to the flash of the alert. If we were to add a `display: none`, perhaps this flash would be imperceptible 🤔 . It would be something to consider in another iteration, especially now that it is technically possible to animate the `display` property but not 100% supported in all browser, yet.

dblanken-yale (approved):
Very clever and works!  Nice find!!!!!

dblanken-yale:
Closing for now as we believe there was talk about possibly moving this somehow to the backend where the node knows beforehand if it should render alerts.

PR #690, opened 2024-06-26
YSP-535: Action Banner Spacing -- MULTIDEV ONLY; DO NOT MERGE

dblanken-yale:
## [YSP-535: Action Banner Spacing](https://yaleits.atlassian.net/browse/YSP-535)

[Real work located in Atomic](https://github.com/yalesites-org/atomic/pull/250)

This also fixes the [previous spotlight issues](https://github.com/yalesites-org/component-library-twig/pull/383), so removal of that fix is also included in [component library twig](https://github.com/yalesites-org/component-library-twig/pull/385)

In this specific case, there exists an empty div where the content moderation block renders.  This empty div does not allow the action banner to appear flush with the header.  We overload the content moderation twig to pass on content if it had it (i.e. the form), but display nothing otherwise.

### Description of work
- Overloads the `block--extra-field-block--node--page--content-moderation-control.html.twig` to pass content through if it exists, or nothing

### Functional testing steps:
- [ ] Add a new page
- [ ] Add an action banner to the banner
- [ ] Verify that the action banner is flush with the header line after save and refresh
- [ ] Add a new page
- [ ] Add a spotlight landscape block to the banner top
- [ ] Verify that the spotlight landscape block is flush with the header line after save and refresh
- [ ] Add a new page
- [ ] Add a spotlight portrait block to the banner top
- [ ] Verify that the spotlight portrait block is flush with the header line after save and refresh

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ The YaleSites platform empowers the Yale community to create digital experiences
 ## Common Scripts
 
 These are the most common npm scripts you may find yourself using:
-(Each is prefixed with `npm run `)
+(Each is prefixed with `npm run`)
 
 ### PR Reviews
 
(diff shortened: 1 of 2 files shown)


miketullo95 (approved):
looks great!

PR #693, opened 2024-07-01
YSP-576: Enable new block permissions

dblanken-yale:
## [YSP-576: Enable new block permissions](https://yaleits.atlassian.net/browse/YSP-576)

While looking at another issue, we noticed that the permissions for the new blocks were not set.  While this doesn't hinder users from using them due to relying on Layout Builder Restrictions, it is our convention to still enable them in case another module relies on these.  This enables them to mimic other blocks in the past.

### Description of work
- Enables permissions on new blocks matching that of older blocks

### Functional testing steps:
- [ ] I don't believe we will experience any changes in current functionality
- [ ] Verify that the items in the files changes do show as enabled for each role

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/user.role.contributor.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/user.role.contributor.yml
@@ -14,15 +14,20 @@ dependencies:
     - block_content.type.divider
     - block_content.type.embed
     - block_content.type.event_list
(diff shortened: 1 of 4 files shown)


dblanken-yale:
@kara-franco @miketullo95 Just need an approval on this to get the permissions through; this includes the inline message block permission now as well.

miketullo95 (approved):
👍

PR #695, opened 2024-07-02
YSP-579: Event dates not always 0 based array

dblanken-yale:
## [YSP-579: Event dates not always 0 based array](https://yaleits.atlassian.net/browse/YSP-579)

For some reason, there are times when the dates array is not 0-based. So instead of relying on this, we reset to take the first item.

### Description of work
- Replaces reference of `$dates[0]` with a variable that uses `reset($dates)` to retrieve first item

### Functional testing steps:
We are not totally sure what gets it to this state, but:
- [x] Create an event
- [x] Clone the event
- [x] Review that both load
- [x] Make changes to the cloned event, adding or deleting dates and other items
- [x] Review that both load


miketullo95 (approved):
Walked through functional steps and everything looks good!

Original Event: https://pr-695-yalesites-platform.pantheonsite.io/events/mike-test-event
Cloned Event: https://pr-695-yalesites-platform.pantheonsite.io/events/clone-of-mike-test-event

dblanken-yale:
Deleted multidev since this change won't show any differences in functionality.

kara-franco (approved):

vinmassaro (approved):
LGTM

PR #697, opened 2024-07-08
YSP-588: Pagination after first page doesn't comply with ascending/descending dates for events

dblanken-yale:
While the standard view for the views block was using the proper sorter, the event specific one was not.  This lets it use it to drive the sort operation.

### Description of work
- Makes the event-specific views "view" to use the `basic_views_sorter`

### Functional testing steps:
- [ ] Make lots of events in the past
- [ ] Create a view for past events, ordering by date in descending order
- [ ] Make sure when traversing the view that it respects that decision

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/views.view.views_basic_scaffold_events.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/views.view.views_basic_scaffold_events.yml
@@ -143,20 +143,20 @@ display:
             format: basic_html
           tokenize: false
       sorts:
-        field_event_date_value:
-          id: field_event_date_value
-          table: node__field_event_date
-          field: field_event_date_value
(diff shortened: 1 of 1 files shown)


miketullo95:
@dblanken-yale testing this out (see view on https://pr-697-yalesites-platform.pantheonsite.io/) i paginated the event view and I created some real old events from 1989 and it seems like its out of order in my event view. The 1989 events are in the middle and some 2024 events came after it 😭

miketullo95 (approved):
after review, this looks to be working! will make a separate ticket that you can make an event without a date. That feels wrong haha

miketullo95:
> after review, this looks to be working! will make a separate ticket that you can make an event without a date. That feels wrong haha

here is that ticket https://yaleits.atlassian.net/browse/YSP-594 😄

PR #700, opened 2024-07-22
YSP-446: Drupal 10.3 Fixes

dblanken-yale:
### Description of work
- Reroll combo patch
- Bump focal point so that patch works with it
- Bump reusable block patch to work with 10.3

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...


vinmassaro (approved):
Do these patches need to be reposted on their respective issues for a 10.3 compatible version? I bumped both issues hoping to get maintainer review, but nothing yet.

dblanken-yale:
@vinmassaro Oh hadn't thought about that.  Ya that might be a good idea.  I just took the original patch and mimic'd it so not sure the right patch for each individually for 10.3 alone.  I'll see what I can do to get good individual versions.

PR #701, opened 2024-07-23
YSP-551: Toggling a block as reusable on a page leads to white screen error page

dblanken-yale:
## [YSP-551: Toggling a block as reusable on a page leads to white screen error page](https://yaleits.atlassian.net/browse/YSP-551)

It turns out cloned blocks were actually still the same block, but with a different revision.  Because of this, when you set it up to be reusable, it would actually fail the save, making the block in an inconsistent state as we then try to create a new block to replace the now reusable (we thought) block.  This attempts to save in the regular way, but in a failure, it assumes it's in this state and breaks off to duplicate itself as a new block so that it can become reusable.

### Description of work
- Add to patch to check if block was saved
  - If it wasn't saved, it clones the block such that it can be

### Functional testing steps:
- [ ] Not sure how easy it will be to reproduce since new and cloned blocks ensure blocks are separate by default.  I could show on local development by bringing down site if that helps.

#### Locally
- [ ] Get this branch down locally and bring down https://development-ys-ism-yale-edu.pantheonsite.io
- [ ] CAS in
- [ ] Make changes to the "ISM and the Divnity School" page (/node/125/layout)
- [ ] Make the "Our Partners" quick links block a reusable block
- [ ] Ensure no errors occur


dblanken-yale:
For reference, the patch was rerolled, but the true change was the addition of [this code](https://github.com/yalesites-org/yalesites-project/pull/701/files#diff-1b4fb24626c6aa9397c4f3c73851dbb6d25b89bbed40da43c98ed4599fd538f4R275-R281).

vinmassaro (approved):

vinmassaro:
@dblanken-yale I added back to the core patch and separated out our local changes to their own patch so we can track them separately since the core patch will likely need updating.

I had trouble verifying behavior between `develop` and this branch though even after pulling down a local copy in your instructions, since I did not get an error running through the steps with the `develop` branch.

dblanken-yale:
@vinmassaro Strange that you didn't get the error locally using the ism development.  I get it consistently without the patch.

dblanken-yale:
I'll run it locally on develop one more time to verify as well.  Do you see it on their development site?  Think it's linked in the Jira ticket.

kara-franco (approved):
Reviewed the code/patch, seems good to me!

PR #704, opened 2024-07-25
YSP-611: Localist: Views errors after localist enable

dblanken-yale:
### Description of work
- If we can't get a taxonomy, make the name an empty string

### Functional testing steps:
- [ ] Create events with event types/experiences of Online or In-person
- [ ] Enable localist and sync
- [ ] Verify that "Online" and "In-person" no longer exist ("In-Person" will exist)
- [ ] Create a page with an events view that includes the events you manually created
- [ ] Load that saved page
- [ ] Verify it loads successfully
- [ ] Visit the logs to verify that either no logs exist from ys_localist


vinmassaro (approved):
Makes sense, LGTM

PR #710, opened 2024-07-30
YSP-534: Add Subscript and Superscript to WYSIWYG editor

dblanken-yale:
### Description of work
- Adds `<sub>` and `<sup>` tags to Basic HTML formats
- Add graphical buttons to the WYSIWYG editor for Basic HTML formats for subscript and superscript

### Functional testing steps:
- [ ] Add a block that utilizes Basic HTML formatters, like the text block
- [ ] Observe that the WYSIWYG contains super and subscript icons
- [ ] Use those to create such items as a H2O reference using subscript or a superscript numeric reference
- [ ] Verify when saved they properly show as super/subscript

### Questions for reviewers
- Is the button placement acceptable in the WYSIWYG editor?  Or should they be moved elsewhere on the bar?
- Should this only be on the Basic HTML formats?  Or should it be on the Restricted as well?


kara-franco (approved):
I love it, the faculty member we met with will be so happy! 

And to answer your Teams questions: the placement makes sense to me and the options should only be available in the Basic HTML.

vinmassaro (approved):
LGTM

PR #712, opened 2024-07-30
YSP-602: Localist: hybrid 'event experience' missing

dblanken-yale:
## [YSP-602: Localist: hybrid 'event experience' missing](https://yaleits.atlassian.net/browse/YSP-602)

Work also done in [Component Library Twig](https://github.com/yalesites-org/component-library-twig/pull/392)

### Description of work
- Adds update hook to add Hybrid if it doesn't exist
- Add Hybrid to the localist event types to remove when updating to the Localist experiences

### Functional testing steps:
Note: Best to do this locally since you can only experience one case

#### Default case: Localist not set up yet
- [ ] Verify that when creating a new event you now see "Hybrid" as an option
- [ ] Go to Settings->Localist Settings
- [ ] Set up localist (use the URL example replacing `enterprise` with `staging`)
- [ ] Save
- [ ] Sync Groups
- [ ] Type ONHA as the group and select it
- [ ] Save
- [ ] Sync (There should be like 1 imported)
- [ ] Visit that new event or add a new event
- [ ] Verify that Hybrid shows only once in the list

##### To emulate those that already have localist enabled before this:
- [ ] Set up local and comment out the `ys_core_update_9008` in `ys_core.install`
- [ ] Enable localist following the instructions above (NOTE: Hybrid will not currently be an option until you enable localist)
- [ ] Uncomment the `ys_core_update_9008` and run `npm run confim`
- [ ] Visit an event or make a new one and verify that there is only one Hybrid term in the experience drop down


dblanken-yale:
@atiddei @kara-franco Multidev is created to verify; localist is not enabled.

kara-franco (approved):
I see Hybrid in the dropdown now! I did not test the other cases with existing events.

PR #722, opened 2024-08-12
fix: allow actions/checkout to run on forked repos

vinmassaro:
### Description of work
- Sets `repository` during actions/checkout workflow step to allow runs from forked repos and falls back to the default if not set.

More info: https://github.com/actions/checkout/issues/551#issuecomment-1154421608


dblanken-yale (approved):
LGTM; it should gracefully fallback to our repo looking at [checkout@v4's docs](https://github.com/actions/checkout/tree/releases/v4.0.0)

PR #723, opened 2024-08-12
YSP-623: Localist: URL field not getting stored properly, and is not getting updated upon syncing

dblanken-yale:
### Description of work
- Adds `process_row` hook to help fields ending in `_url` to have a protocol if it was missing
- Allows the `field_event_cta` to be overwritten on localist sync changes

### Functional testing steps:
- [ ] Set up localist to use the Anthropology group
- [ ] Do an initial sync
- [ ] Go into Localist and update one of the anthropology event external_url to be 'anthropology.yale.edu'
- [ ] Tell localist to do another sync in the multidev
- [ ] Ensure that the event that was changed not only changed the `field_event_cta`, but that it prepended the protocol to make it 'https://anthropology.yale.edu'

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_localist/migrations/localist_events.yml
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_localist/migrations/localist_events.yml
@@ -214,7 +214,7 @@ destination:
     - field_stream_url
     - field_stream_embed_code
     - field_localist_register_enabled
-
+    - field_event_cta
 migration_dependencies:
   required:
     - localist_experiences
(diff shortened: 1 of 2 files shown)


dblanken-yale:
We tested this functionality as a group in Teams, and verify that it is both appending https:// where appropriate and overwriting the external url correctly.

vinmassaro (approved):
Changes look good and testing was successful. Drupal has a method you could use instead of filter_var if you want: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21UrlHelper.php/function/UrlHelper%3A%3AisValid/11.x

dblanken-yale:
@vinmassaro Thanks for pointing that out; I made a switch to that since it feel closer to Drupal and probably is reused in the parser.

PR #728, opened 2024-08-13
YSP-627: Event dates alignment is off

dblanken-yale:
## [YSP-627: Event dates alignment is off](https://yaleits.atlassian.net/browse/YSP-627)

When originally modifying dates, the vertical offset of each were such that it looked ugly and misaligned.  This was due to the bottom margin gin was giving the wrappers.  SmartDate has new css to address this, so upgrading to take advantage of this and to eliminate a patch we had that is now included.

### Description of work
- Update smart_date module to 1.4.1
- Remove our patch "Warning: Trying to access array offset on value of type null" as it's included in 1.4.1

### Functional testing steps:
- [ ] Edit or add a new event
- [ ] Observe that the date fields are now better aligned
- [ ] If you can think of any other places where a date range exists, ensure that it works there as well.


atiddei (approved):

kara-franco (approved):
Straight as an arrow! Looks great.

dblanken-yale:
Going to wait to merge this one till the Component Library PR (different ticket) is merged and a new release is cut so we include it.

vinmassaro (commented):
The Smart Date module has its own CSS, with some styling overrides already for Gin: https://git.drupalcode.org/project/smart_date/-/blob/4.1.x/css/smart_date.css?ref_type=heads

Just wanted to check if we are adding or overriding something we should not be, and assume the default styling from the module should work?

dblanken-yale:
@vinmassaro Good call; I'll look more to see because at the moment it was taking a margin-bottom of a gin variable, not what was shown there.

dblanken-yale:
Looks like that isn't in [our version](https://git.drupalcode.org/project/smart_date/-/blob/4.0.3/css/smart_date.css?ref_type=tags) though; it only addresses the separator.  I'm going to try an upgrade to smart_date to see how it does.  Another solution might be to just reset `--gin-spacing-m` to `0` inside of smart dates.

dblanken-yale:
@vinmassaro Very good call; the latest version of smart_date does have this and fixes it, and also incorporated our 107.patch fix, so no need for it any longer.  I'm going to mark this as a re-review so we can verify updating smart_date didn't break anything else.  So far it hasn't for me.

vinmassaro (approved):

vinmassaro:
Excellent, I like this fix much better!

PR #739, opened 2024-08-23
YSP-639: Change quotes to soft limit

dblanken-yale:
## [YSP-639: Change quotes to soft limit](https://yaleits.atlassian.net/browse/YSP-639)

While creating the AI website, there were quotes that were just over the hard limit of 425.  Instead of gradually increasing this, it was decided to remove the hard limit, allowing those that wish to still go over to do so.

### Description of work
- Pull Quote Block now have soft limit enabled

### Functional testing steps:
- [ ] Attempt to create a Pull Quote
- [ ] Verify that the help text underneath states that the length is "recommended"
- [ ] Attempt to go over the recommended limit of 425
- [ ] Save the block
- [ ] Ensure that all of your quote is displayed

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.block_content.pull_quote.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.block_content.pull_quote.default.yml
@@ -58,8 +58,8 @@ content:
         hide_guidelines: '1'
       maxlength:
         maxlength_js: 425
-        maxlength_js_label: 'Content limited to @limit characters, remaining: <strong>@remaining</strong>'
-        maxlength_js_enforce: true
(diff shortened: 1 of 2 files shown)


kara-franco (approved):
Looks awesome, excellent work! 

The only design-y thing I saw was:
Do we want the image in the Quote callout to display larger if there is more text? Or should the text wrap? 

This shouldn't hold up the change, but we may want to think about how we want this to look!

Screenshot 2024-08-23 at 1 17 59 PM

dblanken-yale:
@kara-franco I agree with that; I think more work needs to be done to push this through; let's talk more about this.  It looks like the site this was geared toward is no longer using as long of text.

@atiddei Let me know if this ends up being not correct.  :D

dblanken-yale:
Actually spoke too soon; I'll be reverting the callout quote change, but keeping it for the quote itself.

PR #752, opened 2024-09-10
YSP-647: PRESIDENT: Ability to bulk Unpublish

dblanken-yale:
### Description of work
- Adds a bulk unpublish option with content moderation
- Adds a bulk publish option with content moderation

### Functional testing steps:
- [ ] Go to any content list
- [ ] Select multiple items using the check boxes next to them
- [ ] Select "Unpublish content" from the bulk actions menu and submit
- [ ] Ensure no errors exist
- [ ] Try to visit one of those pages as an anonymous user--ensure you cannot
- [ ] Go back into the content list
- [ ] Select the same ones you unpublished
- [ ] Select "Publish content" from the bulk actions menu and submit
- [ ] Ensure no errors exist
- [ ] Try to visit one of the pages again as an anonymous user--ensure you can now

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/system.action.publish_moderated_content.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/system.action.publish_moderated_content.yml
@@ -1,0 +1,12 @@
+uuid: d4a2dbbd-0166-48de-b923-1d2bd9247bf7
+langcode: en
+status: true
+dependencies:
+  module:
+    - node
+    - ys_core
+id: publish_moderated_content
+label: 'Publish content'
+type: node
+plugin: 'entity:publish_moderated_action:node'
+configuration: {  }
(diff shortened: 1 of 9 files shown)


dblanken-yale:
Trying to dig more into this, and I found where it's really bombing which is in a module hook callback that specifically is denying access to the `status` key.  Looking at their blames, it stems back to this issue, however I can't really find yet what the issue was to introduce this.

https://www.drupal.org/project/drupal/issues/2068063

dblanken-yale:
My worry is by bypassing the key check, we're opening ourselves up to something.

dblanken-yale:
Dave-logic: I believe they are hindering this in the module access for status in order to keep the regular "Publish" to not show on UX pages.  It's not 100% spelled out, but that's what I'm getting out of the above issue so far.  If that is the case, I do feel overloading it would be appropriate, and as long as we can ensure that the only reason it would be forbidden would be due to content moderation being enabled, I think we could go forward.  I still want to make more sure of this, but I thought about comparing the reason to determine this since we don't have access to the resultset objects of the access at that point, and I only want this to happen in our custom actions to make sure we're not squelching other portions of the system that need that to be denied.

atiddei (approved):

kara-franco (approved):

PR #755, opened 2024-09-16
YSP-652: Views setting Can or Must selection is not selected on configure

dblanken-yale:
## [YSP-652: Views setting Can or Must selection is not selected on configure](https://yaleits.atlassian.net/browse/YSP-652)

Passing the integer value of the operator resulted in the item not being properly selected in the form view.  This fixes this.

### Description of work
- Stops passing integer value to the form for the Can/Must selection; send string instead

### Functional testing steps:
- [ ] Create a view, select an option for Can/Must contain the tags
- [ ] Save
- [ ] Modify the view
- [ ] Ensure that the correct Can/Must option that you selected is still selected
- [ ] Ensure that any items you include actually show on the view still by modifying content to match your criteria


atiddei (approved):

dblanken-yale:
Reporting in on behalf of Pam; this works for her as well.

kara-franco (approved):

PR #759, opened 2024-09-17
Switch drupal/typogrify patch to local version

vinmassaro:
### Description of work
- Upstream merge request for https://drupal.org/issue/3373921 changed and the patch no longer applies. Switch to a local version of this patch with just our changes from that issue, which applies to typogrify 1.2.


dblanken-yale (approved):
Looks good; not sure why but SMARTYPANTS makes me laugh.

PR #761, opened 2024-09-17
fix(views-text): Change any/all text to not include category

dblanken-yale:
## fix(views-text): Change any/all text to not include category

Because we have muiltiple places where Categories could be used, we don't want to specify that such that it's not misleading that this doesn't apply to the Jakala added filters.

### Description of work
- Changes the text displayed on the "Must include"/"Can include" to remove tags and category reference

### Functional testing steps:
- [ ] Add a view
- [ ] Verify the text has changed


vinmassaro (approved):
Simple text change LGTM

atiddei (approved):

PR #763, opened 2024-09-18
fix(remove-exposed-filter-patch): Remove exposed filter patch

dblanken-yale:
### Description of work
- Removes "Ajax exposed filters not working for multiple instances of the same Views block placed on one page" patch

### Functional testing steps:
- [ ] Add an image block
- [ ] Ensure when selecting media you can properly filter

### Repercussion test
- [ ] Add two views on a page with similar setups
- [ ] After saving, filter the first view
- [ ] Observe that the 2nd view received the filter the first view selected and the first view has a spinner waiting for it to finish (it never will)

Jakala is looking into a fix for this.  Once we have a fix, we can introduce it.


dblanken-yale:
@nplowman @marjina-constantin Wanted to ping you on this so you're aware.

vinmassaro (approved):
Tested locally, LGTM

PR #768, opened 2024-09-23
YSP-346: Add pronoun field

dblanken-yale:
## [YSP-346: Add pronoun field](https://yaleits.atlassian.net/browse/YSP-346)

Other work needing review in:
- [Component Library Twig](https://github.com/yalesites-org/component-library-twig/pull/422)

### Description of work
- Adds pronoun field to profile
- Allows display of this profile field in rendering

### Functional testing steps:
- [ ] Add a profile
- [ ] Observe an input for Pronoun
- [ ] When filling this field out, observe it shows on rendering of the profile

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.node.profile.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.node.profile.default.yml
@@ -15,6 +15,7 @@ dependencies:
     - field.field.node.profile.field_media
     - field.field.node.profile.field_metatags
     - field.field.node.profile.field_position
+    - field.field.node.profile.field_pronouns
     - field.field.node.profile.field_subtitle
     - field.field.node.profile.field_tags
     - field.field.node.profile.field_teaser_media
@@ -107,6 +108,7 @@ third_party_settings:
         - field_first_name
         - field_last_name
         - field_honorific_prefix
+        - field_pronouns
         - field_position
         - field_subtitle
         - field_department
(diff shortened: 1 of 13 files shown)


miketullo95:
first review back! woohoo! this looks great, super excited to get this work through. One note, idk if it was discussed... do we want pronouns to show through view previews? The profile pages look great with the pronouns in the meta block 👍

dblanken-yale:
Woohoo!  Welcome back @miketullo95!!!  Actually no discussion was made yet; I more took this on since I had a little time over the weekend.  Was more to spark conversation on what we should do with it.  I'll see if I can add some meta goodness into that so we can see how it'd look.

miketullo95:
> Woohoo! Welcome back @miketullo95!!! Actually no discussion was made yet; I more took this on since I had a little time over the weekend. Was more to spark conversation on what we should do with it. I'll see if I can add some meta goodness into that so we can see how it'd look.

that would be great! I also don't know if its 100% the correct move (still day 1 back 😅). It could be a setting in the view tool but we may want to not add too much in that tool since it has grown a lot recently.

dblanken-yale:
@miketullo95 Ya we have some discussion and need your expertise on some view guidance in some other areas, and yes, it's about to get larger, so we want to minimize as much as we can I think.

miketullo95:
@dblanken-yale we missed this in our bi-weekly but we should p2 it for tuesday! 😄

dblanken-yale:
@miketullo95 Good call; ya again this was more a jumpstart since we didn't really talk through any specifics.  Just took it with my own idea of it so we could see something and make some decisions.  😁 Let's try to on Tuesday.

miketullo95:
@dblanken-yale definitely agree! I think even a safe bet is to approve/merge with whats in the PR now and if users want more control over this field via views, we can add that as a new feature.

atiddei (approved):

PR #773, opened 2024-09-30
Fix deprecation warning on first load

dblanken-yale:
## Fix deprecation warning on first load

We had a $cachable which was not even in the signature of options_allowed_values, yet we were passing something in that then wasn't even being used.  This lets it match the signature found at:

https://api.drupal.org/api/drupal/core%21modules%21options%21options.module/function/options_allowed_values/10

### Description of work

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...


dblanken-yale:
Made a proper ticket so closing the one.

PR #778, opened 2024-10-07
YSP-697: Image Banner not respecting focal point

dblanken-yale:
## [YSP-697: Image Banner not respecting focal point](https://yaleits.atlassian.net/browse/YSP-697)

_Real work done in :_ [Atomic](https://github.com/yalesites-org/atomic/pull/272)

The div that is inserted for the item seems to be modifying how the image banner behaves, skipping the styling of it to match the grand hero.  By creating this, it allows it to bypass this, picking up the styles similar to the grand hero in Drupal.

### Description of work
- Adds a bypass to the item content so that the extra unsettled div does not get rendered

### Functional testing steps:
- [ ] Create an image banner
- [ ] Create a grand hero banner
- [ ] Observe the placement of the image, the boundaries of it, and make sure they match in image shown
- [ ] In their configuration, modify to the "short" version
- [ ] Observe the placement of the image, the boundaries of it, and make sure they match in image shown

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ The YaleSites platform empowers the Yale community to create digital experiences
 ## Common Scripts
 
 These are the most common npm scripts you may find yourself using:
-(Each is prefixed with `npm run `)
+(Each is prefixed with `npm run`)
 
 ### PR Reviews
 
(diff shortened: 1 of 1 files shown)


miketullo95 (approved):
👍

PR #780, opened 2024-10-07
Revert "YSP-697: Image Banner not respecting focal point"

dblanken-yale:
Reverts yalesites-org/yalesites-project#778


dblanken-yale:
Merged wrong PR

PR #795, opened 2024-11-14
docs(git): document about making atomic a safe directory

dblanken-yale:
## docs(git): speak about making atomic a safe directory

In some cases, vendors were receiving strange errors related to the atomic directory already existing.  It turned out the real issue was that git detected strange ownership of it, and would disallow it from being brought down.

While the first fix is to ensure that yalesites-org/* was marked in composer as prefer-dist in the config.json, the next is to add a safe directory to allow this to work again.

### Description of work
- Adds instructions should users run into a safe directory issue


codechefmarc (changes requested):
The command I had to run was:
`git config --global --add safe.directory /app/web/themes/contrib/atomic`
with the "app" specified because it was inside a container - I don't know how this would be different if not run from within Docker - i.e. a separate AMP stack - but I think changing it to that makes sense because this repo is setup to use Lando.

dblanken-yale:
Gotcha; updated the command so it includes app.  Good that I added you!  haha

vinmassaro (approved):

codechefmarc (approved):
Good to go!

PR #792, opened 2024-11-05
Release v1.10.0

yalesites:
## Pull requests
- https://github.com/yalesites-org/yalesites-project/pull/646
- https://github.com/yalesites-org/yalesites-project/pull/700
- https://github.com/yalesites-org/yalesites-project/pull/772
- https://github.com/yalesites-org/yalesites-project/pull/779
- https://github.com/yalesites-org/yalesites-project/pull/789
- https://github.com/yalesites-org/yalesites-project/pull/791
- https://github.com/yalesites-org/yalesites-project/pull/792
- https://github.com/yalesites-org/yalesites-project/pull/793
- https://github.com/yalesites-org/yalesites-project/pull/796
- https://github.com/yalesites-org/yalesites-project/pull/798

## Changes without a pull request:
- 9cd3d0b8c1eb844aafa0ad35072d41cde1524583

PR #798, opened 2024-11-19
chore: last config export for drupal 10.3 upgrade

vinmassaro:
### Description of work
- Export config again for remaining changes needed by #792


dblanken-yale (approved):
Good call.  LG

PR #803, opened 2024-11-21
Clean up unused modules from composer.json

vinmassaro:
### Description of work
- Removes modules from composer.json that are not enabled in `core.extension.yml`:
  - autosave_form
  - block_content_permissions
  - bugherd
  - publishcontent


dblanken-yale (approved):
All seems to be fine; let's bring this into v1110 and test against the RC sites.

PR #809, opened 2024-11-26
YSP-293: Log out of CAS when logging out of drupal

vinmassaro:
## [YSP-293: Log out of CAS when logging out of drupal](https://yaleits.atlassian.net/browse/YSP-293)

> When logging out of Drupal, if the home page is CAS-protected, the user is automatically logged back in because they still have a valid CAS session. This causes confusion because the redirection happens quickly and it appears the user is never logged out, even though they are.

### Description of work
- Sets config to log out of CAS when logging out of Drupal
- Returns user back to home page after logout instead of CAS Server default (www.yale.edu)

### Functional testing steps:
- [ ] Log into the site with CAS
- [ ] Log out of Drupal
- [ ] Verify you are taken to the home page
- [ ] Log back into the site and verify you are prompted to re-login to CAS (new session since old CAS session was killed with Drupal logout)


dblanken-yale (approved):
Nice; I was worried about the case where the front page was CAS protected, but even then it takes you to log in again, which is great.  🕺 Nice work/find!

PR #816, opened 2024-12-02
fix: update focal_point to 2.1.2 so patch #2830678 still applies

vinmassaro:
### Description of work
- Patch updated in https://www.drupal.org/project/focal_point/issues/2830678 now only applies to focal_point 2.1.2

### Functional testing steps:
- [ ] Verify `composer install` can complete correctly and apply the patch


dblanken-yale (approved):
YOLO :D

PR #828, opened 2024-12-12
fix: run workflow after creating a release to clean up multidevs

vinmassaro:
### Description of work
b382a15 broke the behavior of release_pr.yml running when a release pull request was closed. This change triggers the workflow to run after running release.yml.


dblanken-yale (approved):
LGTM

PR #830, opened 2024-12-13
IYY-266: Updates existing Power BI embeds to work with new Power BI embedder

codechefmarc:
### Description of work
- Adds an update hook to fix existing Power BI embeds to work with new Power BI code

### Functional testing steps:
- [ ] Visit the test site and verify that on the home page no errors exist about Power BI embed
- [ ] Scroll down and verify that the Power BI embed loads


dblanken-yale (approved):
No errors for me locally, even with the same DB.  :D

PR #833, opened 2024-12-19
IYY-285: Site editors can add a caption to Image Banner

joetower:
### Description of work
- Adds image caption field (`field_text`) to the image banner
- Wires the caption
- Component Library: https://github.com/yalesites-org/component-library-twig/pull/450
- Atomic: https://github.com/yalesites-org/atomic/pull/292

### Functional testing steps:
- [ ] Visit test page: https://pr-833-yalesites-platform.pantheonsite.io/page-with-image-banner
- [ ] Verify you see an image caption under the Image Banner


codechefmarc (changes requested):
All of this looks good to me, the only thing I can see is when you are in layout builder, the caption shows up on top of the image. It looks fine once you save, so not sure if you want to make any adjustments:
Screenshot 2024-12-20 at 8 51 01 AM

codechefmarc:
Oh, and this is not part of this ticket, but I also noticed that the image banner is using the "default" view mode for images, which doesn't use a responsive image style and also doesn't work with focal point. I'll check with Duncan to see if we should add to their Jira board.

joetower:
@codechefmarc I pushed up a fix for the caption in layout-builder. It was picking up way too broadly defined `.caption` styles from Drupal. 

I'll look into the image styes being used for the Image Banner. My Lando is refusing to run today, so it might not happen until the new year.

codechefmarc:
I talked with David about it earlier today and they are aware of the image style for the image banner. So, no worries in looking into it, I believe they already have a ticket in place to fix eventually.

miketullo95:
hi everyone! this looks great, only two points from me

- [ ] can we make image caption field optional?
- [ ] can we reduce the padding after the image caption field (see amazing mockup below 😆 )
Screenshot 2025-01-06 at 8 43 38 AM

joetower:
@miketullo95 Thank you!

1. For the caption being required: I didn't realize it was required :cry: . I need to create a new field to use in this instance, instead of re-using the caption field we are already using elsewhere (content image/text-with-image). I'll remove that field and add a new one and let you know when it's ready for another look. 

2. For the spacing, I didn't add more spacing to the caption or to the banner for this ticket, than the spacing that exists on production. I'm happy to talk more about making spacing changes, but for the sake of clarity, what you're seeing is the bottom margin on the Image Banner. Would you like to reduce the bottom margin on the Image Banner component regardless of whether it has a caption or not? Or would you only like to reduce the bottom margin if the image banner has a caption? 

Please see the screenshare below. Thank you!

IYY-caption-spacing

joetower:
@miketullo95 Actually, I was incorrect about the caption field needing to be required. It _can_ be changed to optional and I do not have to add a new field. 😌 

Please let me know how you'd like to address the spacing change whenever you have a chance to take a look and consider it. Thank you!

miketullo95 (approved):
@joetower this looks g2g! based on our chat in our meeting on 1/7, the spacing issue will be its own separate ticket 😄

dblanken-yale:
@joetower Just making sure that the intent was to allow this much inside of the caption?  If so there is a hidden H2 in there that you can't see there.  heh

joetower:
@dblanken-yale Yes, the way we're handling the caption field on the Image component is the same: it uses the same field (this PR re-uses that same field from the image component) and places the responsibility on the content author to make good decisions about what to enter and not enter in a caption field.

dblanken-yale (approved):
@joetower I agree with that--noticed those aren't being referenced anywhere in that styling.  LOL

Thanks for answering those; got my seal of approval then.  :D Nice work!

dblanken-yale:
@joetower Mind placing this one in the IYY branch?

joetower:
@dblanken-yale I update all three to `iyy`. 👍 

Should I go ahead and merge them all?

PR #836, opened 2025-01-02
IYY-299: Show help text about Locks to only those that can lock things

codechefmarc:
### Description of work
- Patches Layout Builder Lock module to prevent layout builder lock help text from showing up if the current user does not have permission to alter locks.

### Functional testing steps:
- [ ] Login to the site as an admin
- [ ] Enter layout builder for an existing page
- [ ] Scroll to the bottom and click "Add section"
- [ ] Choose either the "One column" or "Two Column" layout
- [ ] Verify that there is a message below the administrative label that says "Locks can be configured when the section has been added."

Screenshot 2025-01-02 at 11 37 13 AM

- [ ] Logout of the site or login to the site in an incognito window as an Editor or Contributor user
- [ ] Perform the same process as above
- [ ] Verify that there is no longer a message about the locks as those users do not have access to edit locks

Screenshot 2025-01-02 at 2 17 11 PM


duncancm9:
That works! Thank you

miketullo95 (approved):
👍

PR #850, opened 2025-01-16
YSP-776: Salesforce (FormAssembly) embed source

dblanken-yale:
### Description of work
- Adds FormAssembly embed

### Functional testing steps:
- [ ] Log in through CAS
- [ ] Create a new page
- [ ] Create a new embed block, using the sample in the embed instructions
- [ ] Paste it in
- [ ] Observe it works even if it did not produce a form (since we have no form to show of our own)


dblanken-yale:
@miketullo95 Thinking that right now we have been lucky that logos exist, but wondering if we should have some default image that embeds can use if a logo doesn't exist?  I plan on making one on FA soon, but just thinking ahead.  Like, does LibCal have one?  Should we have a default that can be used?

dblanken-yale:
This is basically ready, but going to keep in WIP till we get a real example.

dblanken-yale:
Closing as the user of this decided to go down a different path.

PR #851, opened 2025-01-16
YSP-797: Add default display mode selection in views tool

dblanken-yale:
## [YSP-797: Add default display mode selection in views tool](https://yaleits.atlassian.net/browse/YSP-797)

In a specific cas of a calendar, this view mode does not exist in other content types in the view tool.  Sometimes this results in nothing being selected for a view mode, making the default view be a real render of the page, causing a recursive render of the page if it is included in the list of items in the view.  This ensures that if you go to a content type that no longer has the selected view mode, it defaults to a card, which we will need to make sure is on every content type.  If this ever changes, we'll need a hash to define that for each content type.

### Description of work
- Adds a check that defaults the view mode if it doesn't exist as a possible view mode for a content type

### Functional testing steps:
- [ ] Create an Event-based calendar view
- [ ] Modify this and change it to a Page-based view (do not select a mode)
- [ ] Save
- [ ] Ensure the page loads and that it defaulted to the card based render for each item in the view


miketullo95:
@dblanken-yale it didnt work 😢. created a calendar, configured that block, just clicked page and hit save and got this: 
Screenshot 2025-01-16 at 4 24 49 PM

dblanken-yale:
Going to do some more work; in addition to this doing differently locally, I'm seeing options disappearing?  Going to try to fix all of that.

miketullo95 (approved):
👍

PR #852, opened 2025-01-16
YSP-758 :: Add taxonomy filter drop downs in Views block

laura-johnson:
### Description of work
- Add Custom Vocab vocabulary
- Add Audience vocabulary
- Add Audience and Custom Vocab fields to all content types
- Adjust the scaffold view to include new exposed filters for these vocabs
- Add new fields to the views block

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...


miketullo95:
This is looking great @laura-johnson ! Here are my notes

- [ ] with the new filter options, we should probably allow the filters/search to expand to two lines in a flexible container. See screenshot with everything turned on.
Screenshot 2025-01-17 at 12 50 07 PM
- [ ] im not able to add terms to the new vocabs (audience and custom)
- [ ] we would like the ability to rename the custom vocab at the taxonomy level, not the view level. Can that be a setting on `/admin/structure/taxonomy/manage/custom_vocab/overview`

laura-johnson:
Thanks @miketullo!
> 
> * [ ]  with the new filter options, we should probably allow the filters/search to expand to two lines in a flexible container. See screenshot with everything turned on.
Agreed, I'll make that adjustment.

> * [ ]  im not able to add terms to the new vocabs (audience and custom)
I think I accidentally deleted the terms in the vocabularies... I'll fix that.

> * [ ]  we would like the ability to rename the custom vocab at the taxonomy level, not the view level. Can that be a setting on `/admin/structure/taxonomy/manage/custom_vocab/overview`
I feel like i'm missing something here... you can already edit the vocabulary name globally at https://pr-852-yalesites-platform.pantheonsite.io/admin/structure/taxonomy/manage/custom_vocab. If we add a new field for the vocabulary name, where do we want that alternate name to appear? If you can tell me how this is going to be used that will probably clarify it for me.

miketullo95:
> I feel like i'm missing something here... you can already edit the vocabulary name globally at https://pr-852-yalesites-platform.pantheonsite.io/admin/structure/taxonomy/manage/custom_vocab. If we add a new field for the vocabulary name, where do we want that alternate name to appear? If you can tell me how this is going to be used that will probably clarify it for me.

@laura-johnson Oh I think I realized what's going on. So I only have platform admin permissions, so I cannot see the setting to rename the vocab name. Could this be available to most user permissions (excluding editors)? 

If thats not it, we can always discuss in a zoom if needed

dblanken-yale:
@laura-johnson Correct me if I'm wrong, but sounds like the editing through taxonomy for a site builder would be problematic.  Would it be possible to create a site setting for this name and on save, modify the specific taxonomy name for the user?  Would that be easier?

dblanken-yale:
It will be unique to the site, but any reference to it on the site will be that new name yes.  So, for instance, yalesites might name it "type", but another instance of YaleSites my name it "external tags".  My hope is we just keep referencing the machine name of it to change the name so it'd be ok.  Or are you thinking about config imports?  I didn't think of that till now.

laura-johnson:
I see, yes I wasn't sure what the purpose of overriding the vocabulary name was, it makes sense now that we want a custom vocabulary per site. I think it should be okay if I add it to the config ignore settings, that will prevent it from being overridden during config import.

dblanken-yale:
Ah gotcha; ya the case was that someone wanted to have a taxonomy that we didn't feel the platform should have.  So we thought if we had a "custom" one that could be renamed, each site could have the ability to have this if they need it, and that could be used in things like the views tool for filtering, etc.  Yep, config ignore should do it.  🕺  Thanks so much for doing this!

ntay90 (approved):

laura-johnson:
@dblanken-yale @miketullo95 I made those updates and also updated the style so that the fields flex onto two lines if there are more than 3 filters. You can update the vocabulary name in Site Settings under Custom Vocabulary Name. Take a look and let me know if that all looks good... 

https://pr-852-yalesites-platform.pantheonsite.io/new-vocabs-test-page

miketullo95:
@laura-johnson this looks great! final testing notes:

- [ ] All filter dropdowns besides Event Category not showing in an event view (see event view here - https://pr-852-yalesites-platform.pantheonsite.io/test-page-mike)
- [ ] Can we removed the custom vocab label from the view form and just pass through the label set in Site Settings on the front end of all views?
- [ ] Is there an easy way we can make the search the only thing on the first row (and span the whole row) if all 3 filters are applied? 
- [ ] When there are two rows of search/filters - could we reduce just the vertical gap between the filter rows in half. Keeping horizontal as 1rem but changing vertical to .5rem

Slightly off topic, but if you try to filter on my test page mentioned in the first bullet, we run into the bug https://yaleits.atlassian.net/browse/YSP-678, that should be coming up next! 😆

laura-johnson:
@miketullo95 I made those updates. Let me know what you think!

miketullo95 (approved):
woohoo! this looks great 👍

laura-johnson:
@dblanken-yale Is it okay if I add a default value of 'Custom Vocab' and then make that field required in the Site Settings form? So if someone edits the form and removes the default value, they'll get a message that it is required. But because there is a default value, they won't be prevented from saving the form otherwise.

dblanken-yale:
@laura-johnson Perfect!  Ya just didn't want to have the possibility of no label.

miketullo95:
@laura-johnson one last thing, could we pass through the custom vocab name if someone fills out the setting on the content type form? for example; if I called the vocab "Resource Type", can the field name on all content types say "Resource Type:" instead of the custom vocab name?

laura-johnson:
> @laura-johnson one last thing, could we pass through the custom vocab name if someone fills out the setting on the content type form? for example; if I called the vocab "Resource Type", can the field name on all content types say "Resource Type:" instead of the custom vocab name?

That would be doable, I'd have to update the name of the vocabulary name if one of those field titles is updated and then alter that field to display the vocabulary name rather than the field title. Does that sound right? Is there any reason they might want different names for that field on different content types? I'm wondering if changing the default field behavior would cause any confusion...

laura-johnson:
@miketullo95 I added the update to the field labels, and added those to config ignore, so they won't be overridden by config import.

miketullo95:
> @miketullo95 I added the update to the field labels, and added those to config ignore, so they won't be overridden by config import.

amazing! this is g2g in my opinion and can go into dev pending a review from @dblanken-yale

dblanken-yale (changes requested):
@laura-johnson This is looking amazing!  There was one piece I think we can remove though; since the audience is going to be a "first class citizen" in the content types, I don't think we need them to rename that in the views tool--just that custom taxonomy.  Would you mind removing that functionality from the views tool.  It could just be `Audience` instead. (Unless @miketullo95 has a preference)

Otherwise so far I can't break it.  :D Great work!

laura-johnson:
> Would you mind removing that functionality from the views tool. It could just be `Audience` instead. 

I'll make this change.

laura-johnson:
@dblanken-yale I made that last change, there is no Audience label option in the views tool.

dblanken-yale (approved):
Awesome job!  This works so well!

PR #857, opened 2025-01-29
YSP-760: Editing an image popup doesn't show save button

dblanken-yale:
## [YSP-760: Editing an image popup doesn't show save button](https://yaleits.atlassian.net/browse/YSP-760)

This displays the save button that is hidden by GinLb.  This makes me wonder the implications or the reason why they chose to do this across all dialogs, but it does work.  My real worry here is that users might not realize when they change the image that it will change the image for that media item--so they're not choosing a new image, they're replacing the current image across the site.

It also does not update the sidebar since it doesn't refresh the page on save.

### Description of work
- Shows the Save button for the media edit within a block edit dialog

### Functional testing steps:
- [ ] Create a new page
- [ ] Add a block that has an image and save it
- [ ] Edit the block
- [ ] Click the pencil icon
- [ ] Verify that there is a `Save` button displayed
- [ ] Make changes to the image
- [ ] Save
- [ ] Save the block
- [ ] Verify you see your changes


miketullo95 (approved):
can confirm i see the save button, and actions i take in that modal (changing focal point, alt text, tags, etc) work after saving

PR #869, opened 2025-02-08
IYY-271: HOTFIX Fixes ambiguous nid error on some views

codechefmarc:
### Description of work
- Fixes an issue where with some views, there was a white screen error relating to an ambiguous nid in the query
- Please test with `ys-library-yale-edu`, @dblanken-yale 

### Functional testing steps:
- [ ] Load the previously troublesome view
- [ ] Verify that there are no longer errors
- [ ] Test the functionality of the original feature:
- [ ] Create an event or edit an existing event
- [ ] Add a view of type events
- [ ] Save the block
- [ ] Verify that the current event is not in the view
- [ ] Re-edit the view block and select "Include this content in view"
- [ ] Save the block
- [ ] Verify that the current event is now in the view


dblanken-yale (approved):
This fixes the issue.  Thanks for doing this!

PR #871, opened 2025-02-10
IYY-279: Visitors can listen to Posts - aka Audio player component

joetower:
### Description of work
- Adds Audio component
- Themes custom player
- Adds Audio component block configuration
- Component Library: https://github.com/yalesites-org/component-library-twig/pull/468
- Atomic: https://github.com/yalesites-org/atomic/pull/307

### Functional testing steps:
- [x] Navigate to: https://pr-871-yalesites-platform.pantheonsite.io/posts/2025-02-10-post-with-audio-option
- [x] Verify you can play the audio
- [x] Verify you can change the playback speed of the audio file
- [x] Verify you can increase or decrease the audio volume
- [x] Verify you can click the timeline to jump to a specific spot in the file
- [ ] Navigate to: https://pr-871-yalesites-platform.pantheonsite.io/posts/2025-02-10-another-post-with-audio
- [ ] Verify you can play the audio file
- [ ] Test responsiveness

https://github.com/user-attachments/assets/05f5f3f5-f3fe-464f-afdc-cd9177c32c8c

---
- [ ] Add a new Audio component to a post
- [ ] Verify it works as expected

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/block_content.type.audio.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/block_content.type.audio.yml
@@ -28,6 +28,6 @@ third_party_settings:
         url: ''
         external: ''
 id: audio
-label: Audio
+label: Audio Player
 revision: false
 description: null


codechefmarc (changes requested):
Overall this looks great and works perfectly. I noted the lack of the progress line in component library. Here are some other questions and comments:
* Audio file title - should this be changed to simply "Title"? @miketullo95 question
* Do we want to allow the listed HTML tags on the audio file title? I seem to recall we were hiding those (see teaser text as example).
Screenshot 2025-02-10 at 4 25 07 PM
* When I initially add a new audio file, I don't see the Audio file title field, but I do when I go to edit an existing one:
Screenshot 2025-02-10 at 4 22 26 PM
* No image for layout builder browser, @miketullo95  I assume will create one?
* I saw an accessibility warning:
Screenshot 2025-02-10 at 4 24 03 PM

myvaughn:
We will want @ntay90 to look at this widget. I notice when using the keyboard activating the play button causes the focus to be lost (it should remain on the play button). Also, although not necessarily just an accessibility issue, the volume control design is off for me in Chrome, there is no line for the slider (attached screenshot). It took me a while to figure out that what I thought was a random circle floating to the right of the speaker icon was the volume slider. I also thought there was something wrong with the mute button--I expected to be able to activate the speaker icon to mute the audio. 
image

myvaughn:
The slider line doesn't appear for me for the playback control either. I tested in Safari too and the same thing: 
image

joetower:
@myvaughn The player's appearance should now be fixed in Chrome and Safari. I learned you can not selector-stack pseudo shadow-dom elements for Firefox and Webkit contexts. They each need their own separate entry. 

The pause button will receive focus now, when keyboard navigating and after pressing enter on the play button.

The volume icon is now wrapped in a button which, when clicked, will mute the video and the icon changes. If clicked again, the icon changes and the volume is reset to 50%.

---

@codechefmarc I replaced the title field with a plain text version. Your feedback has been addressed (minus the assets we need from Mike T) please give this another look. 

I found that accessibility warning interesting, but I'm wondering if we need to add an aria attribute (`aria-describedby`?). I'll look forward to a11y feedback.

codechefmarc:
Re-tested and looks good to me!

codechefmarc:
@miketullo95 - Preview image has been added and I switched the label so it comes from the block and not the media file. Since this is a new component, I didn't update any existing test audio blocks, but if you add a new one, it will have a default value of "Listen to this content".

duncancm9:
Preview image and label override look good!

duncancm9 (approved):

web/profiles/custom/yalesites_profile/config/sync/block_content.type.audio.yml
```diff
@@ -0,0 +1,33 @@
...
+        destination: default
+        url: ''
+        external: ''
+      anonymous:
+        active: 0
+        destination: default
+        url: ''
+        external: ''
+id: audio
+label: Audio
```
duncancm9, line 31:
```suggestion
label: Audio Player
```

duncancm9:
@miketullo95 this should be good to go for a final review. thank you

ntay90 (changes requested):
This looks pretty good accessibility-wise, though there are a few things that need adjusting/adding to:

- [ ] The playback speed button should indicate which speed is currently selected to screen reader users. This could be accomplished using an `aria-describedby` attribute pointing to the `id` of current speed.
- [ ] In a similar vein, once the list of playback speeds is expanded, the button with the currently selected speed will need an `aria-pressed="true"` attribute.
- [ ] The slider will need some additional attributes to be clear to screen reader users. Since the slider is advancing in whole seconds, set the `step` on the `<input>` to `1`, with the `max` set to the max amount of seconds, and `min` set to `0`. If possible, add an `aria-valuetext` attribute that updates with the appropriate value + text, such as `aria-valuetext="1 minute 25 seconds elapsed"` or `aria-valuetext="85 seconds elapsed` depending on difficulty. This `aria-valuetext` attribute must update in conjunction with the slider's value.
- [ ] There is a visually hidden `<label>` adjacent to the slider that is not associated with the slider and is redundant for screen reader users. It should be removed, or an `aria-hidden='true"` added to it.
- [ ] More usability than accessibility, but making the entire speed select area clickable would be helpful, in addition to a hover state. It was a bit unclear where I was supposed to click to expand it, and it may be difficult for a dictation user to expand.

PR #880, opened 2025-02-13
YSP-678 Multiple views exposed filters on a page only work on last instance

laura-johnson:
### Description of work
- Create a new version of the multiple views exposed filters patch that preserves the original form IDs for media library forms: https://www.drupal.org/files/issues/2025-02-13/3163299-126-ajax-exposed-filters-not-media-library.patch

From this issue: https://www.drupal.org/project/drupal/issues/3163299

### Functional testing steps:
- [ ] Add multiple views to a page, verify filters to every view and do not affect others on the same page
- [ ] Add a media block, verify the media library exposed filters work

I have set up a testing page [here](https://pr-880-yalesites-platform.pantheonsite.io/pam-test-page) with multiple views.

Note that my approach with the patch is to go with the solution from comment 120 [here](https://www.drupal.org/project/drupal/issues/3163299#comment-15829959), but I adjusted it so that it does not affect media library forms. It seems like there may be a more graceful way of not including forms which are being added in Layout Builder, but I didn't see one. I didn't want to add any gin_lb classes to a core patch. I thought it better to exclude media library forms because media library is in core.


dblanken-yale:
@laura-johnson I absolutely love that these patches are working!  Nice find!!!!!!!

dblanken-yale (approved):
Approved; we might want this moved to another target branch to bring into the release.  @miketullo95 If so, I'll merge it into my feedback one when we're ready.

miketullo95 (approved):
woohoo!

miketullo95:
> Approved; we might want this moved to another target branch to bring into the release. @miketullo95 If so, I'll merge it into my feedback one when we're ready.

yeah merging to the feedback PR would be great

PR #883, opened 2025-02-14
v1.13.0 Feedback Fixes 2

dblanken-yale:
### Description of work
- Pinned events now show first
- Updated to use newest CL/atomic release

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...


nJim (approved):
I tried to find a more surgical solution but this project has grown to include so many different types of nested entities that getting the cache context correct has been a real challenge. This gets the job done and fixes other potential problems. It’s just what you need right now.

miketullo95 (approved):

PR #888, opened 2025-02-20
YSP-826: Hotfix: clear render cache for view items

dblanken-yale:
### Description of work
- Added `clearViewItemRenderCache` method to clear the render cache for view items to ensure options are respected.
- Refactored `getView` method to use `view_args` array for setting arguments.
- Removed redundant case for "count" in `getView` method since it's not used.

### Functional testing steps:
- [ ] Create a new page with a new events view, setting the following options on the view:
  - [ ] Select `Events`
  - [ ] Hide `Add to Calendar` _on_
  - [ ] Limit the number of items per page to _1_ (for pagination)
- [ ] Note after saving that `Add to Calendar` does not show, even when traversing pagination
- [ ] Create a new page with a new events view, setting the following options on the view:
  - [ ] Select `Events`
  - [ ] `Hide Add to Calendar` _off_
  - [ ] Limit the number of items per page to _1_ (for pagination)
- [ ] Note that `Add to Calendar` shows, even when traversing pagination
- [ ] Visit the original create page with `Add to Calendar `already set to hide
- [ ] Verify that the view displays items without the `Add to Calendar` button still
- [ ] Test with non-event views as well to ensure all is well (you can use other options such as `show category`)

### Alternative test
- [ ] Visit the already configured versions:
  - [ ] [Test Page A - Teaser off, add to calendar on](https://pr-888-yalesites-platform.pantheonsite.io/view-test-page-a)
  - [ ] [Test Page B - Teaser on, add to calendar off](https://pr-888-yalesites-platform.pantheonsite.io/view-test-page-b)
- [ ] Paginate and verify it behaves well in both

### NOTE

We will want to test this on a large site with many events to see the performance hit this may have.  We will be copying a larger site's DB to this multidev when smaller tests have passed.

Devel generate was used to generate 2900+ events on this multidev for stress tests.

### Developer Notes

This was quite a hard issue; we needed to be able to execute the view twice, but we could not clone the existing view, and once a view is executed, it cannot be executed again.  Also, cloning resulted in subobject not being created correctly, and pagination on the existing view vanished.  We believe somewhere there might be reference pointers using the same items since we did not want pagination on the `get all` version.  We were able to pull the view again, then repopulate it with the already executed view, which then has the options we need.  The implication here is that this must run before each view, causing a lot of database deletes, followed immediately by insertions.

### Special thanks
Want to thank @codechefmarc for his time working on this with me.  This was a hard one and I don't think I would have come up with a solution so fast without his help!


dblanken-yale:
Closing for now as this doesn't perform well for large view numbers.  We will keep this in mind as we try to re-imaging the views block.

codechefmarc:
Just noting that there may be a way to use `hook_views_pre_view` and then inside that, determine if it's an ajax request and do the same thing we're doing up there, but only on the next subset of items. So, when you load the page, it'll clear cache for page 2, but not page 3. When the ajax is requested for page 2, it'll clear the cache for page 3 and so on. This is from ChatGPT so no promises it works at all, but wanted to at least note I looked into it a bit and worth trying on for size:

```
function mymodule_views_pre_view(ViewExecutable $view, &$display_id, array &$args) {
  // Only target specific views.
  if (!in_array($view->id(), ['views_basic_scaffold', 'views_basic_scaffold_events'])) {
    return;
  }

  // Detect if this is an AJAX request.
  $is_ajax = \Drupal::request()->isXmlHttpRequest();
  if (!$is_ajax) {
    return;
  }

  // Get the current page and items per page.
  $current_page = $view->pager->getCurrentPage();
  $items_per_page = $view->getItemsPerPage();

  // Determine the next page dynamically.
  $next_page = $current_page + 1;

  // Create a new instance of the view to fetch the next set of results.
  $view_next = \Drupal\views\Views::getView($view->id());
  if ($view_next) {
    $view_next->setDisplay($view->current_display);
    $view_next->setItemsPerPage($items_per_page);
    $view_next->setCurrentPage($next_page); // Set it to the next page
    $view_next->args = $view->args; // Copy contextual arguments
    $view_next->exposed_data = $view->exposed_data; // Copy exposed filters
    $view_next->execute();

    // Collect only the node IDs that will appear in the next page.
    $node_ids = [];
    foreach ($view_next->result as $row) {
      if (isset($row->_entity) && $row->_entity instanceof \Drupal\node\NodeInterface) {
        $node_ids[] = 'node:' . $row->_entity->id();
      }
    }

    // Invalidate cache **only for the nodes appearing on the next page**.
    if (!empty($node_ids)) {
      \Drupal::service('cache_tags.invalidator')->invalidateTags($node_ids);
    }

    // Debugging: Log the invalidated nodes (for testing).
    \Drupal::logger('mymodule')->debug('Invalidated nodes for next page: ' . implode(', ', $node_ids));
  }
}```

dblanken-yale:
Oooo let me give this a go; ya if this is a way to pre-process before the Ajax call I'm all for it for now.

dblanken-yale:
@codechefmarc

PR #897, opened 2025-02-25
Hotfix: v1130: Remove pass by reference on $display_id

dblanken-yale:
### Description of work
- Removes wrong function signature definition so that installs don't complain

### Functional testing steps:
- [ ] If this builds, it is successful


codechefmarc (approved):
Tested this locally with all caching turned on, fixes the error.

PR #905, opened 2025-03-03
YSP-824 Admin Label automatically is the block name

laura-johnson:
### Description of work
- Gets the admin label from the plugin definition or falls back to the plugin ID and adds it as the default admin label value. This is applied to layout builder block add/update forms.

### Functional testing steps:
- [ ] Add a LB block to any node layout. Verify there is a default admin label (for the text block it'll be "Text", etc.)


dblanken-yale (approved):
Sounds small, but I bet a LOT of people are going to love not having to worry about this.

Nice work

PR #907, opened 2025-03-05
LIB: Overlay PNG

Studio-Tom:
## [LIB: Overlay PNG]

Other work done in Atomic and CL

### Description of work
- Adds functionality for image insertion in Grand Hero block overlay
- Hides header text following the image

### Functional testing steps:
- [ ] Confirm Grand Hero field is present for Overlay PNG
- [ ] Configure the Grand Hero block on a page and add a PNG to that field and save
- [ ] Check for proper image display (visible and not extending beyond container)
- [ ] Confirm any header text is hidden


miketullo95:
@Studio-Tom two changes and this would be good to go:

For the authoring experience, can we add a drop down labeled `Heading Type` with options `text` (default) and `image`. Once the user selects image, conditionally show the media upload for `Banner Overlay PNG`

Second, could we add some image size restrictions? It looks great with your example, but we are allowed to upload anything which leads to some not so great examples. Use something like the `School Logo` under content > footer settings > mega footer.

dblanken-yale (changes requested):
We spoke about it but wanted to get it in here so we have some paper trails 😁 

To simplify the logic a lot, can you:

1. Remove the heading default text
2. Remove the conditional `text/image` select
3. Create a new checkbox underneath the Heading level field at the bottom for `Replace heading with image`
4. On check, allow that PNG media select to appear below that checkbox instead of with the Heading text field

I think that'll simplify a lot and be something Drupal can better handle the JS for without the extra.

miketullo95 (approved):

PR #911, opened 2025-03-12
IYY-343 :: Views tool: Allow to filter taxonomy on the new custom vocab

laura-johnson:
### Description of work

When we show Categories, we get the option to filter Included terms.Can we add this same functionality of the option to filter included terms on the Custom Vocab?

### Functional testing steps:
- [ ] Add terms with sub-terms to Custom Vocab
- [ ] In the views tool, use the Custom Vocab Filter - Included Terms field to allow only the subterms of one of the parent terms to be selected. Ensure some content is tagged with the subterms.
- [ ] Verify that only the subterms from that parent term are available in the exposed filter and that it works to filter content tagged with those subterms.
- [ ] Verify that if the Custom Vocab is renamed, the new label is used in the field.

custom-vocab-term-filter


laura-johnson:
@miketullo95 @dblanken-yale I added the term selection for Custom Vocab the same way that Category has it. But I think that either I'm missing something or the category term selection isn't working properly right now... 
1. As editor I can only choose one term to include (it seems like I should be able to choose more than one)
2. When I publish the view, I'm not able to select any terms (as a user).

I have a pages view here: https://pr-911-yalesites-platform.pantheonsite.io/example-landing-page

I've tried selecting terms for both Category and Custom Vocab, they are working the same way but it doesn't seem right. I get the same result locally with just Category, and no code updates.

I branched off of master.

If you can show me an example where it's working as expected, I think I'll be able to figure out what's happening here. Thanks!

duncancm9:
This is working for me! 

cc: @miketullo95  Just flagging a bug that exists here as well as in the Page Category, but I don't think should block this from going in.

If you have a piece of content tagged with two sub categories, and you filter to both of those categories, you see the piece of content twice:

CleanShot 2025-03-25 at 15 21 42@2x

duncancm9 (approved):

miketullo95:
This looks great! I'm going to make a follow-up bug ticket for what @duncancm9 discovered above. 

One thing before I approve: @laura-johnson could you change the label to this new filter + the same filter for the custom drop down to "Filter by [taxonomy vocab name] Parent Term".  Ex: "Filter by Category Parent Term" and "Filter by [Custom vocab] Parent Term". 

Additionally, can you add help text "Select a parent term to show content tagged with that terms sub-items. This ignores content tagged as the parent term and any other parent terms in the vocabulary."

laura-johnson:
I made that update, but not sure how to trigger the CI again to get it up to the multidev!

miketullo95 (approved):

dblanken-yale (changes requested):
It looks like there may need to be more items represented in the form selector array.  I think I'm seeing inconsistencies in the toggle of it in a new block modal vs the sidebar display.  It's a little of a headache that array, so if you want to just make one for each case you see I think it'll work.  In the past, I think it was `_array`, `_selector`, `_ajax`, and one without anything (double check me).

So case in point: Add a new page view block (or just changing content types?), but do not select to show the custom vocab.  Then go edit it and it shows the filter drop down.  But when you check to include the custom vocab, it toggles it away.

laura-johnson:
@dblanken-yale I tried adding more items to the form selector array as you suggested... nothing worked for me. Then I found the issue: https://github.com/yalesites-org/yalesites-project/pull/911/commits/b2abde1384312ba5ea96fb3f020a996a3b950a63

I had added those fields to be replaced by ajax when the entity type changed, but when that happens they don't get the correct state from the feature selection checkboxes. Removing that seems to fix the issue.

dblanken-yale (approved):

PR #913, opened 2025-03-17
YSP-840: Add Event Integration Sync Button for All Users

dblanken-yale:
### Description of work
- Adds a new Integrations landing page
  - Displays each integration with configuration if you have access to
  - If enabled, displays the sync button to sync now

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...


dblanken-yale:
Going with a module-based approach.  This will also allow us to later have a place to put an integration framework so that we have consistency/helpers for new integrations for consistency and conformity.

PR #919, opened 2025-03-21
feat: ysp-738 adding comment to trigger ci

laura-johnson:
### Description of work

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
@@ -659,3 +659,5 @@ function _ys_core_disable_event_fields(&$form, $form_id) {
     }
   }
 }
+
+// Adding this to create a multidev.


dblanken-yale:
This is working in the multidev, however I think we only ever experienced this on a live site, which made this very hard to troubleshoot.  @vinmassaro Any ideas on how we could get this on a dummy live instance; think it'll be harder due to needing the CL update, too.  We could cut a release, but we may have to cut another to remove it if it doesn't work?

dblanken-yale (approved):
This is working, even on a live site: https://live-ys-menu-debug-yale-edu.pantheonsite.io

Nice work @laura-johnson!!!!!!!!

dblanken-yale:
I added don't merge to this one since it was just for the multidev.  Atomic and CL should be good to go with Mike's approvals.

miketullo95 (approved):

miketullo95:
> I added don't merge to this one since it was just for the multidev. Atomic and CL should be good to go with Mike's approvals.

readded the dont merge right after reading this LOL

PR #924, opened 2025-04-07
YSP-417 :: Move from Google Analytics to Google Tag Manager module

laura-johnson:
### Description of work
- Adds and configures Google Tag
- Adds permissions for Site Admin and Platform Admin to administer Google Tag
- Adds a menu item in the Settings menu for Google Tag administration
- Adds Google Tag to config ignore

### Functional testing steps:
- [ ] As Site Admin or Platform Admin, verify you see Google Tag Settings in the Settings drop-down menu, verify you can edit the settings

**Note:** The Google Tag module should automatically migrate Google Analytics settings to Google Tag (this happens when Google Tag is enabled). So I think for now we should leave Google Analytics enabled, and when Google Tag is deployed it should be migrated on live sites. Then we can disable and remove Google Analytics.

### DIFF
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -52,6 +52,7 @@
     "drupal/gin_lb": "1.0.0-rc8",
     "drupal/gin_moderation_sidebar": "1.0.0",
     "drupal/google_analytics": "4.0.2",
+    "drupal/google_tag": "^2.0",
     "drupal/hide_revision_field": "2.5.0",
     "drupal/honeypot": "2.1.4",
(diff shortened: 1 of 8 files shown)


vinmassaro:
@laura-johnson all of this work looks great, except the migration from Google Analytics is not happening if I add an analytics ID prior to switching to your branch and running `drush deploy`. I can see the migration code does run from `google_tag_install`, but a Google Tag ID is not added to the module config. If I uninstall and then re-enable the google_tag module, it then actually saves, so I wonder if there is a chicken/egg issue here.

laura-johnson:
Thanks @vinmassaro, I'll do some testing and take a look at this.

PR #933, opened 2025-04-16
fix: specify path to CA PEM cert to prevent auth errors

vinmassaro:
### Description of work
- Sets the "Custom Certificate Authority PEM Certificate" path so the CAS module does not need to autodiscover it, preventing authentication errors.


dblanken-yale (approved):
Yay it's working!  Nice find!

PR #937, opened 2025-04-17
YSP-803: Reduce Excessive Padding Between Breadcrumbs and Banner Blocks

dblanken-yale:
### Description of work
- Adjusts padding from banner blocks, breadcrumbs, and titles to be more consistent and tighter
- Treated the first element of a 70/30 like we are single content where it would not have padding

### Functional testing steps:
- [ ] Make a page, post, event, and profile
- [ ] In each try different combinations of page title settings, breadcrumb setting, and content collection setting
- [ ] Ensure that the spacing seems consistent and makes sense


miketullo95:
@dblanken-yale i've thought about this more... I think this my final decision for this lol.

When we have breadcrumbs, zero marign top for breadcrumbs. See example: 
Screenshot 2025-04-21 at 9 32 05 AM

When there are no breadcrumbs, add 10px margin top for page title. See example:
Screenshot 2025-04-21 at 9 33 07 AM

dblanken-yale:
@miketullo95 Give it a go now.  I'm going to rebuild this one more time to make sure it has the latest and greatest.

miketullo95 (approved):

PR #940, opened 2025-04-21
YSP-351 :: Allow for more affiliations in Profiles

laura-johnson:
### Description of work
- Allow unlimited cardinality in the affiliations field
- Make it a chosen field

### Functional testing steps:
- [ ] Go to a profile node, verify you can add multiple affiliations

Note: I checked the full profile node and all of the view displays available for profile and I don't see the affiliations displaying anywhere currently. I'm not sure if there is anywhere else that needs to be checked as far as display/functionality.

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Plugin/Block/PostMetaBlock.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Plugin/Block/PostMetaBlock.php
@@ -98,7 +98,7 @@ class PostMetaBlock extends BlockBase implements ContainerFactoryPluginInterface
 
     /** @var \Drupal\node\NodeInterface $node */
     $node = $this->requestStack->getCurrentRequest()->attributes->get('node');
-    if (!($node instanceof NodeInterface || $node->bundle() !== 'post')) {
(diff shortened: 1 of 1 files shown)


laura-johnson:
Hmm, I'm seeing an error on edit layout... investigating.

laura-johnson:
@dblanken-yale I was able to trigger a bug in the view component, but I'm not sure how I did it. I can't reproduce it now. I believe what happened is that I had a type selected (profile), but no display was selected (ie grid etc). And this caused the wrong block to be called, which caused a fatal error. I pushed up a fix before I copied the error unfortunately but the error was something like:

Call to member function first() on null on line 115 of /web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Plugin/Block/PostMetaBlock.php

In that block: $node->field_author->first() was being called on a non-post node.

I was getting it when I tried to access node/[nid]/layout

So I added a check in the [Type]MetaBlocks to check the bundle, which prevents the fatal error, although it doesn't fix whatever caused it. I don't think it's related to my update to the affiliations field.

dblanken-yale:
@laura-johnson Thank you for finding that!  No I don't think it's related to the affiliations also.  Ooo that's a big issue as it should default to the first if none are selected; it should always be there.  I'll take another look.

miketullo95 (approved):

PR #945, opened 2025-04-22
Fix CAS login under Lando

vinmassaro:
### Description of work
When setting the CAS cert path in #933 to prevent auth errors in Pantheon environments, CAS logins broke under Lando. This PR overrides the CAS cert settings when running locally.


dblanken-yale (approved):
Settings work even if the UI doesn't show them (not a YaleSite issue).  Approved!

PR #955, opened 2025-04-25
YSP-895: Accessibility: Bypass block- Content collection -- MULTIDEV ONLY

dblanken-yale:
### Description of work
- Work actually done in [Atomic](https://github.com/yalesites-org/atomic/pull/348)

### Functional testing steps:
- [ ] See atomic steps as this is only for the multidev


dblanken-yale:
Just marking for approval here so we have label tracking since atomic doesn't have them all.

ntay90 (approved):

PR #956, opened 2025-04-25
YSP-892: Design treatment for active Content Collection parent menu items and location move

dblanken-yale:
### Description of work
- A mock of the look and feel if we were to move the content collection above the title near the breadcrumbs
- Work done in:
  - [ ] [Atomic](https://github.com/yalesites-org/atomic/pull/349)
  - [ ] [Component Library Twig](https://github.com/yalesites-org/component-library-twig/pull/510)

### Functional testing steps:
- [ ] In the multidev
  - [ ] Create a nested content collection item
  - [ ] Navigate to it
  - [ ] See the new location above the title
  - [ ] Ensure that the top level parent in the menu is decorated to show that inside of it is an active link
  - [ ] Create a top level only item (no sub items)
  - [ ] Visit it
  - [ ] Make sure that it is also decorated to show you that link is active


miketullo95:
@dblanken-yale first piece of feedback I dont want to lose before the weekend. 

- [ ] could we dynamically add 10px of padding on the page title only if there is a content collection nav present? (I dont want to mess up the padding change we did with the breadcrumb change recently).
Example:
Screenshot 2025-04-25 at 4 29 45 PM

miketullo95:
@dblanken-yale showed this to IYY. this is GTG. we can move foward with the node title move, active link treatement, and the alignment fix here 🎉

miketullo95 (approved):

PR #965, opened 2025-05-01
YSP-899: PostMetaBlock throwing first() error with field_authors

dblanken-yale:
### Description of work
- Add default value checks for view mode in Views block

### Functional testing steps:
- [x] Create a view, specifying `Event` and `Calendar` as the type and mode
- [x] Save
- [x] Edit the block, and select anything other than `Event`
- [x] Ensure that on save, even though it may not show as `Selected` that `Card` view is rendered
- [x] Edit the block again
- [x] Ensure that the first mode is selected now
- [x] Select `Event` and `Calendar` again
- [x] Save
- [x] Edit the block yet again
- [x] Move to other content types; ensure that only in the first reload the first item is not selected

Note that the last item above is another bug where even though the card is selected, the changing of the DOM makes it not selected.  Selections afterward do show, which is why after the first traversal away from the Calendar, they do show.


miketullo95 (approved):
definitely fixed. but it highlights the calendar -> post then not showing the post view options bug big here... but its not related to this fix.

laura-johnson:
Nice!

PR #985, opened 2025-05-15
Hotfix: YSP-951: Breadcrumbs not showing more than 2 trails levels -- For Multidev Only

dblanken-yale:
### Description of work
- Real work done in
  - [Component Library Twig](https://github.com/yalesites-org/component-library-twig/pull/524)
  - [Atomic](https://github.com/yalesites-org/atomic/pull/361)

### Functional testing steps:
- [ ] Create a multi-level content collection (3+ deep)
- [ ] Verify that the breadcrumbs completely show the breadcrumb path when you visit one of those deeply nested items


miketullo95:
reviewed in both atmoic and CL. Checked both CC nav and main site nav and works as inteneded

PR #987, opened 2025-05-22
Hotfix: Bundle error guard, views text changes

dblanken-yale:
### Description of work
- Guards against a NULL node for bundle checks
- Changed views `News by Year` to `View by Year`
- Changed pinned label text to convey it's purpose
- Gave more view search context of what is searched to the user

### Functional testing steps:
- [x] Ensure you can visit events and they render
- [x] Create a new page and create a view
- [x] Ensure that the search checkbox now conveys that it only searches titles
- [x] Ensure that the pinned to top toggle conveys it is not about sorting, but about the display of already pinned items
- [x] Check to display Year filters
- [x] On save, ensure that the Year filter name is now `View by Year` instead of `News by Year`


miketullo95 (approved):
looks good to me!

PR #988, opened 2025-05-28
YSP-966: Drag and Drop is not saving

dblanken-yale:
### Description of work
- Refactored the two-column layout template to mimic the one column version.  This fixes drag and drop in that it provides the move url to use when dragging and dropping.

### Functional testing steps:
- [x] [Log in via CAS](https://pr-988-yalesites-platform.pantheonsite.io/cas)
- [x] [Create a new page](https://pr-988-yalesites-platform.pantheonsite.io/node/add/page)
- [x] Create a new text block inside the existing one column layout
- [x] Create a 70/30 layout
- [x] In each, add a text block
- [x] Drag the text block from the one column layout to the 70/30 (either side)
- [x] Save
- [x] Edit
- [x] Ensure that it was successfully moved and saved to the 70/30
- [x] Try all other layouts with any other blocks and move them

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/layouts/two_column/layout--two-column.html.twig
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/layouts/two_column/layout--two-column.html.twig
@@ -14,15 +14,10 @@
 
 {% set layout__base_class = layout__base_class|default('yds-two-column') %}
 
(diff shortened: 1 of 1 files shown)


dblanken-yale:
Keep in mind this is only fixing the drag and drop, not the other issue where the drag cannot start well within a 70/30 from one side to the other.  That will be in a separate ticket.

miketullo95 (approved):

PR #990, opened 2025-05-29
YSP-638: Header/Footer pencil icon appears on right side instead of left

dblanken-yale:
### Description of work
- Real work done in [Atomic](https://github.com/yalesites-org/atomic/pull/365)


miketullo95:
@dblanken-yale we noticed we can do this to the pre-built form too ☹️

dblanken-yale:
I think we can fix that one. I'll try it tomorrow.

PR #996, opened 2025-06-05
YSP-884 :: Site Setting - Create Font Pairing Option with Mallory as Heading Option

laura-johnson:
### Description of work
- Adds add setting to site settings
- Adds data-font-pairing attribute to body tag

### Functional testing steps:
- [ ] Verify option exists in site settings to select either YaleNew or Mallory
- [ ] Verify the form input/options meet the requirements
- [ ] Verify a data-font-pairing attribute appears in the body tag for all pages, this will be used for heading font and styles

Note there is FE work needed also which will follow.


robherba:
Hi @laura-johnson, the adjustments have been made. Below is a summary of the changes:
-  Removed the font pairing logic from the site header.
-  Set YaleNew as the default font when the font-pairing attribute is missing.

miketullo95 (approved):

PR #997, opened 2025-06-05
YSP-904: YPS: Add Image Background to Action Banner and Callout Blocks

dblanken-yale:
### Description of work
- Adds `field_overlay_background_image` to the CTA Banner block content type
- Adds `field_overlay_background_image` to the Callout block content type  
- Updates form and view displays to include the new field for both block types
- Configures the field to reference media images using the media library
- Makes the `field_link` field optional for CTA Banner blocks (no longer required)
- Adjusts field weights for improved form layout on both content types
- Ensures proper dependencies and display settings for the new overlay background image field
- Other work completed in:
  - [Atomic](https://github.com/yalesites-org/atomic/pull/367)
  - [Component Library Twig](https://github.com/yalesites-org/component-library-twig/pull/529)

### Functional testing steps:
- [x] Verify CTA Banner blocks can have an overlay background image field added
- [x] Verify Callout blocks can have an overlay background image field added  
- [ ] Test that the media library integration works for selecting background images
- [x] Confirm the overlay background image displays correctly on both block types
- [ ] Verify that the CTA Banner link field is now optional (can be left empty)
- [x] Test form layout and field ordering to ensure proper weight configuration
- [x] Confirm existing CTA Banner and Callout blocks continue to work without issues
- [ ] Test permissions and access for different user roles when using the new field

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.block_content.callout.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.block_content.callout.default.yml
@@ -6,10 +6,12 @@ dependencies:
     - block_content.type.callout
     - field.field.block_content.callout.field_callout
     - field.field.block_content.callout.field_instructions
+    - field.field.block_content.callout.field_overlay_background_image
     - field.field.block_content.callout.field_style_alignment
     - field.field.block_content.callout.field_style_color
   module:
     - markup
+    - media_library
     - paragraphs
     - paragraphs_features
 id: block_content.callout.default
@@ -46,6 +48,13 @@ content:
     weight: 0
     region: content
     settings: {  }
+    third_party_settings: {  }
+  field_overlay_background_image:
+    type: media_library_widget
+    weight: 7
+    region: content
+    settings:
+      media_types: {  }
     third_party_settings: {  }
   field_style_alignment:
     type: options_select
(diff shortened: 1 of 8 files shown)


miketullo95:
@dblanken-yale looks like once we start hitting breakpoints (991px) the overlay size shows 'too much' of my image? is that a my image issue or something we can target? See the before/after in the screenshots. I guess my goal would be to keep the same image treatment throughout the breakpoints? Idk if thats a bad idea lol

**Before > 991px**
Screenshot 2025-06-06 at 10 49 57 AM

After < 991px
Screenshot 2025-06-06 at 10 49 48 AM

miketullo95:
> @miketullo95 I know we talked about it, but wonder if us doing the style-based solution would give us what we really want. Mind if I give it a shot before really evaluating this PR?

yeah give it a try! my only other attempt would be making a smaller pattern to see if its less nasty when going to smaller screens

dblanken-yale:
Ok building; I attempted the background style solution.  I do think we're getting more consistent looks and feels even on mobile.  This might be the better way.  Let me know what you think.  I took the image in my local and shrunk it down a bit to get a better wave overlay.  We might want to go smaller like you said.  If so provide me with one that looks great and I'll add it to storybook so we have a reference for people.  We might even add this into a small description on storybook to guide users to make good image size choices for it.

dblanken-yale:
@miketullo95 Callout is no longer gradient; I'll hold off on the image size until we see what we can come up with.

miketullo95:
> @miketullo95 Callout is no longer gradient; I'll hold off on the image size until we see what we can come up with.

ok I think i added the "true" action banner background size, something funky happening in there? LOL 
Screenshot 2025-06-12 at 10 13 58 AM

dblanken-yale:
@miketullo95 Looking at the image, it looks like it might an image issue?

Screenshot 2025-06-16 at 10 34 38 AM

miketullo95 (approved):
i used a fixed image and this looks great! excited to get this in

ntay90 (approved):

PR #1005, opened 2025-06-16
YSP-885 :: Add Environment Indicator Module

laura-johnson:
### Description of work
- Adds and configures environment indicator module for local, dev, test, multidevs

### Functional testing steps:
- [ ] Verify environment indicator appears 
- [ ] Verify environment indicator is configured for all environments

### Notes: 
- I need to verify how it should be configured for live environments
- I added colors based on how a different site has this module configured, we may want to update the colors


laura-johnson:
A few notes: 

1. I notice on other sites we've configured for this that a common option is not to have the environment indicator on live, just on local/dev/test/multidev. So that's an option.

2. I don't have an option out of the box to have two lines, ie if we want to add supporting text to a new line. That will take some extra work. For now I've added it on one line and we can give that a try. 

3. For local development on other sites we add the branch name, which is a helpful feature. Right now I have "Local Environment - [branch name]" as the text for local.

4. For mutidev I currently am using #e1821f for the hex, and the text is Multidev - [name of the multidev environment, ie pr-1005]

miketullo95 (approved):

PR #1006, opened 2025-06-16
YSP-920 :: Clean Up User Addition Experience for Non-Platform Admins 

laura-johnson:
### Description of work
- Hide add user action and user menu items for non-platform_admins
- Disable caching for admin items on the user list page to avoid caching between users issues

### Functional testing steps:
- [ ] Verify 'Add user' is removed from the Gin toolbar for non-platorm_admins and 'Add CAS User(s)' is now 'Add NetId User(s)'
- [ ] Verify on the CAS add user(s) form, the email field is hidden for all roles
- [ ] Verify there is an alert on the add users page "To add users with NetIds go here [link]"

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.links.action.yml
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.links.action.yml
@@ -30,3 +30,10 @@ ys_core.link_add_profile:
   title: 'Add profile'
   appears_on:
     - view.manage_profiles.page_1
+
+cas.bulk_add_cas_users:
+  title: 'Add NetId user(s)'
+  route_name: cas.bulk_add_cas_users
+  appears_on:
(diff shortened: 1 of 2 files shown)


miketullo95:
@laura-johnson one quick change, could we change the instances where we say "CAS" to "NetID"? 
For example: 

- Field label:  "CAS username(s)" -> "NetID's" 
- Field Help text: "Enter one username per line." -> "Enter one NetID per line"
- Navigation label: "Add CAS Users" -> "Add NetID Users"

miketullo95:
@laura-johnson notes from meeting

- [ ] relabel "Add CAS user(s)" to "Add NetId user(s)"
- [ ] remove drop down link for "Add users"
- [ ] in the "Add Users" page, add an in-line alert before the field inputs to saying something along the lines of "To add users with NetId's go here"

miketullo95 (approved):

PR #1018, opened 2025-07-09
YSP-974: Migration Tool: Profiles CSV Import/Export

nJim:
### Description of work

### Functional testing steps:
- [ ] Step 1
- [ ] Step 2
- [ ] Step 3
- [ ] ...


miketullo95:
@nJim 

one small piece of feedback here, and correct me if this actually would be more effort than its worth. Instead of the preview being behind a toggle and forcing the user to hit "Process CSV" could we have a secondary/primary button situation?


Secondary being a preview CTA, primary being an Import CTA

PR #1031, opened 2025-07-21
YSP-1066 :: Create Dedicated Event Calendar View Block

laura-johnson:
### Description of work
- Adds event calendar block
- Adds event calendar field type, widget, display formatter

### Functional testing steps:
- [x] Add event calendar block to a page in layout builder. [Demo here](https://pr-1031-yalesites-platform.pantheonsite.io/event-calendar-test)
- [x] Verify it displays an event calendar 
- [x] Verify all existing Calendar view blocks are converted to event calendar blocks
- [x] Verify the Calendar view mode is removed from the View block
- [x] Add events, tag with various taxonomies
- [ ] Add exposed filter options to the calendar, verify they filter the events properly in combination with calendar navigation (previous/next month)
- [x] Verify the search exposed filter works as expected

Demo: https://pr-1031-yalesites-platform.pantheonsite.io/event-calendar-test

Atomic PR with block template: https://github.com/yalesites-org/atomic/pull/374
CL PR to adjust weekday header size: https://github.com/yalesites-org/component-library-twig/pull/545

### Notes:

I created a new field type, field widget, and field formatter for the new Events Calendar Block. It actually only needs the formatter right now to render the calendar, so the widget contains no fields. But when it comes time to add the enhanced features, we can use the widget (EventsCalendarDefaultWidget) to pass data over to the formatter (EventsCalendarDefaultFormatter) in JSON format and we can then pass that data to the calendar service.

We can follow the pattern of the ViewsBasicDefaultWidget and ViewsBasicDefaultFormatter to pass params in JSON format from the field values to the formatter.


miketullo95:
@laura-johnson im not sure if this is front end is needed, but it looks like the day labels are bigger in this multi dev compared to live before the release we did yesterday. I was going to give a before and after but it looks like it was caused by some work in the release. Could we try and size down the headers in this PR?

laura-johnson:
> @laura-johnson im not sure if this is front end is needed, but it looks like the day labels are bigger in this multi dev compared to live before the release we did yesterday. I was going to give a before and after but it looks like it was caused by some work in the release. Could we try and size down the headers in this PR?

@miketullo95 Does this look right to you? I think I see where that font size is getting overridden but want to make sure.

Organisms___Calendar_-_Calendar_⋅_Storybook

miketullo95:
@laura-johnson found a small bug here:

- [ ] when you click "x more events" the modal only stays open for a brief second then goes away. i'm assuming some javascript with the filters is causing this
https://github.com/user-attachments/assets/a7e0d39a-e555-4d1b-a424-c8c9b729b965

also i noticed a small inconsistency between the calendar filters and the view filters. Could we:
- [ ]  make the search bar come before the filters 
- [ ] have the search bar dynamically be on the same line as the filter when there are <= 2 filters

calendar search/filters:
Screenshot 2025-08-25 at 4 06 26 PM

view search/filters:
Screenshot 2025-08-25 at 4 06 34 PM

miketullo95:
@laura-johnson icon here: 
Calendar

PR #1039, opened 2025-07-25
Hotfix: YSP-1087: Environment Indicator Changes

dblanken-yale:
## [Hotfix: YSP-1087: Environment Indicator Changes](https://yaleits.atlassian.net/browse/YSP-1087)

__REMINDER: Move target to `main` before merging__

### Description of work
- Updated the environment indicator key from 'dev' to 'development' in settings.php to target the correct environment that users can build and go live with.
- Created an code-based style to override VERY targeted environment indicator defaults
- Created a secret menu item to toggle the display of the environment indicator on a site; this was in an effort to get ahead of any future issues we may have with CAS restricting sites and it possibly showing for users.  Default is on.

#### Development message
- [ ] Trust?  Unless we can get it to a dev/development environment.

#### Bold removal
- [ ] Visit the site and observe the font-weight of the text in the banner
- [ ] Verify it's no longer bold

#### Toggle
- [ ] Ensure that the banner is showing
- [ ] Go to Site Settings as a platform admin, and notice a toggle at the very bottom
- [ ] Turn it off
- [ ] Save and visit the site
- [ ] Verify the banner no longer shows
- [ ] Go toggle it back on
- [ ] Verify the banner now shows

__NOTE__: Not liking the page-level CSS injection, but I just didn't think it was worth overloading the whole twig file for that one CSS change.  I'm willing to take that debt on for now--if we make more changes, we probably will need to rethink this.  I also removed a CSS styling for the indicator that was never getting applied.


miketullo95:
@dblanken-yale it may just be a case of the Mondays for me, but is this missing a multi dev?

dblanken-yale:
@miketullo95 Did we want this in as well?  If I don't hear from ya I'm going to include it Friday since it's mostly visual changes.

PR #1040, opened 2025-07-31
fix(layout): reorder breadcrumbs block in page-meta layout

dblanken-yale:
## [YSP-892: Design treatment for active Content Collect parent menu items](https://yaleits.atlassian.net/browse/YSP-892)

__NOTE: This is to give some context for a meeting we have regarding this request__

### Description of work
- Moves the breadcrumbs underneath the content collection for review and discussion

### Functional testing steps:
- [ ] Make a content collection with subpages
- [ ] View the location of the custom navigation when visiting the parent or any subpage


miketullo95:
@dblanken-yale I believe we can close this one out since we are not moving the breadcrumbs

PR #1041, opened 2025-07-31
YSP-980: Improve Layout Builder Sidebar User Experience

dblanken-yale:
### Description of work
- Implement fully context-aware sidebar state management for Layout Builder.
- Manage Settings and Edit Layout interfaces now have independent sidebar state, width, and behavior.
- Gear icon is hidden and sidebar always open in Manage Settings; Edit Layout allows toggling and resizing (240px-560px).
- State persists per context, with real-time sync and no cross-contamination.
- Includes new CSS, JS, library definition, and conditional attachment.

__CAVEAT: The possible issue here is this is very hard to override, which is why you see it do what Gin would do anyway, and then we modify it, causing movement after page load.  This might not be worth it to keep this going over time.  The only other option is to patch their javascript, which would also be hard to maintain.__

#### Prerequesites
- [ ] Clear existing sidebar state: Open browser DevTools -> Application/Storage tab -> Local Storage -> Delete all keys starting with Drupal.gin.sidebar and YaleSites.layoutBuilder
- [ ] Refresh the page to ensure clean state
#### Main Testing Flow
- [ ] Go to Manage Settings
- [ ] Verify that there is no `toggle sidebar` cogwheel visible
- [ ] Resize the sidebar to be smaller
- [ ] Click Edit Layout and Content
- [ ] Verify that by default the sidebar does not show
- [ ] Use the `toggle sidebar` cogwheel to make it visible
- [ ] Make this as large as it will let you (keep in mind there is a limit set by gin)
- [ ] Go back to Manage Settings
- [ ] See the sidebar is still small
- [ ] Go back to the Edit Layout and Content
- [ ] Verify that the sidebar is visible and that it's width is large
- [ ] Hide the sidebar
- [ ] Go to Manage Settings
- [ ] Verify it's not hidden there
- [ ] Click Edit Layout and Content
- [ ] Verify that it is hidden there


dblanken-yale:
Closing this for now until we're more ready to deal with this.

PR #1044, opened 2025-08-04
Hotfix: re-export config to allow drush deploy to fully import

vinmassaro:
### Description of work
- Re-exports config to allow full config import to complete


dblanken-yale (approved):
So strange; I fixed that in a PR somewhere.  LOL Def approve!

PR #1096, opened 2025-09-23
YSP-368: Display Platform Version Number in Dashboard

dblanken-yale:
### Description of work
- Add logic to `DashboardController` to read the YaleSites platform version from the profile info file using the `info_parser service`.
- Pass the version to the dashboard theme and render it in the template if present.
- Update theme definition to include the new variable.

### Functional testing steps:
- [ ] Visit the dashboard (Content->Dashboard)
- [ ] As of this version, you should see "Platform Version: 2.9.0"


miketullo95:
@dblanken-yale this is pretty much done, but one scope creep option that I kinda want to add.. could we also include the version number in the environment indicator? in this example it would be "Multidev - pr-1096 - V2.9.0"

dblanken-yale:
@miketullo95 That is actually in another PR.  Let me find it.  heh

https://github.com/yalesites-org/yalesites-project/pull/1109

PR #1215, opened 2026-02-23
1105 :: Sustainability news migration

laura-johnson:
### Description of work
- Adds migration module with migration yml files and NewsToLayoutBuilder plugin for migration of sustainability news

Note: We won't be able to run the migration on a multidev because it's not feasible to connect to a source database and files from a multidev, but I imported my database and the news files from my local to the multidev after I ran the migration as a demo. For the live migration I'll need to checkout the sustainability repo and add the remote (d7) database connection to run the migration once the migration files have been merged.

### Functional testing steps:
- [ ] You can see the migrated news here: https://pr-1215-yalesites-platform.pantheonsite.io/posts
- [ ] Edit a post. 
-- Verify the teaser text and teaser image have been migrated as expected
- - Verify any category tags from D7 have been migrated to the category field.
- [ ] Edit layout for a post. 
-- Verify the title and publish date have been migrated to the title and metadata block. 
-- Verify the body has been migrated to a text block. Verify the text looks clean and that any [[ nid: xxx ]] tags have been removed
-- Verify the image has been migrated to an image block at the top of the post.
- [ ] In the layout, edit an image.
-- Verify the image has been migrated to a media item
-- Verify the alt text has been migrated
-- Verify the caption has been migrated
-- Verify that the media has been tagged with "Imported from migration"

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_migrate/modules/ys_migrate_sustainability_news/README.md
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_migrate/modules/ys_migrate_sustainability_news/README.md
@@ -1,0 +1,166 @@
+# YaleSites Migrate: Sustainability News
+
+Migrates Drupal 7 **news** nodes from the Sustainability Yale D7 site into
+Drupal 10 **post** nodes on the current YaleSites platform, including taxonomy
+terms, images, media entities, and Layout Builder sections.
+
+---
+
+## Requirements
+
+### Drupal modules
+
+The following modules must be enabled before running the migration:
+
+- `ys_migrate` (parent module)
+- `ys_migrate_sustainability_news` (this module)
+- `migrate_plus`
+- `migrate_tools`
+
+Enable with:
+
+```bash
+lando drush en ys_migrate_sustainability_news
+```
+
+---
+
+## Step 1 — Set up the D7 database connection
+
+The D7 source database must be available as a second database connection named
+`d7_sustainability` in `web/sites/default/settings.local.php`:
+
+```php
+$databases['d7_sustainability']['default'] = [
+  'database' => 'd7_sustainability',
+  'username' => 'root',
+  'password' => '',
+  'host'     => 'database',
+  'port'     => '3306',
+  'driver'   => 'mysql',
+  'prefix'   => '',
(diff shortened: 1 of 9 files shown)


web/profiles/custom/yalesites_profile/modules/custom/ys_migrate/modules/ys_migrate_sustainability_news/config/optional/migrate_plus.migration.ys_sn_files.yml
```diff
@@ -0,0 +1,101 @@
...
+#
+#   From a local archive:
+#     ddev exec tar -xzf /var/www/html/d7-files.tar.gz \
+#       -C /var/www/html/web/sites/default/files/d7_source/
+#
+#   From a local directory (outside the container):
+#     cp -r /path/to/d7/sites/default/files/* web/sites/default/files/d7_source/
+#
+# After migration, remove the staging directory:
+#   ddev exec rm -rf /var/www/html/web/sites/default/files/d7_source
```
dblanken-yale, line 32: @laura-johnson Would we be able to have the mirroring lando documentation as well?  If I remember right you're using ddev, but our docs and scripts still currently use lando, so wouldn't want to confuse others trying to re-use this in the future.

web/profiles/custom/yalesites_profile/modules/custom/ys_migrate/modules/ys_migrate_sustainability_news/README.md
```diff
@@ -0,0 +1,166 @@
...
+### Drupal modules
+
+The following modules must be enabled before running the migration:
+
+- `ys_migrate` (parent module)
+- `ys_migrate_sustainability_news` (this module)
+- `migrate_plus`
+- `migrate_tools`
+
+Enable with:
```
dblanken-yale, line 20: @laura-johnson Also thinking here maybe we should be explicit that this is to be done on a local environment?  Just again worried about those that might not have a lot of D7 migration experience, so maybe somewhere here we're explicit that they need to bring down the repo to perform this?

dblanken-yale (approved):
Going to resolve since we're eventually going toward ddev.

PR #1232, opened 2026-03-09
#1120 :: Bump up text size for in-line alert content - DEMO ONLY, DO NOT MERGE

laura-johnson:
### Description of work
- Use default rather than condensed token for message text

### Functional testing steps:
- [ ] Check the text size of the message here: https://pr-1232-yalesites-platform.pantheonsite.io/inline-message-demo

CL PR: https://github.com/yalesites-org/component-library-twig/pull/624

### DIFF
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -27,6 +27,7 @@
     "drupal/anchor_link": "3.0.0-beta1",
     "drupal/auto_entitylabel": "3.3.0",
     "drupal/better_exposed_filters": "6.0.5",
+    "drupal/bigmenu": "2.1.0",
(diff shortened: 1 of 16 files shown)


miketullo95:
@laura-johnson crap, i may take the blame on this one. This is actually meant to target the "in-line message" content block, not the site-wide alert. my use of "alert" was wrong in the ticket 😦. appologies for that. Could we target the in line message content text instead?

miketullo95:
Screenshot 2026-03-23 at 1 32 13 PM

laura-johnson:
@miketullo95 I put the alert back to condensed, and gave the default size to the inline message component text. [Demo here](https://pr-1232-yalesites-platform.pantheonsite.io/inline-message-demo)

miketullo95 (approved):

PR #1236, opened 2026-03-19
#1137 :: Mobile: Unintended mega menu left and right scrolling - DEMO ONLY, DO NOT MERGE

laura-johnson:
### Description of work
- Disable horizontal scrolling in mega menu

### Functional testing steps:
- [ ] Use dev tools and/or mobile device to test mobile menu
- [ ] Verify there is no ability to horizontal scroll

CL PR: https://github.com/yalesites-org/component-library-twig/pull/626


laura-johnson:
@miketullo95 This is ready for review. Note: I couldn't always reproduce this, I think the content has something to do with it. I also saw that on some pages you could move the whole page container including header, body, footer... like [here](https://dev-yalesites-platform.pantheonsite.io/generic-homepage). I _think_ my fixes have fixed both issues. Please take a look and let me know what you think.

miketullo95:
> @miketullo95 This is ready for review. Note: I couldn't always reproduce this, I think the content has something to do with it. I also saw that on some pages you could move the whole page container including header, body, footer... like [here](https://dev-yalesites-platform.pantheonsite.io/generic-homepage). I _think_ my fixes have fixed both issues. Please take a look and let me know what you think.

i think maybe the mega menu was also a factor. I tested this PR out on mobile and desktop and looks to be fixed. the final test will be when we cut RC's and test it on the site the original bug was reported on (usability)

miketullo95 (approved):

PR #1241, opened 2026-04-07
1173: "Moderation control" placeholder text visible on published nodes

dblanken-yale:
### Description of work
- Adds `ExtraFieldPlaceholderSubscriber` to `ys_core` — an event subscriber
  on `SECTION_COMPONENT_BUILD_RENDER_ARRAY` (priority 50) that removes
  `#markup` from Layout Builder ExtraFieldBlock builds outside of preview mode
- Targets `content_moderation_control` by default; supports an `['all']`
  wildcard or `[]` to disable entirely
- While investigation suggests the visible text on divinity.yale.edu stems
  from a stale Varnish/page cache entry, the placeholder should never be
  reachable by anonymous users under any circumstances — this ensures it
  cannot be cached in that state going forward
- Adds 8 unit tests covering all TARGETED_FIELDS modes and preview passthrough

### Functional testing steps:
- [ ] Visit a published event, page, post, profile, or resource as an
  anonymous user — confirm "Placeholder for the 'Moderation control' field"
  text does not appear
- [ ] Open a node in Layout Builder edit mode — confirm the "Placeholder for
  the 'Moderation control' field" label still appears in the editor preview
  for the Moderation control block
- [ ] After deploying to pkp, clear Pantheon page cache for affected URLs and
  confirm the placeholder text is gone

Closes yalesites-org/YaleSites-Internal#1173

Again, I could not recreate this, and feel a cache clear was the answer to this, but this does ensure that if it happens to attempt to be shown it will remove it.  The other option is to close this and wait for the next time this happens to implement.  I am hesitant to add this code if it's not widespread.

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/EventSubscriber/ExtraFieldPlaceholderSubscriber.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/EventSubscriber/ExtraFieldPlaceholderSubscriber.php
@@ -1,0 +1,73 @@
+<?php
+
+namespace Drupal\ys_core\EventSubscriber;
+
+use Drupal\layout_builder\Event\SectionComponentBuildRenderArrayEvent;
+use Drupal\layout_builder\LayoutBuilderEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Prevents ExtraFieldBlock placeholder text from rendering on published pages.
+ *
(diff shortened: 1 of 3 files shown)


miketullo95:
@dblanken-yale i think we close this and keep this in mind when it comes to a site admin cache clear function issue

miketullo95:
unless you think we should commit these changes anyways?

miketullo95:
per restocking 4/20/26 -  we are going to close this but keep the branch in case we really need this. Current solution is to cache clear

PR #1246, opened 2026-04-17
#1160 :: SoundCloud embed code with double-encoded track ID fails validation

laura-johnson:
### Description of work
- Update soundcloud regex to make double-encoded prefix optional

### Functional testing steps:
- [ ] Add Soundcloud embed with the new double-encoded format (see ticket), verify it passes validation and works properly. Verify the original style of embed also passes validation and works properly

Demo: https://pr-1246-yalesites-platform.pantheonsite.io/soundcloud-test

(the top one is the example from the ticket, and the bottom one is the original embed code version)


laura-johnson:
@miketullo95 @dblanken-yale I'm not sure why the CI is failing. The error message is not very helpful.  Let me know if you know what the issue is already or if you'd like me to dig in.

miketullo95 (approved):

PR #1251, opened 2026-04-24
#1158 :: Adding link without https:/ in quick links block causes error

laura-johnson:
## [#1158: Adding link without https:/ in quick links block causes error](https://github.com/yalesites-org/YaleSites-Internal/issues/1158)

Drupal core: when you type a scheme-less string like google.com, core's LinkWidget rewrites it to internal:google.com, an intentionally-invalid value,expecting its own validator to reject it. That behavior is core's, and on its own it just produces a validation error, not a crash.

linkit (contrib) is what caused the fatal error. Its widget calls $item->getUrl() while rebuilding the form, and the deployed version did so without a try/catch. That uncaught InvalidArgumentException is the white screen. That's a contrib bug, but we already have a patch for it.

I've added a normalizer so that we can avoid the validation error. The linkit patch takes care of the white screen.

### Description of work
- The fix adds a normalizer to prevent the validation error. 

### Functional testing steps:
- [ ] Add a quick links block with at least one link pointing to a url with no protocol (like google.com). Verify there is no white screen.


miketullo95:
@laura-johnson sorry, not sure if this one is ready for review yet or not, but i did some testing.

- [ ] on a page, add a quick links block
- [ ] in the link field, add a link "google.com"
- [ ] try to save block, get the expected error - this is fine
- [ ] in that same link field, add "https://www." to the beggining of that link to try and fix the expected error
- [ ] white screen error 


Type | php
-- | --
Monday, April 27, 2026 - 11:08
mt927
https://pr-1251-yalesites-platform.pantheonsite.io/layout_builder/add/block/overrides/node.108/2/content/inline_block%3Aquick_links
https://pr-1251-yalesites-platform.pantheonsite.io/layout_builder/add/block/overrides/node.108/2/content/inline_block%3Aquick_links
InvalidArgumentException: The internal path component 'google.com' is invalid. Its path component must have a leading slash, e.g. internal:/foo. in Drupal\Core\Url::fromInternalUri() (line 416 of /code/web/core/lib/Drupal/Core/Url.php).
Error
32.219.6.36


https://pr-1251-yalesites-platform.pantheonsite.io/admin/reports/dblog/event/38659

laura-johnson:
@miketullo95 Thanks, I'll add better validation!

miketullo95 (approved):

PR #1270, opened 2026-05-13
#561 :: Blocks containing max paragraph children error

laura-johnson:
### Description of work
These are all kind of special cases:
- Quick links actually doesn't have child paragraphs, it's a Link field with unlimited cardinality. So I had to add a helper function in the code so that our validation ignores empty items (there is no option not to add new Link items).
- Gallery has a minimum of 2 but no maximum, so I made no change there.
- Link Grid had a maximum cardinality set in the paragraph settings, so that once you reach 4 Link Lists, there is no option to add another one
- I updated Tabs to have maximum cardinality of 5 tabs set in paragraph settings also, so now once you reach 5, there is no option to add another.

### Functional testing steps:
- [ ] Add a block of type Quick Links Add the max number of items. Edit the item in Layout Builder. Verify that if an empty item is added you can still save. Verify that if you fill that item, the validation is triggered.
- [ ] Add a block of type Tabs. Verify that once you have added 5, you do not have an option to add more. Verify you must add at least 2.
- [ ] Add a block of type Link Grid. Verify that once you have added 4 Link Lists, you do not have the option to add more.

Demo: https://pr-1270-yalesites-platform.pantheonsite.io/node/128/layout

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/field.field.block_content.tabs.field_tabs.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/field.field.block_content.tabs.field_tabs.yml
@@ -25,94 +25,37 @@ settings:
       tab: tab
     negate: 0
     target_bundles_drag_drop:
-      accordion:
-        weight: 31
-        enabled: false
       accordion_item:
         weight: 32
-        enabled: false
-      basic_view:
-        weight: 33
-        enabled: false
-      callout:
-        weight: 34
         enabled: false
       callout_item:
         weight: 35
         enabled: false
-      cta_banner:
-        weight: 36
+      cta_link:
+        weight: 14
         enabled: false
       custom_card:
         weight: 37
         enabled: false
-      custom_cards:
-        weight: 38
-        enabled: false
-      divider:
-        weight: 39
-        enabled: false
-      embed:
-        weight: 40
-        enabled: false
-      event_cards:
-        weight: 41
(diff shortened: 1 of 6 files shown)


miketullo95 (approved):

miketullo95 (approved):

web/profiles/custom/yalesites_profile/config/sync/field.field.block_content.tabs.field_tabs.yml
```diff
@@ -25,94 +25,37 @@ settings:
       tab: tab
     negate: 0
     target_bundles_drag_drop:
-      accordion:
```
dblanken-yale, line 28: Do you mind explaining these changes a little more? I think this is a non-issue since these aren't allowed anyway, but wanted to make sure this wasn't residual from an old db export?

web/profiles/custom/yalesites_profile/config/sync/field.storage.block_content.field_tabs.yml
```diff
@@ -14,7 +14,7 @@ settings:
   target_type: paragraph
 module: entity_reference_revisions
 locked: false
-cardinality: -1
```
dblanken-yale, line 17: When the config imports, what happens to any existing site that had more than 5 items already? The next time they edit they'd have to remove or does this change remove their data?

PR #215, opened 2023-02-04
YALB-902: AX: Hide "Advanced" metadata tab

marcb4k:
### Description of work
- Removes advanced metatags from node edit forms

### Functional testing steps:
- [x] Edit and existing or create a new node
- [x] In the sidebar, open the Metadata details accordion
- [x] Verify that there is only a "Basic tags" section and the "Advanced tags" no longer appears.

PR #1272, opened 2026-05-15
#215 :: External links in menus do not show external link

laura-johnson:
### Description of work
- Adds preprocess to make the menu URL an external link. This causes the correct attribute to be added so that it displays properly

### Functional testing steps:
- [ ] Add a node with an external link specified in edit node/manage settings
- [ ] Use the menu settings in edit node/manage settings to add a menu item to any menu
- [ ] Verify the menu link goes to the external URL and the menu item has the correct external link decoration

Demo: https://pr-1272-yalesites-platform.pantheonsite.io/ see "Externally linked node test" in the menu

Node settings here: https://pr-1272-yalesites-platform.pantheonsite.io/node/128/edit

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module
@@ -470,6 +470,62 @@ function ys_core_preprocess_menu(&$variables) {
       }
     }
   }
+
+  // Replace menu item URLs with external source URLs where applicable.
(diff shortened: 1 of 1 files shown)


miketullo95 (approved):

dblanken-yale (commented):
@miketullo95 and I are going to discuss a possible inconsistency with the resources content type, but I had a question about how the menu reacts to external source changes?  Do we need to cache invalidate in those instances?

PR #1277, opened 2026-05-22
fix: re-export cas config

vinmassaro:
### Description of work
04fdce78fca20e143be43a318b78ac255527f80b introduced bad config changes that are fixed by this config re-export.


dblanken-yale (approved):
Tested locally and works.

PR #1296, opened 2026-06-18
1310: RC: CAS-protected page shows maintenance screen to unauthenticated users

dblanken-yale:
### Description of work
- Guard `$block['#plugin_id']` with a null-coalescing default in `YSLayoutBanner::build()`. When a CAS-protected page was saved, search_api's post-request indexing rendered the node and ran this method; a single-block content region whose block render array lacked `#plugin_id` passed `null` to `str_contains()`, raising an undefined-array-key warning and a deprecation. That error surfaced the site maintenance screen to unauthenticated visitors instead of the expected CAS login redirect.
- Existing moderation-control hide behavior is unchanged.
- Add unit coverage for the missing-`#plugin_id` path and the published-moderation-control hide path.

### Functional testing steps:
- [ ] On the multidev, create a new page, add it to the site menu, enable CAS protection, and save.
- [ ] Open the page in an incognito (unauthenticated) window.
- [ ] Confirm you are redirected to CAS login rather than seeing the "Site under maintenance" screen.
- [ ] Confirm no PHP warnings/deprecations are logged during the save/indexing of a Banner-layout node.

References yalesites-org/YaleSites-Internal#1310

Created with AI

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -110,12 +110,7 @@
   "config": {
     "audit": {
       "ignore":[
-        "SA-CONTRIB-2025-060",
-        "SA-CORE-2026-005",
-        "SA-CORE-2026-006",
-        "SA-CORE-2026-007",
-        "SA-CORE-2026-008",
-        "SA-CORE-2026-009"
+        "SA-CONTRIB-2025-060"
       ]
     },
     "preferred-install": "dist",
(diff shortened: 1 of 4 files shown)


miketullo95 (approved):
confirmed these testing steps results in a correct way:

1. Create a new page
2. Add it to the site menu
3. Enable CAS protection on the page
4. Save
5. Open the page in an incognito window (unauthenticated)

PR #1299, opened 2026-06-18
1161: Views rework

dblanken-yale:
## [1161: Views Block Architectural Rework](https://github.com/yalesites-org/YaleSites-Internal/issues/1161)

Refactors the monolithic `view` block into a layered, per-content-type architecture (one stacked PR covering the whole epic).

### Description of work
- **[#1163](https://github.com/yalesites-org/YaleSites-Internal/issues/1163)** — abstract `ViewsBasicWidgetBase` + bundle-keyed `LISTING_BUNDLES` definition.
- **[#1306](https://github.com/yalesites-org/YaleSites-Internal/issues/1306)** — per-instance render isolation (clone scaffold view, unique pager id); fixes [#906](https://github.com/yalesites-org/YaleSites-Internal/issues/906).
- **[#1164](https://github.com/yalesites-org/YaleSites-Internal/issues/1164) / [#1165](https://github.com/yalesites-org/YaleSites-Internal/issues/1165) / [#1166](https://github.com/yalesites-org/YaleSites-Internal/issues/1166) / [#1167](https://github.com/yalesites-org/YaleSites-Internal/issues/1167)** — Post / Event / Page / Profile widgets + 13 listing bundles.
- **[#1316](https://github.com/yalesites-org/YaleSites-Internal/issues/1316) / [#1317](https://github.com/yalesites-org/YaleSites-Internal/issues/1317)** — authoring-form UX audit (spec) + grouped sections, labels, help text.
- **[#1318](https://github.com/yalesites-org/YaleSites-Internal/issues/1318)** — client-side live mockup preview.
- **[#1168](https://github.com/yalesites-org/YaleSites-Internal/issues/1168)** — picker grouping (4 categories) + per-display-mode placement restrictions.
- **[#1169](https://github.com/yalesites-org/YaleSites-Internal/issues/1169)** — in-place `view` to bundle migration deploy hook (all-revisions layout rewrite).
- **[#1170](https://github.com/yalesites-org/YaleSites-Internal/issues/1170)** — deprecate legacy `view`; supersede `post_list`/`event_list`/`directory`; admin warning; CHANGELOG.
- **[#1171](https://github.com/yalesites-org/YaleSites-Internal/issues/1171)** — QA regression matrix + automated coverage (33 tests).
- **[#1172](https://github.com/yalesites-org/YaleSites-Internal/issues/1172)** — README / architecture documentation.
- **[#1366](https://github.com/yalesites-org/YaleSites-Internal/issues/1366)** — closed multi-select filters show the selected term chips (with an accessible "+N more" overflow badge) instead of "(X) items selected".
- **[#1367](https://github.com/yalesites-org/YaleSites-Internal/issues/1367)** — exposed filters and search auto-run (Better Exposed Filters autosubmit over core Views AJAX); the Apply button is hidden when JS is on and kept as a no-JS fallback.
- **QA feedback fixes** (from the review rounds on this PR):
  - Recursion guard for `views_pre_view` (fixed the "Add Section" hang and the exposed-filter 500 that prevented faceting).
  - Live mockup preview scoped to the create form, varying by display mode and inheriting Gin's color scheme.
  - **Context-derived listing-card heading levels** — result cards render H2 when the block has no heading and H3 when nested under one, across card/list/condensed/directory for post, event, page, and profile (no more skipped levels).
  - **Accessible result-link hover** — hover uses the section's on-theme foreground (`--color-layout-content`, AA on every theme) instead of `--color-slot-two`, so it stays legible on every section background and never disappears when the background uses the hover slot.
  - Title-link shadow + date/byline (overline) contrast on themed sections; card column counts by section (full=3 / 50-50=2 / 33-33-33 & 30=1).
  - **AA contrast on themed sections** — list-view result separators, the pager (links/hover/current-page marker), and the category/tag term labels use the section's on-theme tokens instead of fixed colors that failed WCAG AA on slots 2/4/5. Default (non-themed) appearance is unchanged.
  - **Category pass-through verified + regression test** — category filtering works end-to-end (a post list narrows via the exposed dropdown); a kernel test (`ExposedCategoryFilterTest`) locks the config contract (exposed, multi-value category filter on both scaffold views).
  - **Filter-bar layout** — the scaffold exposed-filter controls lay out on a grid **capped at four columns**, sized to the form's own column width, with the **search field on its own full-width first row**; it stacks gracefully in constrained sections (50/50, 70/30) and no longer packs 5+ filters onto one row or overflows the region.
  - **Latest QA round (07-05):**
    - **Apply button removed** — now that filters auto-run (#1367), BEF's `autosubmit_hide` hides the submit button (kept in the DOM as a no-JS fallback).
    - **Auto-submit advisory note removed** — the visible "Results update automatically…" note is gone; auto-submit is still conveyed to assistive tech by the polite aria-live result-count announcement.
    - **Filter chips reflow on removal** — removing a truncated term now recomputes the visible chips + "+N more" against the closed layout (was leaving a stale "N more") — recompute on `chosen:hiding_dropdown`.
    - **Resource formatter NULL guard** — `ViewsContentResourcesDefaultFormatter` no longer warns ("array offset / exposed_widgets on null") when its manager returns NULL during search_api indexing; mirrors the `ViewsBasicDefaultFormatter` guard. (The separate search_api "headers already sent" terminate-time/session error is a contrib issue and should get its own ticket; the guard does not address it.)
- chore: temporarily ignore `SA-CORE-2026-005..009` in composer (resolved — branch now rebased on develop's Drupal 10.6.11; the ignore commit was removed).

Companion PRs (same `1318-views-rework` branch in each repo):
- **atomic** — yalesites-org/atomic#466 (listing-card node templates: context-derived heading levels; auto-submit announcement + advisory-note removal).
- **component-library-twig** — yalesites-org/component-library-twig#649 (link shadow + overline contrast in themed sections; accessible result-link hover token; **filter-bar grid capped at four columns with search first**; region-width card columns; AA contrast for pager/separators/term labels; **filter chips + reflow-on-removal**).

Flagged follow-ups (not in this PR): staging migration validation with real data ([#1169](https://github.com/yalesites-org/YaleSites-Internal/issues/1169) / [#1170](https://github.com/yalesites-org/YaleSites-Internal/issues/1170)); manual Layout Builder placement QA ([#1171](https://github.com/yalesites-org/YaleSites-Internal/issues/1171)); cross-repo removal of predecessor `atomic` templates ([#1170](https://github.com/yalesites-org/YaleSites-Internal/issues/1170)); design/accessibility sign-off ([#1316](https://github.com/yalesites-org/YaleSites-Internal/issues/1316) / [#1317](https://github.com/yalesites-org/YaleSites-Internal/issues/1317) / [#1318](https://github.com/yalesites-org/YaleSites-Internal/issues/1318)); **Resource View block** integration deferred to its own ticket ([#1338](https://github.com/yalesites-org/YaleSites-Internal/issues/1338)) — it rides a dedicated parallel module (`ys_views_content_resources`) and needs alignment with the ys_views_basic extension model; block-picker icons ([#1336](https://github.com/yalesites-org/YaleSites-Internal/issues/1336)) and the authoring-form UX/UI audit ([#1337](https://github.com/yalesites-org/YaleSites-Internal/issues/1337)); search_api terminate-time "headers already sent" session error surfaced while indexing a resource block (separate contrib issue).

### Functional testing steps:
- [ ] Place each new listing block (Posts/Events/People/Pages - Card/List/Condensed, People - Directory) and confirm correct settings + render.
- [ ] Confirm placement restrictions: List/Directory blocked from sidebar/50-50/33-33; Card/Condensed everywhere.
- [ ] Place two listing blocks with different settings on one page; confirm no cross-block clobbering ([#906](https://github.com/yalesites-org/YaleSites-Internal/issues/906)).
- [ ] Run `drush deploy` on a copy of real data; confirm legacy `view` blocks migrate with no visual change.
- [ ] Confirm legacy/predecessor blocks no longer appear in the picker.
- [ ] Confirm result-card heading levels: a listing block with no heading → first card is H2 (no skipped level); a block with a heading → H2 block heading + H3 cards.
- [ ] Confirm the result-link hover stays legible across all section background themes, including a section whose background uses the hover color slot.
- [ ] Confirm event-category faceting filters results (recursion fix).
- [ ] Confirm category pass-through on a post list: enable the category exposed filter, select a term → results narrow to matching posts.
- [ ] In a themed section (slots 2/4/5), confirm list-view separators, pager (including the current-page marker), and category/tag term labels meet contrast and stay legible; confirm default (white) sections are unchanged.
- [ ] **Filter-bar grid**: with a listing that has search + 3–4 filters, confirm the search field is on its own full-width first row, the filters lay out at most four per row at equal widths, and in 50/50 and 70/30 sections they stay within the region (no horizontal overflow); a very narrow column collapses to a single column without overflowing.
- [ ] **Auto-submit**: pick a filter value (or type 3+ characters in search) → results update on their own with no Apply button visible; a screen reader announces the new result count; keyboard focus stays on the control. With JS disabled, the Apply button appears and submits.
- [ ] **Filter chips**: select several terms in a multi-select filter, close the dropdown → chips show with a "+N more" badge; remove one of the collapsed/truncated terms and reopen/close → the visible chips and "+N more" recompute correctly (no stale count).
- [ ] **Resource block**: place/configure a Resource View block and load a page that indexes it; confirm no "array offset on null" / "exposed_widgets on null" warnings in the log for `ViewsContentResourcesDefaultFormatter`.
- [ ] **Preview panel height (2026-07-29)**: in **Gin dark mode**, add an **Events - Card Grid View** block and open the **Field display & filters** tab. The dashed preview box on the right should wrap tightly around the mockup card. Under **Exposed Filter Options** tick **Show Category** then **Show Custom Vocab** — their conditional fields appear on the left and the dashed box must **not** grow (no large empty panel beside the controls).
- [ ] **Safari number steppers (2026-07-29)**: on the **Display options** tab in **Safari**, the up/down stepper arrows on **Skip the first results** (and on **Items** / **Items per Page**) should be light and clearly visible against the dark input, not dark-on-dark. Re-check in Gin **light** mode — arrows should still be correct there (the fix is scoped to dark mode).
- [ ] **Single-column fallback**: narrow the browser until the Field display tab collapses to one column — the preview should stack below the controls as before.
- [ ] **Exposed filter accordion (2026-07-29)**: on **Field display & filters**, each exposed filter should be its own bordered row. **Show Search** / **Show Audience** (and **Show Year** on posts) are header-only with no chevron. **Show Category** and **Show Custom Vocab** show a chevron; switching one on discloses its settings in that row and switching it off collapses them and disables the controls. Confirm the settings appear *below* their own checkbox, not further down the column.
- [ ] **Preview no longer stretches (2026-07-29)**: with the preview visible, switch on **Show Category** then **Show Custom Vocab**. The dashed preview box must stay the same height throughout (it shares a row only with the field-display checkboxes now).
- [ ] **Filter values still round-trip**: switch on Search + Category, set a **Category Filter Label**, save the block and the layout, then reopen the block — the same filters should be on, the others off, and the label preserved. On the front end, confirm only the enabled filters are exposed (in particular that turning **Show Audience** *off* actually removes the audience filter).
- [ ] **Per-type field options**: on an event listing tick **Hide Add to Calendar link**, and on a post listing tick the eyebrow option; save and reopen to confirm each persists.

- [ ] **Chip labels not clipped (2026-08-05)**: on the front end, select terms in a multi-select filter whose labels contain a descender (a "y", "g", "p" — e.g. an **Audience** term like "Faculty"), then close the dropdown. The full label must be visible inside the chip, with no letters cut off along the bottom edge. Select enough terms to force the `+N more` badge and confirm the chips and the badge read at the same size on a single line, and that the filter control is still the same height as the search field beside it.
- [ ] **Config modal stays reachable (2026-08-05)**: in a **tall browser window** (this only shows when the form initially fits without scrolling — maximise on a large display), add a **Posts - Card Grid View** block and open the **Results & filters** tab. Tick **Show Category**. The dialog must stay fully inside the window with its buttons reachable, re-centring itself as it grows — previously its bottom edge went below the viewport and could not be scrolled to until you resized the window. Untick it and confirm it re-fits again. Also confirm the same toggle behaves on **an existing block's** Configure (off-canvas) panel, which should scroll internally as before.

- [ ] **Front-end filters are not dark for logged-in editors (2026-08-06)**: put Gin in **dark mode**, then — **logged in** — open a page with a listing block that has exposed filters. Every filter control must render **light** (white box, dark text), the same as it looks logged out; no dark bars. Open a dropdown: the panel is white with dark options. Repeat on a page with an **Event Calendar** block, which is the other front-end surface that loads this CSS.
- [ ] **Authoring form is still dark (2026-08-06, regression check)**: still in Gin dark mode, open **Configure** on an existing listing block and confirm the Category/Tag dropdowns *in the form* are still **dark**. That styling is meant to be there — only the front-end leak was removed.
- [ ] **Pill sits inside its box (2026-08-06)**: on the front end, select a term in a multi-select filter. The pill must sit **inside** the control with the control's border **unbroken all the way around, including underneath the pill** — previously the bottom border disappeared behind the pill. Select enough terms to trigger "+N more": the row stays one line and the control keeps its height. Open the dropdown — the box grows, every chip is visible, nothing spills past the border — then close it and confirm it returns to one line at the original height. Check both **logged in** and **logged out**; they should look identical.

References yalesites-org/YaleSites-Internal#1161

Created with AI


miketullo95 (changes requested):
@dblanken-yale Found 4 bugs testing on the pr-1299 multidev, likely from the latest QA-round work. Details below with exact locations where I could pin them down.

**1. Exposed filter multiselect renders with admin-theme styling on the front end, logged-in users only (incognito is fine)**

Screenshot shows the Audience/Custom Vocab "Choose some options" chosen widget rendering as a solid dark bar instead of the normal light front-end style. This only reproduces when authenticated.

I haven't ruled out a caching artifact yet — on Pantheon, anonymous requests get served through the CDN (fully purged/re-primed on deploy) while authenticated requests bypass the CDN and hit Drupal's internal render/dynamic-page cache directly. If that wasn't cleared after the last deploy to this multidev, logged-in users would see stale markup while anon doesn't. Can you run `drush cr` on the multidev and re-test logged in first? If it still reproduces after a clean cache clear, it's a real bug and I'd look at whether the exposed filter block's render array is picking up a different library set or cache context for authenticated users (nothing in this repo's diff attaches `ys_views_basic` CSS libraries outside the block config `$form` itself, so if it's leaking onto the front end for real, that's the place to check).

**2. Term filter shows old "(1 term selected)" text instead of the new chip design, logged-in users only**

Same symptom pattern as #1 (logged-in only, incognito fine) — the #1366 chip/badge work not showing up reads like the same stale-cache possibility. Please check after a `drush cr` before treating as a regression; if it's still showing the old markup post-clear, there's a cache tag/context gap on the closed multiselect summary render that isn't invalidating per role.

**3. Term pill clipped at the bottom, logged-out users**

This one's a real CSS bug, not cache. In `component-library-twig`, [`components/01-atoms/forms/_yds-form.scss:215-223`](https://github.com/yalesites-org/component-library-twig/blob/1318-views-rework/components/01-atoms/forms/_yds-form.scss#L215-L223):

```scss
.chosen-container {
  ...
  height: 3.125rem;
  ...
}
```

paired with `.chosen-choices` at `:237-252` setting `overflow: hidden` on the flex row inside that fixed 50px box. The `li.search-choice` chip's own padding/line-height renders taller than 3.125rem can hold, so `overflow: hidden` clips the bottom of the pill. Suggest either bumping the height on `.form-item__dropdown` (line 206) and `.chosen-container` (line 223) to fit the chip's actual rendered height, or trimming the chip's vertical padding to fit inside 3.125rem — whichever keeps the single-line "+N more" closed-state design from #1366 intact.

**4. Configure block modal locks scroll when a setting (e.g. Show Category) extends its height; resizing the browser window fixes it**

In `yalesites-project`, `ViewsBasicWidgetBase::buildExposedFilterAccordion()` ([`ViewsBasicWidgetBase.php:756`](https://github.com/yalesites-org/yalesites-project/blob/1318-views-rework/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/Plugin/Field/FieldWidget/ViewsBasicWidgetBase.php#L756)) builds the accordion row, but the actual disclosure is pure CSS in [`views-basic.css:571-572`](https://github.com/yalesites-org/yalesites-project/blob/1318-views-rework/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/assets/css/views-basic.css#L571-L572):

```css
.vb-filter-row--has-settings:not(:has(input[type="checkbox"]:checked)) .vb-filter-row__body {
  display: none;
}
```

Toggling the checkbox reveals the body via `:has()` alone — no JS event fires to tell the off-canvas/gin_lb dialog to recompute its scrollable height. Once the newly-revealed content grows the form past the modal's already-computed height, the dialog's scroll region goes stale until something else (a window resize) forces recalculation. I didn't find any resize/reposition trigger anywhere in this PR's JS. Suggest wiring the accordion checkboxes' `change` event to explicitly re-trigger the dialog's resize/reposition (e.g. `$(window).trigger('resize')` or whatever event gin_lb's off-canvas dialog listens for) — probably in `views-basic.js` alongside the other behaviors there.

All 4 blocking for this pass given the size of this epic and how many QA rounds have already gone into filter/modal behavior — flagging now before it gets buried under the rest of the diff.

dblanken-yale:
## Follow-up to address on this PR

One unscoped rule in `views-basic.css` still reaches the front end — flagged in https://github.com/yalesites-org/yalesites-project/pull/1299#issuecomment-5209361801.

**The line number in that comment has drifted.** On the current branch tip, line 72 is blank. The actual offender is **line 82**, `.chosen-container-multi .chosen-choices` (3 declarations), sitting under the `/* chosen-specific */` heading. Everything else in the file is already scoped to `.views-basic--*`, `.layout-builder-configure-block`, `.gin--dark-mode`, `#layout-builder-modal`, or `.vb-filter-*`.

Prefixing that selector with `.layout-builder-configure-block` closes it.

**Why here rather than a ticket:** it is a one-line scope fix in a file this PR already grew from 306 to 916 lines, and the PR has a review round coming anyway. The larger front-end/admin library split described in the same comment is *not* in-branch work — that is being ticketed separately. This is only the residual leak.

Setting to `needs work` so it does not get merged past.

dblanken-yale:
## Review follow-up addressed: the residual `views-basic.css` leak is closed

Pushed to `1318-views-rework` (`f35c7e2e0..7d584e0ed`). Companions got a `develop` merge only —
no code changes were needed in either this round: component-library-twig#649
(`04837580..74dcad0c`), atomic#466 (`be3683a2..51496324`).

Nothing is waiting on you. Two small corrections to the finding's detail are at the bottom;
neither changes the fix.

### The finding

> One unscoped rule in `views-basic.css` still reaches the front end. […] the actual offender is
> **line 82**, `.chosen-container-multi .chosen-choices` […] Prefixing that selector with
> `.layout-builder-configure-block` closes it.

**Confirmed and fixed as specified.** I re-derived the selector list rather than take the line
number on faith: of the 10 rules in the file that mention `chosen`, exactly one lacked an admin
qualifier, and it was that one at line 82. The other nine already carry
`.layout-builder-configure-block`.

I kept to the boundary you drew — this is only the residual leak. The larger front-end/admin
library split is not in this branch, and I did not touch the other unqualified selectors in the
file (`.grouped-items …`, `.glb-form-type--radio …`); unlike Chosen, those classes only ever
exist on the authoring form.

### Measured, not inferred — both directions

`/blocks-for-visreg/events-calendar` turns out to be a real reproduction: a **public,
anonymous-visible** page that loads `views-basic.css` and renders three `chosen-enable`
multi-selects through `EventCalendarFilterForm` — precisely the path you described. I measured
`getComputedStyle(…).borderRadius` on all three widgets with the fix stashed in and out:

| | Front end (anonymous) | LB block-configure form |
|---|---|---|
| Before (leaking) | `0px` | `0px` |
| After (scoped) | **`4px`** | `0px` |

So the leak is closed on the front end and the reset is **unchanged** in the admin UI, which is
the half worth checking twice. I also traced all six attach sites of
`ys_views_basic/ys_views_basic` to be sure nothing else lost the reset: the four authoring
widgets all render inside the configure-block form, and `ViewsWizardForm` renders radios and no
Chosen widget at all.

### A guard, so this is the last time it's found by eye

Added `ys_views_basic/tests/src/Unit/AdminCssScopingTest.php`. This is the second time the leak
has been caught by reading, and the file already carries a long comment asking people not to
unscope these rules — a comment doing a test's job. Nothing else covers it: `npm run
lint:styles` globs `**/*.scss` and never reads this `.css` file, which is a fair part of why it
survived.

The guard fails on any Chosen selector missing `.layout-builder-configure-block` or
`#layout-builder-modal`. It deliberately does **not** accept `.gin--dark-mode` as a scope, for
the reason your own comment block in the file gives.

A code-review pass caught that my first parser silently dropped anything nested in an at-rule —
a leak inside `@media` would have passed green, which is worse than no test. Fixed, and the
parser is now exercised by eight cases (bare rule, inside `@media`, inside `@supports`, one
unscoped member of a grouped selector, `.gin--dark-mode` alone, plus the negatives) so it can't
pass vacuously. Verified it still fails on the pre-fix stylesheet and passes on the fixed one.

### Two corrections to the comment's detail

Minor, and they don't change the fix — but they change what "fixed" means, so they're worth
stating:

1. The comment says the rule has **3 declarations**; it has **2** (`border`, `border-radius`).
2. Of those two, **only `border-radius` was actually leaking.** On the front end the widgets'
   `border` computes to `0px none` from another stylesheet regardless, so `border: unset` was
   inert there. The observable defect was squared-off corners on the exposed filters where the
   theme intends 4px — visible next to the search field, which kept its radius. It is a smaller
   visual bug than the wording implies, but a real one.

### Verification

- `lando composer code-sniff` (Drupal + DrupalPractice, the CI command): **exit 0**, clean.
- `lando composer lint:php`: **exit 0**.
- `ys_views_basic` Unit suite: **129 tests, 388 assertions, 0 failures**. The 2 errors are the
  pre-existing `ExposedFilterAccordionTest` `ContainerNotInitializedException`s, unchanged from
  the baseline I took before starting. Net +9 tests.
- `npm run lint:styles` reports 1536 errors, **all pre-existing and unrelated** — every one from
  a single generated artifact, `atomic/_yale-packages/tokens/build/scss/tokens.scss`, that only
  exists in a `gyst`'d checkout. This diff contains no SCSS.

### Testing steps

On `/blocks-for-visreg/events-calendar` **logged out**, the Category / Audience / Custom Vocab
filter boxes should have the same rounded corners as the Search field beside them. Then open
that block in Layout Builder → the "Include/Exclude content that uses the following tags or
categories" pickers should still be square-cornered, unchanged.

dblanken-yale:
## Decision: fix the Chosen chip's accessible name on this PR

Fixing it here rather than carrying it as a follow-up. It is a WCAG 2.1 AA 4.1.2 failure on a public control - one per selected term - and it should not ship without a name.

**Scope:** give `.search-choice-close` an accessible name in `yds-select.js`'s `updateChoiceSummary()`, which already has the term label to hand.

This PR is already at `needs work` for the unscoped `views-basic.css` rule, so this joins that round rather than adding one.

dblanken-yale:
## Review remediation — findings 1299-2 and 1299-3

**TL;DR:** the collision guard for `flattenBuiltElements()` landed in `413174333` — the narrow first step only, as the review scoped it. The positional-view-argument concern is recorded on the ADR ticket rather than as a parallel ticket. The two big items (the `#after_build` redesign, the 1,398-line god class) remain deferred and untouched. Side effect worth knowing: **two tests in this file were erroring and now pass.** Label untouched, no re-request of review.

<details>
<summary><b>1299-2 — the collision the convention was holding by hand</b></summary>

`flattenBuiltElements()` keys the flattened map by each element's **own** key and merges with `$flat += ...`. That is a union: the first occurrence wins, any later duplicate is dropped, and nothing is logged. Four `#after_build` callbacks relocate elements into new wrappers, so two elements anywhere in one flattened subtree that happen to share a key quietly become one.

It is not a theoretical loss. `massageFormValues()` reads `include_operator` and `exclude_operator` straight off that map with **no null-coalesce** (`ViewsBasicWidgetBase:440-441`), so a collision on those does not fail — it stores the wrong element's value.

`buildTermOperator()` already avoids exactly this by hand, giving each operator a globally unique key rather than a generic one, and its docblock says so. That is a convention held in a comment. The new test is what holds it: both flattened subtrees, every relocating callback applied in its registered order, asserting that flattening loses nothing.

**Two design points worth calling out, because a weaker version of each would have looked identical and tested less:**

1. **The fixture models `tag_filters_intro` as the container it actually is**, with its `heading` and `help` children — not as a markup leaf. Those two are the most collision-prone keys in either subtree, precisely because every *other* intro in the widget is uniquely keyed (`exposed_filter_options_intro`, `result_content_intro`) and these two are not. A leaf fixture would have left the only keys at real risk outside the guard. Mutation-checked: adding a second container with a `heading` child now fails, naming the key.

2. **The losslessness assertion compares the key list to `array_keys()` of the map, not the two counts.** A count matches under drift that swaps one key for another — if the test's walk descended into a wrapper production treats as a leaf, both sides report one key and the trees are entirely different. Both walks emit in the same order and `+=` preserves insertion order, so identity holds and is strictly stronger. Also mutation-checked: making the walk diverge from the production descent rule fails the identity assertion, where a count comparison passes.

**Explicitly not done, per the review:** keying the flattened map by path, the `#after_build` relocation redesign, and the 1,398-line `ViewsBasicWidgetBase` refactor. All three are deferred by the review document itself; only the collision test was in scope.
</details>

<details>
<summary><b>Side effect: two pre-existing errors in this test file are fixed</b></summary>

`groupFieldDisplayRow()` builds its intro with the global `t()` and concatenates the result, which forces `__toString()` on the `TranslatableMarkup` and so needs a container with `string_translation` on it. There was none, so **every test exercising that callback was erroring with `ContainerNotInitializedException` before reaching a single assertion** — `testFieldOptionsAndPreviewShareOneDisplayRow` and `testEntitySpecificFieldOptionsJoinResultContent`.

A `setUp()` adding the stub container fixes both. Core's `UnitTestCase::setUp()` calls `\Drupal::unsetContainer()` first thing, so this cannot leak into another test class and needs no `tearDown()`.

```
ys_views_basic Unit, before:  129 tests, 388 assertions, 2 errors
ys_views_basic Unit, after:   131 tests, 408 assertions, 0 errors
```
</details>

<details>
<summary><b>1299-3 — recorded on the ADR, not as a new ticket</b></summary>

The positional-view-argument concern is now a comment on **yalesites-org/YaleSites-Internal#1162**, the Views Rework ADR ticket that owns the design — rather than a parallel ticket competing with it: yalesites-org/YaleSites-Internal#1162#issuecomment-5671330825

Substance: the argument list is an undeclared positional contract shared by the widget, every plugin and every view, with nothing checking that the three agree — and they already do not (yalesites-org/YaleSites-Internal#1685, where Resources listings silently ignore the offset). The suggestion for the ADR is to decode once into a value object and have plugins read named properties, so positions live in exactly one place. The comment also notes that **#1532 already landed the interim mitigation** — all 13 reads converted to `viewArgumentIndex()`, order pinned by `ViewArgumentOrderTest` — so nothing is blocked on the decision.
</details>

<details>
<summary><b>Verification</b></summary>

- `lando composer code-sniff` — **exit 0**. (It initially failed: two new method names had consecutive capitals and tripped Drupal's lowerCamel rule. Renamed. Worth noting the full run caught it where a narrower per-file invocation would not have.)
- `ys_views_basic` Unit suite: **129 tests / 388 assertions / 2 errors → 131 tests / 408 assertions / 0 errors.** Baseline captured before any edit.
- Mutation-verified three ways: a duplicate `heading` fails naming the key; a duplicate `include_operator` proves the first element is the survivor and the lost value is a plausible one; and making the test's walk diverge from the production descent rule fails the identity assertion.
- No production code changed — the diff is one test file.

**atomic#466 and clt#649 need no change for this work-unit.** 466-1 (JS unlinted in atomic CI) is tracked on yalesites-org/YaleSites-Internal#1684 as recorded future work, and 649-1 was a labels observation, which is out of bounds for this pass.
</details>

miketullo95:
**TL;DR:** Tagging this `work in progress` and pulling `needs review`. This is the epic branch for the Views Block Rework, not a review unit yet.

Sub-task PRs are still merging into this branch and building the epic up. Reviewing it now would mean reviewing a moving target, so we're holding off until the last child lands.

When the epic is complete we'll swap `work in progress` back to `needs review` and do a full pass on the whole body of work. No action needed here in the meantime.

PR #1311, opened 2026-06-30
YISP-127: search resource authors and hide excluded categories

square360-george:
## What
Two changes to the resource view component (`ys_views_content_resources`) for YISP-127:

1. **Author names in search.** The resource full-text search now covers author names. Affiliated authors (`field_authors` → Profile nodes) are searched via a runtime relationship to the Profile title; non-affiliated authors (a multi-value `double_field`) via both name columns. Behind a new toggleable **Authors** option in Search Fields. Existing options — including the in-use **Journal/Publication Name** field — are unchanged.
CleanShot 2026-06-29 at 21 16 17

2. **Category exposed filter.** Any term used to exclude content is now dropped from the Category filter dropdown, since selecting an excluded category would always return zero results.
CleanShot 2026-06-29 at 21 17 11

## How
- **No schema change.** The `double_field` name columns are exposed to Views via `hook_views_data_alter`; the search handlers (relationship, fields, DISTINCT) are injected at runtime only when the Authors option is enabled, so normal listings keep their existing query.
- Search and category logic is extracted into pure helpers (`expandSearchFields`, `authorSearchHandlerDefinitions`, `reduceCategoryTermsForExposure`) for testability.

## Tests
Unit tests in `tests/src/Unit/ViewsContentResourcesManagerTest.php` (4 tests, `@group yalesites`). Verified against ISPS data locally: author search returns correct counts (affiliated + non-affiliated), excluded categories drop from the filter, and existing search options are unaffected.

## Note — stacked on #1310
This branch was cut from `YISP-124`, so the diff currently also includes that commit (#1310, Basic HTML on the Resource Description field). Please merge **after #1310**; once it lands, this PR will show only the YISP-127 commit. The two changes are otherwise independent.

🤖 Generated in collaboration with Claude Code


web/profiles/custom/yalesites_profile/modules/custom/ys_views_content_resources/ys_views_content_resources.module
```diff
@@ -58,6 +58,28 @@ function ys_views_content_resources_views_data_alter(array &$data) {
...
+  // resource search combine filter cannot match on last names. Registering
+  // both columns lets the search handlers in ViewsContentResourcesManager feed
+  // them to the combine filter.
+  if (isset($data['node__field_nonaffiliated_authors'])) {
+    $columns = [
+      'field_nonaffiliated_authors_first' => t('Non-affiliated author first name'),
+      'field_nonaffiliated_authors_second' => t('Non-affiliated author last name'),
+    ];
+    foreach ($columns as $column => $label) {
+      $data['node__field_nonaffiliated_authors'][$column] = [
```
dblanken-yale, line 73: This is a straight assignment, not a merge. Drupal core's `views_field_default_views_data()` already populates this exact key (`group`, `title short`, `argument`, `filter`, `sort`) for every double_field column before `hook_views_data_alter()` runs — this wipes all of that, leaving only `title`/`help`/`field`. Doesn't break anything today since `content_resources` doesn't use these columns as a filter/sort/argument yet, but whoever tries to add "sort by last name" later will find the handler silently missing. Might be worth merging into the existing entry instead of overwriting it.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_content_resources/src/ExposedTaxonomyFilterOptions.php
```diff
@@ -0,0 +1,151 @@
...
+   *   are ignored.
+   * @param int|null $parent_tid
+   *   Optional parent term id; when set, only its descendants are offered.
+   *
+   * @return bool
+   *   TRUE when the filter was constrained, FALSE when it was left untouched
+   *   (missing filter, no vocabulary on it, or nothing to constrain).
+   */
+  public function apply(array &$filters, string $filter_name, array $excluded_tids, ?int $parent_tid = NULL): bool {
+    if (empty($filters[$filter_name]['vid'])) {
```
dblanken-yale, line 69: New failure mode the old hardcoded-vocab code couldn't hit: if a Views UI re-save of `content_resources` ever drops `vid` off an exposed `taxonomy_index_tid` filter, this returns `FALSE` and term exclusion just stops working — no error, no log. Not necessarily asking for a change, just flagging it since it's a silent one.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_content_resources/src/ExposedTaxonomyFilterOptions.php
```diff
@@ -0,0 +1,151 @@
...
+      return FALSE;
+    }
+
+    $excluded_tids = array_map('intval', $excluded_tids);
+    $parent_tid = $parent_tid ?: NULL;
+
+    // Nothing constrains the filter: leave the full vocabulary available.
+    if ($parent_tid === NULL && !$excluded_tids) {
```
dblanken-yale, line 77: The class docblock says "the filter config is left untouched when nothing constrains it," but the early return above only checks whether `$excluded_tids` is empty globally, not whether any of those ids are actually in *this* filter's vocabulary. So excluding a Category term while Audience is also exposed still reloads the whole Audience vocab and flips `expose.reduce` to `TRUE` on it, even though nothing was actually removed from it. `testExcludedIdsOutsideVocabularyAreIgnored` asserts this exact shape, so maybe it's intentional — if so the docblock could use a tweak; if not, might be worth checking the vocab before doing the reload.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_content_resources/src/ViewsContentResourcesManager.php
```diff
@@ -178,16 +204,14 @@ public function setupView(&$view, $params = []) {
...
-        $vid = "resource_category";
-
-        // Limit the filter to specific terms if provided.
-        $filters[$category_filter_name]['value'] = $this->getChildTermsByParentId($paramsDecoded['category_included_terms'], $vid);
-        $filters[$category_filter_name]['limit'] = TRUE;
-        $filters[$category_filter_name]['expose']['reduce'] = TRUE;
-      }
+      // Offer only the included parent's children (when set) and never a term
+      // used to exclude content, which would always return zero results.
+      $this->exposedTaxonomyFilterOptions->apply(
```
dblanken-yale, line 209: This same `apply()` call happens again at line 237 (custom vocab) and inside the loop a few lines down (audience, academic year, discipline, areas of study, geographic areas) — up to 7 calls total, and each does a full `loadTree()` whenever `$excluded_terms` is non-empty, even for vocabularies the exclusion has nothing to do with. A block with several filters enabled and any category exclusion set ends up doing 6 wasted tree loads on every render. Not urgent, just noting the cost.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_content_resources/src/ViewsContentResourcesManager.php
```diff
@@ -157,6 +178,11 @@ public function setupView(&$view, $params = []) {
     // Retrieve the current filter options from the view's display settings.
     $filters = $view->getDisplay()->getOption('filters');
 
+    // Terms the editor used to exclude content. Every exposed taxonomy filter
+    // drops these from its options: a visitor selecting one would always get
+    // zero results.
+    $excluded_terms = ExposedTaxonomyFilterOptions::normalizeTermIds($paramsDecoded['filters']['terms_exclude'] ?? []);
```
dblanken-yale, line 184: `ExposedTaxonomyFilterOptions::normalizeTermIds()` and the existing `getTermId()` further down in this file both unwrap the same legacy `['target_id' => id]` shape, independently. Any reason not to have this call the existing helper instead of reimplementing it?

web/profiles/custom/yalesites_profile/modules/custom/ys_views_content_resources/src/ViewsContentResourcesManager.php
```diff
@@ -446,6 +478,147 @@ public function setupView(&$view, $params = []) {
...
+   *
+   * @param array $available
+   *   Available term options keyed by term id.
+   * @param array $excluded
+   *   Term ids to remove.
+   *
+   * @return array
+   *   The available terms with the excluded ids removed.
+   */
+  public function reduceCategoryTermsForExposure(array $available, array $excluded): array {
```
dblanken-yale, line 618: This and `getChildTermsByParentId()` further down don't seem to have any callers left outside their own unit tests — looks like `setupView()` goes through `ExposedTaxonomyFilterOptions` directly now instead. If that's right, probably safe to delete both rather than carry unused public methods forward.

web/profiles/custom/yalesites_profile/modules/custom/ys_views_content_resources/src/ViewsContentResourcesManager.php
```diff
@@ -446,6 +478,147 @@ public function setupView(&$view, $params = []) {
...
+   * unit tested. Affiliated authors (field_authors) reference Profile nodes, so
+   * a relationship to the Profile exposes its title. Non-affiliated authors
+   * (field_nonaffiliated_authors) are a multi-value double_field whose first
+   * and second name columns are both needed. Every field is excluded from
+   * display and exists only to feed the combine filter.
+   *
+   * @return array
+   *   An array with 'relationships' and 'fields' keys of Views handler config.
+   */
+  public function authorSearchHandlerDefinitions(): array {
```
dblanken-yale, line 525: This and `expandSearchFields()` below are the two methods that actually implement author search, but I don't see a test referencing either — only `ExposedTaxonomyFilterOptions` got unit tests in this PR. The description mentions 4 tests; might be worth double-checking those actually cover this path.

dblanken-yale (changes requested):
Went through this locally against the diff and Drupal core's Views internals — solid approach to making the non-affiliated author names searchable through the double_field columns and the Profile relationship. Left a few things below, most worth a look before merge, a couple are just open questions.

PR #1322, opened 2026-07-03
1350: Content Collections: Add ability to delete an entire collection

dblanken-yale:
### Description of work
- Adds a **Delete collection** action to the Content Collections admin overview (Content → Manage Content Collections) plus a permission-gated confirmation form.
- Deleting a collection **dismantles the book outline** (removes the collection grouping and navigation) while **keeping every page as standalone content** — no page node is deleted.
- **Behavior decision:** the ticket's acceptance criteria left open whether child pages are deleted along with the collection or kept. Per the decision on this work, pages are kept as standalone content (nothing is destroyed); editors who want a page gone can delete it individually.
- Pages are removed deepest-first so the underlying Book module never promotes child pages into new standalone collections (`BookManager::deleteFromBook()` only promotes a top-level page's children, and by the time the root is reached it has none).
- Restricted to users with the `administer book outlines` permission (Editor, Site Administrator, Platform Admin) — the same permission that already gates the Content Collections admin screens.
- The confirmation screen lists every page that will be kept, states the action cannot be undone, and avoids a heading-level skip on the confirm page (WCAG 2.1 AA).
- Kernel test coverage: the dismantle logic (keeps pages, no orphan collections, branching trees, isolation between separate collections), the confirm form (builds, lists pages, rejects non-root nodes, submits), and the overview delete link (targets the correct collection).

> Note: contrib Book's `BookManager::deleteFromBook()` emits pre-existing `Undefined array key "bid"` notices in some contexts. This is reproducible with the existing single-page "Remove from outline" action and with zero YaleSites code involved, so it is contrib behavior out of scope for this change; the delete outcome is correct regardless.

### Functional testing steps:
- [ ] As an Editor/Site Administrator, create a Content Collection: create a Page, set its **Content Collection** to "- Create a new collection -", save, then add two or three child Pages to that collection.
- [ ] Go to **Content → Manage Content Collections** and confirm the collection is listed with a **Delete collection** action.
- [ ] Click **Delete collection**; confirm the screen names the collection, lists the pages that will be kept, and warns the action cannot be undone.
- [ ] Confirm the deletion; verify you return to the Content Collections overview and the collection is gone.
- [ ] Verify every page still exists as a standalone Page with its title intact, and that **no** new one-page collections were created.
- [ ] Verify a Contributor (without `administer book outlines`) cannot reach the Content Collections overview or the delete action.
- [ ] Confirm no "title cannot be null" white screen occurs (regression check from #880).

References yalesites-org/YaleSites-Internal#1350


miketullo95 (changes requested):
copy changes on `/admin/structure/book/X/delete` - changing the admin page title and copy content before deleting the collection. 

### Delete the "Example Landing Page" collection?

Your pages won't be deleted. They'll stay published as standalone content — you'll just lose the collection's shared navigation and grouping. This can't be undone.

Pages that will remain as standalone content:

miketullo95 (approved):

PR #1327, opened 2026-07-05
929: Pre-built Form (Contact Us): Edit Icon Visible outside of Edit Layout and Content

dblanken-yale:
### Description of work
- Logged-in administrators/editors saw a contextual edit icon on Pre-Built Form (Contact Us) blocks when simply **viewing** a page — outside of "Edit Layout and Content" — and it linked straight to submission data via the Webform "Results" link.
- Root cause: the Webform module unconditionally attaches its `webform` contextual-links group (Test / Results / Build / Settings) to every rendered submission form, ungated by route or edit mode. This is separate from the block-level Layout Builder edit link, which YaleSites already preview-gates.
- Fix (in `ys_core`): a new `WebformContextualLinksSuppressor` plus a `#pre_render` appended in `ys_core_form_alter()` on webform submission forms, which removes that contextual-links group on every route except Layout Builder editing routes. The `#pre_render` runs after the Webform module's `#after_build` (which attaches the group) and before the contextual placeholder is built, so the icon is never rendered. The links are kept in "Edit Layout and Content"; everywhere they are removed they are either the leak itself or redundant with the webform admin's own local-task tabs. Server-side route permissions are unchanged — this removes the front-end affordance, not the access control.
- Added unit tests (mirroring the existing `Editoria11yRouteSuppressorTest`) covering the route decision and the pre-render removal.

### Functional testing steps:
- [ ] Log in as an administrator or editor.
- [ ] Visit a page containing a Pre-Built Form (Contact Us) block (e.g. `/contact-us`) in normal view (NOT "Edit Layout and Content").
- [ ] Confirm there is no contextual edit icon/pencil on the form and no Test/Results/Build/Settings menu — and the form still renders and can be submitted.
- [ ] Open the same page via "Edit Layout and Content" and confirm the block is still editable (block-level edit still works).
- [ ] Confirm the webform can still be managed from the Webform admin UI (its Results/Build/Settings tabs).

References yalesites-org/YaleSites-Internal#929

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/WebformContextualLinksSuppressor.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/WebformContextualLinksSuppressor.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace Drupal\ys_core;
+
+use Drupal\Core\Security\TrustedCallbackInterface;
+
+/**
+ * Decides when to strip the Webform module's contextual "edit" links.
+ *
+ * The Webform module unconditionally attaches a "webform" contextual-links
+ * group (Test / Results / Build / Settings) to every rendered submission form.
+ * On YaleSites that surfaces an edit icon — and a direct link to submission
+ * data — to anyone with "access contextual links" who merely views a page
+ * containing a Pre-Built Form block, outside of Layout Builder's "Edit Layout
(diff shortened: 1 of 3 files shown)


miketullo95 (approved):
pass functional test - can go into the post 10.6 release branch

PR #1328, opened 2026-07-05
1033: Allow aria-level attributes on heading tags within the CKEditor Source View.

dblanken-yale:
### Description of work
- Permit the `aria-level` attribute on heading tags (h2-h6) in the Basic HTML text format, so content editors can give a heading a logical outline level independent of its visual level via the CKEditor Source Editing view.
- `filter.format.basic_html`: added `aria-level` to the `allowed_html` attributes on each heading so the filter keeps it.
- `editor.editor.basic_html`: added `aria-level` to the CKEditor Source Editing `allowed_tags` for h2-h6 so CKEditor's General HTML Support does not strip it when leaving Source view.
- Added a kernel test (`BasicHtmlAriaLevelTest`) asserting `aria-level` survives the filter on every heading and that the editor and filter configs stay in sync.
- Scope note: Basic HTML does not include `<h1>` (Drupal reserves h1 for the page title and the heading plugin only enables h2-h6), so this covers h2-h6.

### Functional testing steps:
- [x] Edit content using the Basic HTML editor (for example, an Event's Description field).
- [x] Open the CKEditor Source Editing view (the `</>` Source button, under the toolbar overflow menu) and add a heading with an aria-level, e.g. `<h2 aria-level="3">Example heading</h2>`.
- [x] Leave the Source Editing view, then re-open it and confirm the `aria-level="3"` attribute is still present (not stripped).
- [x] Save the content and confirm the `aria-level` attribute persists on the rendered heading.
- [x] Confirm a non-permitted attribute (e.g. `data-foo="x"`) is still stripped when leaving Source view.

References yalesites-org/YaleSites-Internal#1033

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/editor.editor.basic_html.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/editor.editor.basic_html.yml
@@ -119,11 +119,11 @@ settings:
         - '<blockquote cite>'
         - '<ul type>'
         - '<ol type>'
-        - '<h2 id>'
-        - '<h3 id>'
-        - '<h4 id>'
-        - '<h5 id>'
-        - '<h6 id>'
+        - '<h2 id aria-level>'
+        - '<h3 id aria-level>'
+        - '<h4 id aria-level>'
+        - '<h5 id aria-level>'
+        - '<h6 id aria-level>'
         - '<a hreflang>'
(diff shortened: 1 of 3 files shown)


miketullo95 (approved):
passed functional testing - can be moved into post 10.6 release branch

PR #1332, opened 2026-07-07
1371: Remove LibCal embed support (unsupported, security vulnerability)

dblanken-yale:
### Description of work

LibCal is not a supported embed type on YaleSites and its embed source carried a known security vulnerability (the `LibCal` source injected author-supplied embed markup into `drupalSettings` and loaded remote scripts from `schedule.yale.edu`). Rather than patch an unsupported integration, this removes it entirely from the `ys_embed` module:

- Deletes the `LibCal` embed source plugin (`src/Plugin/EmbedSource/LibCal.php`, id `libcal`).
- Deletes `js/LibCal.js`, `css/LibCal.css`, and the `springshare-libcal.png` thumbnail.
- Deletes `ys_embed.libraries.yml` — its only library was `ys_embed.libcal`.
- Removes `ys_embed_preprocess_block()` from `ys_embed.module`; its entire body only attached the LibCal library/jQuery to libcal embeds.
- Adds a kernel test (`LibCalRemovalTest`) that asserts the `libcal` source is no longer discoverable or selectable, while a retained source (`instagram`) still is — a regression guard against re-introducing the vulnerable source.

Notes for reviewers:
- The historical update hook `ys_embed_update_10001()` (which copied the LibCal icon) is intentionally left in place — past update hooks are an immutable migration ledger, and `_copy_embed_file()` tolerates the now-missing source file (try/catch), so it cannot fatal. `ys_embed_install()` scans the icon directory dynamically, so fresh installs are unaffected.
- Existing content with `embed_source = 'libcal'` degrades gracefully: `EmbedSourceManager::loadPluginById()` returns the `broken` source for any unknown id, so such embeds render as broken rather than fataling. A platform-wide audit of existing LibCal content on live sites should happen before/after this ships (could not be checked from the local environment).

### Functional testing steps

1. Go to `/media/add/embed` and open "Learn about supported formats and options".
2. Confirm the "Adding Embedded Media" source list no longer includes **LibCal** (other sources — Bluesky, Instagram, Google Maps/Calendar, Localist, Power BI, etc. — remain).
3. Run the module tests: full `ys_embed` suite is green (23 unit + 3 new kernel = 26 tests).

### Testing

- `ys_embed` PHPUnit suite: `OK (26 tests, 39 assertions)` (was 23; +3 new kernel tests).
- `phpcs` (Drupal, DrupalPractice) clean on changed files.
- Visually verified in the editor UI that LibCal no longer appears in the embed source list.

References yalesites-org/YaleSites-Internal#1371

---
Created with AI


miketullo95 (approved):
approved - move into post 10.6 branch

PR #1342, opened 2026-07-08
1318: Block authoring form — tabs + two-column field display

dblanken-yale:
## [1318: Views Block: Live Mockup Preview in Authoring Form](https://github.com/yalesites-org/YaleSites-Internal/issues/1318)

Isolated, closeable PR for the block-authoring-form **tabs + two-column** rework (items #4 and #5 from the 2026-07-06 QA round on yalesites-org/yalesites-project#1299). Targets **`1318-views-rework`** (not `develop`) so it can be merged into that branch or **closed to discard the experiment** with zero impact on #1299.

### Description of work
- Convert the listing-block authoring form from stacked accordions to one horizontal tab bar: **General | Field display & filters | Tags, sorting & pinned | Display options**. A `hook_form_alter` + `#process` unites the outer block fields (heading, link, padding, admin title, reusable) with the widget's inner `details` groups via core `#group` + field_group `horizontal_tabs` (field_group can't reach inside a field widget, so it's done in code). Widget group keys and field `#parents` are unchanged, so saving is unaffected.
- Two-column **Field display & filters** tab: controls left, live preview right (CSS-only, create-form-scoped, stacks in a narrow dialog, Gin dark-mode safe).
- Declare the `field_group` dependency.

### Functional testing steps:
- [ ] Add a listing block ("Posts - Card Grid View") in Layout Builder — the config form shows one horizontal tab bar with General first/active.
- [ ] Confirm the four tab buttons are evenly spaced and do **not** overlap (fix for the ~20px overlap gin_lb's `.glb-claro-details !important` margin caused in the LB modal).
- [ ] General holds heading, link, padding, admin title, and the reusable toggle.
- [ ] Other tabs hold their controls (incl. Event Time Period under Display options for events); no stray/empty groups.
- [ ] The Field display tab shows fields left + preview right on a wide dialog, stacking on a narrow one; Gin dark mode OK.
- [ ] Save a block and confirm the listing renders with the chosen settings (params unchanged).
- [ ] Edit an existing listing block — tabs render (no preview/two-column on edit), saves correctly.

Known cosmetic follow-up: after the widget groups are tabbed out, the now-empty `group_user_selection` container renders as a stray empty `<div>` (harmless). A reusable "block form tabs" pattern was documented for applying this to other block forms later.

**Review rework (#1342) — verify the front-end fixes:**
- [ ] In Gin **dark mode**, open a listing block config form: the tab content area sits on the lighter gray form surface (not near-black), and the **General** tab label lines up with the fields' left edge below it.
- [ ] On **Field display & filters**, the live preview stays **light** (white card, dark text) even in dark mode, and the controls + preview form an even **50/50** split with the preview sitting in the right column (not floating pinned in the corner); the left controls have no large empty gap.
- [ ] No stray empty box remains below the tabbed groups; the block still saves, and for **event** blocks the Event Time Period control still appears under **Display options**.

References yalesites-org/YaleSites-Internal#1318

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/assets/css/views-basic-form-tabs.css
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/assets/css/views-basic-form-tabs.css
@@ -1,13 +1,15 @@
 /*
  * Listing block form tabs (#1299).
  *
- * Gin strips field_group's default horizontal-tab dividers/padding and applies
- * negative margins, so the tab labels collide (they read as one run of text).
- * Re-space them into clearly separated, padded tabs sitting on a shared
- * baseline. Loaded only on the listing block authoring forms (via the
- * ys_views_basic/listing-form-tabs library, attached in the tabs form_alter)
- * and scoped to the block configure form, so no other field_group tabs are
- * touched.
+ * In the Layout Builder modal, gin_lb tags each horizontal-tab button with
+ * .glb-claro-details, whose `margin: 0 -0.25em 0 -1.25em !important`
+ * (gin_lb/dist/css/gin_lb.css) pulls every tab ~20px left over its neighbour,
+ * so the labels collide (they read as one run of text). Re-space them into
+ * clearly separated, padded tabs sitting on a shared baseline; the margin reset
+ * needs !important to win against that rule. Loaded only on the listing block
+ * authoring forms (via the ys_views_basic/listing-form-tabs library, attached
+ * in the tabs form_alter) and scoped to the block configure form, so no other
+ * field_group tabs are touched.
  */
 .layout-builder-configure-block .horizontal-tabs-list {
   display: flex;
@@ -20,7 +22,7 @@
 
 .layout-builder-configure-block .horizontal-tabs-list .horizontal-tab-button {
   float: none;
-  margin: 0;
+  margin: 0 !important;
   padding: 0;
   border: 0;
   list-style: none;
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/assets/css/views-basic-form-tabs.css
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/assets/css/views-basic-form-tabs.css
@@ -34,3 +34,22 @@
   white-space: nowrap;
   text-decoration: none;
 }
+
+/*
+ * Align the first tab's label with the panel's left edge. The tab bar's left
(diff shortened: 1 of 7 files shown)


miketullo95 (changes requested):
Hey @dblanken-yale, thanks for splitting this out as its own PR, makes it a lot easier to review the tabs work on its own. We ran through it and found a few front-end issues we want fixed before this is ready to merge into `1318-views-rework`.

**Dark mode: tab panel background**
The tab content area (General tab, presumably the others too) renders on a near-black background in dark mode. It should match the lighter gray the rest of the form uses instead. `views-basic-form-tabs.css` only styles the tab bar itself (`.horizontal-tabs-list` and its children, lines 14-36), nothing in there overrides whatever default Claro/Gin background the panes wrapper is falling back to. Can we add a background-color rule for the panes wrapper in this same file?

**Tab bar / panel border alignment**
The left edge of the "General" tab label doesn't line up with the left border of the panel below it, there's a small stagger. Looks like it's coming from the `padding: 0.5rem 1rem` on `.horizontal-tab-button a` (views-basic-form-tabs.css:33) not matching the left inset the pane content uses below it. Can we get the first tab's text flush with the panel's left edge?

**Live preview shouldn't follow dark mode**
In the Field display & filters tab, the preview card is meant to represent the front-facing content, which is always rendered in light mode on the live site. Right now it inherits Gin's dark mode colors (see the doc comment at views-basic-preview.css:10-12, plus the `--gin-bg-app` / `--gin-bg-layer` bindings at line 20 and lines 44-45). Can we pin the preview card to light-mode colors regardless of the admin theme's mode, so it actually shows what the front end will look like?

**Two-column layout isn't a real 50/50 split**
Right now `grid-template-columns: minmax(0, 1fr) minmax(14rem, 18rem)` (views-basic-preview.css:180) gives the preview a narrow, capped-width column while the left column flexes to fill the rest. Combined with `position: sticky; top: 0;` pinning it to the very top (lines 198-205), the preview reads as floating in the top right instead of sitting in a proper right-hand column, especially clear when the left column is short (see the "Field display & filters" tab on shorter blocks). Can we make this an actual 50/50 split so the preview feels like part of the layout instead of pinned in the corner?

**One more from your own notes**: the stray empty `<div>` left behind by `group_user_selection` once its fields are tabbed out (ys_views_basic.module), called out as a known cosmetic issue in the PR description. Let's get that cleaned up before merge too.

None of this touches the plumbing (form_alter, #process, field parents), it's all CSS/visual. Let us know if any of these need more context.

Thanks! Mike

miketullo95 (approved):

PR #1355, opened 2026-07-13
1394: Hotfix: Campus Groups event sync is non-functional — CampusGroupUrl fatals with ArgumentCountError

dblanken-yale:
### Description of work
- `CampusGroupUrl::__construct()` now accepts and forwards the 5th `DataParserPluginManager $parserPluginManager` argument to migrate_plus's `Url::__construct()`, matching its required signature. This fixes the fatal `ArgumentCountError` that stopped the `campus_groups_url` source plugin from instantiating.
- Because the fatal was raised during migration source-plugin discovery, it had disabled Campus Groups event sync (cron and manual import) **and** aborted `drush migrate:status` for every migration on any site with `ys_campus_groups` enabled. Both now work.
- Added PHPUnit unit coverage for the source plugin (`CampusGroupUrlTest`): successful construction with a configured endpoint, injection of the `x-cg-api-secret` auth header when a key is resolved, and the key-lookup helper.

**Review-round fixes (@miketullo95 multidev findings):**
- **Sync-screen database warning:** the "Sync now" controller (`RunMigrations`) now strips the harmless, unrelated core `migrate_drupal` legacy-database warning ("No database connection configured for source plugin variable") that was decorating an otherwise-successful sync. The match is narrow to that one message on that one screen — all other errors, and every other migrate location (dashboard, drush, cron), are untouched. Covered by `RunMigrationsTest`.
- **`addcslashes(): Passing null` deprecation:** each taxonomy lookup step in `campus_groups_events.yml` (`field_tags`, `field_localist_event_type`, `field_category`, `field_localist_group`, `field_event_status`, `field_event_topics`) now leads with a `skip_on_empty` (`method: process`) guard, so an empty/missing source value no longer reaches `EntityLookup::query()` → `Connection::escapeLike(null)`. Behavior-preserving for imported content; also prevents a possible whole-row failure when `event_topics` is missing. Covered by `CampusGroupsEventsNullGuardTest`.

### Functional testing steps:
- [ ] On an environment with `ys_campus_groups` enabled, run `drush migrate:status` — it lists migrations (including `campus_groups_events`) with **no fatal** (previously threw `ArgumentCountError`).
- [ ] Trigger the Campus Groups sync (cron or the manual import) — the `campus_groups_events` migration runs instead of fataling.
- [ ] On the Campus Groups settings page, click **Sync now** — you get the "Synchronized N events." success message and **no** "Failed to connect to your database server… No database connection configured for source plugin variable" error beside it.
- [ ] Confirm events still import (`drush migrate:status campus_groups_events` shows the imported count), including events whose feed omits category / group / status / tags / topics — those fields are simply left unset, and no `addcslashes(): Passing null` / `escapeLike` deprecation is logged.
- [ ] Run the module's unit tests — 9 tests pass:
  `vendor/bin/phpunit -c /app/phpunit.xml --group ys_campus_groups web/profiles/custom/yalesites_profile/modules/custom/ys_campus_groups/tests`

### Notes
- **Base branch:** this PR targets **`master`** (hotfix), not `develop`. Note that `build_deploy_and_test.yml` has `branches-ignore: master`, so automated CI (unit tests, static analysis, multidev deploy) does not run on this PR; the review-round changes were verified locally (PHPUnit green, `phpcs` Drupal + DrupalPractice clean).
- **Test-file overlap with #1354 (expected, trivial reconcile):** open PR yalesites-org/yalesites-project#1354 also adds this same `CampusGroupUrlTest.php` (on `develop`) with the earlier "GAP" tests that deliberately lock in the current bug. Once this hotfix forward-merges to `develop`, #1354 should drop its version of that file in favor of this one — its own comments already say to delete the bug-locking test once the fix lands.

References yalesites-org/YaleSites-Internal#1394

---
Created with AI

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,3 +110,5 @@ lando xdebug-off           # Disable Xdebug
 - **Custom Modules**: Use `ys_` prefix (e.g., `ys_core`, `ys_themes`)
 - **Branch Strategy**: `develop` (primary), `master` (releases), `YALB-XXX-description` (features)
 - **Semantic Release**: Automated versioning on master branch using conventional commits
+
+
--- a/README.md
+++ b/README.md
@@ -110,5 +110,3 @@ lando xdebug-off           # Disable Xdebug
 - **Custom Modules**: Use `ys_` prefix (e.g., `ys_core`, `ys_themes`)
 - **Branch Strategy**: `develop` (primary), `master` (releases), `YALB-XXX-description` (features)
 - **Semantic Release**: Automated versioning on master branch using conventional commits
-
-
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_campus_groups/migrations/campus_groups_events.yml
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_campus_groups/migrations/campus_groups_events.yml
@@ -119,44 +119,67 @@ process:
      from_format: 'Y-m-d\TH:i:s.0000000P'
      to_format: 'U'
   field_tags:
-    plugin: entity_generate
-    entity_type: taxonomy_term
-    value_key: name
-    bundle_key: vid
-    bundle: tags
-    source: event_tags
+    -
+      plugin: skip_on_empty
+      method: process
+      source: event_tags
+    -
+      plugin: entity_generate
+      entity_type: taxonomy_term
+      value_key: name
+      bundle_key: vid
+      bundle: tags
   field_localist_event_type:
-    plugin: entity_generate
-    entity_type: taxonomy_term
-    value_key: name
-    bundle_key: vid
-    bundle: localist_event_type
-    source: event_type
+    -
+      plugin: skip_on_empty
+      method: process
+      source: event_type
+    -
+      plugin: entity_generate
+      entity_type: taxonomy_term
+      value_key: name
+      bundle_key: vid
+      bundle: localist_event_type
   field_category:
-    plugin: entity_generate
-    entity_type: taxonomy_term
-    value_key: name
-    bundle_key: vid
-    bundle: event_category
-    source: event_category
+    -
+      plugin: skip_on_empty
+      method: process
+      source: event_category
+    -
+      plugin: entity_generate
+      entity_type: taxonomy_term
+      value_key: name
+      bundle_key: vid
+      bundle: event_category
   field_localist_group:
-    plugin: entity_generate
-    entity_type: taxonomy_term
-    value_key: name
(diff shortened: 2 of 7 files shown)


miketullo95 (changes requested):
Thanks for the quick turnaround on this one, @dblanken-yale! The constructor fix itself is right: it matches what migrate_plus's `Url::__construct()` expects, and the unit tests cover the cases that matter (successful construction, API key header injection, key lookup). Since `CampusGroupUrl` doesn't override `create()`, there's nothing else needed there.

That said, we found a couple things testing on the multidev that need a look before this merges:

**1. New DB error surfaced during "Sync now"**

On the Poorvus Campus Groups instance (`/admin/yalesites/campus_groups/sync`), clicking "Sync now" shows a "Synchronized 8 events" success message right alongside this error: "Failed to connect to your database server. The server reports the following message: No database connection configured for source plugin variable." Neither `campus_groups_taxonomy.yml` nor `campus_groups_events.yml` declares a `variable` source plugin, so it's not obvious where this is coming from. Can you track down where it's firing and confirm whether it's actually blocking anything or just noise?

**2. Deprecation notice in the logs from that same sync run**
```

Deprecated function: addcslashes(): Passing null to parameter #1 ($string) of type string is deprecated in Connection->escapeLike() (line 1392 of Connection.php)

```

The trace runs through `EntityLookup->query()`, called from `MigrateExecutable->processPipeline()`, called from `CampusGroupsManager->runMigration()` (`CampusGroupsManager.php:141`), called from `runAllMigrations()` (line 101), triggered by `RunMigrations->runAllMigrations()`. That points at one of the `entity_lookup` / `entity_generate` steps in `campus_groups_events.yml` (`event_category`, `event_group`, `event_status`, `event_tags`, or the `event_sources` lookup) getting a null value from the source XML. It's just a deprecation today, but PHP is heading toward making that a hard TypeError, so it's worth a null check before these values hit the lookup.

**3. Base branch mismatch (needs your confirmation)**

The PR description says this targets `master` as a hotfix, but the actual base branch is `develop`. That matters because `build_deploy_and_test.yml` has `branches-ignore: master`, so if this got retargeted to `master` like the description says, none of our CI (unit tests, static analysis, multidev deploy) would run on it at all. Can you confirm which base branch you actually want and update the PR description to match whichever it ends up being?

Nice test coverage on the constructor fix itself, that part's solid.

miketullo95 (approved):
re-ran the sync and didnt get any errors! looks good to me!

PR #1374, opened 2026-07-17
998: Optimize Gallery Block Image Loading and File Size Limits

dblanken-yale:
### Description of work
- Add two gallery-only media view modes and displays so gallery images load lazily at sized derivatives, without changing the shared card or other view modes:
  - `gallery_modal` — the lightbox image now uses the `image_max_width` responsive style (uncropped, up to 2400px, `sizes: 100vw`) loaded lazily. Previously the modal rendered each gallery item's **original full-size file eagerly** into the hidden DOM on page load (via the `full` -> `media.image.default` fallback), so every gallery downloaded all originals up front. Repointed the `gallery_modal_media` paragraph display at this view mode.
  - `gallery_thumbnail` — gallery grid thumbnails keep the `card_secondary_3_2` responsive style but load lazily (`eager` -> `lazy`).
- Cap image uploads at **20 MB platform-wide** on the shared `field_media_image` (was the PHP default, ~100 MB), and add editor help text guiding authors to compress/resize large photos before uploading. Platform-wide scope was confirmed with the team: a gallery-only cap is not cleanly achievable because media entities are shared across the platform.
- **Single work unit with the companion atomic PR** — atomic's twig switches the gallery thumbnail to the `gallery_thumbnail` view mode defined here, so the two must be reviewed and merged together.
- Other work completed in: yalesites-org/atomic#490, yalesites-org/component-library-twig#668

### Functional testing steps:
- [ ] On a page with a Gallery block, inspect a gallery thumbnail `` — `loading="lazy"` with `3_2_*` derivatives; the grid looks unchanged from before.
- [ ] Open a gallery image in the lightbox — the modal image is a sized `max_width_*` derivative (not the original file) and loads on open (lazy). On a wide viewport it picks a candidate matching the display width (sharp, not upscaled).
- [ ] View page source before opening any modal — gallery modal images are not eager-loaded on initial page load.
- [ ] (component-library-twig#668) On a fresh gallery page load, open DevTools Network and confirm the modal `max_width_*` derivatives are **not** requested until you open the lightbox (previously they all loaded ~4.5s after page load). The modal still fades in/out, and prev/next/close work.
- [ ] (component-library-twig#668) Open a gallery item with a caption longer than 100 characters — the caption heading/area renders at full height and the "+" toggle expands the full text (not clipped to zero height).
- [ ] Upload an image larger than 20 MB (e.g. via the Media Library) anywhere — it is rejected with Drupal's file-size error; the field help text advises compressing/resizing.
- [ ] Post cards and other components sharing `card_secondary_3_2` are unchanged (still eager, same style).

References yalesites-org/YaleSites-Internal#998

---
Created with AI


miketullo95 (changes requested):
@dblanken-yale we tested this on the pr-1374 multidev and most of the checklist checks out:

- Gallery thumbnails: lazy, correct `3_2_640` derivative, grid looks unchanged. Pass.
- Modal image: sized `max_width_1600` derivative instead of the original file. Pass.
- 20 MB upload cap and help text: confirmed working.
- Post cards / other views sharing `card_secondary_3_2`: still eager, unchanged. Pass.

One blocker though. The modal images aren't actually deferring like the ticket asks. We loaded the gallery page fresh and never opened a modal, but all three `max_width_1600` modal images fetched about 4.5s after page load anyway.

Root cause looks like `.media-grid__modal` (the modal wrapper) is hidden with `visibility: hidden` instead of `display: none`. Native `loading="lazy"` skips elements that are removed from layout, but `visibility: hidden` still occupies its spot on the page, so the browser treats it as on screen and fetches it right away regardless of the lazy attribute we're now setting at [core.entity_view_display.media.image.gallery_modal.yml:24](https://github.com/yalesites-org/yalesites-project/blob/af8c3dfca7ebba2695202cb3e77c771be7b9e67e/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.media.image.gallery_modal.yml#L24).

So the config change here is correct, but it can't deliver the "don't eager-load originals" goal until the modal's hidden state uses something the browser will actually respect for lazy loading (`display: none`, `content-visibility`, or swapping in a real `src` on open via JS). That's likely a change on the atomic side (yalesites-org/atomic#490) rather than this repo, since this PR is config-only. Can we get that piece sorted before merging, since it's the main point of the ticket?

Everything else looks good to merge once that's resolved.

miketullo95 (approved):
Re-tested this on the pr-1374 multidev now that the fix has landed. Confirmed:

- Fresh page load, no modal opened: zero `max_width_*` requests fire, and all modal `` elements read `naturalWidth: 0`. `.media-grid__modal` computes to `display: none` while inactive (not `visibility: hidden` anymore).
- Opening the lightbox fetches the `max_width_2400` derivative on demand and renders it correctly.
- Prev/next navigation moves to the correct item with the correct derivative.
- Close fades out and returns cleanly to `display: none`.
- 20 MB upload cap, help text, and thumbnail behavior (still `3_2_*`, still eager on post cards/other views) are unchanged from the last pass.

This resolves the blocker I flagged last round. Approving. component-library-twig#668 and atomic#490 approved separately since this is one work unit across three repos.

@dblanken-yale

miketullo95 (approved):
Re-tested this on the pr-1374 multidev now that the fix has landed. Confirmed:

- Fresh page load, no modal opened: zero `max_width_*` requests fire, and all modal `` elements read `naturalWidth: 0`. `.media-grid__modal` computes to `display: none` while inactive (not `visibility: hidden` anymore).
- Opening the lightbox fetches the `max_width_2400` derivative on demand and renders it correctly.
- Prev/next navigation moves to the correct item with the correct derivative.
- Close fades out and returns cleanly to `display: none`.
- 20 MB upload cap, help text, and thumbnail behavior (still `3_2_*`, still eager on post cards/other views) are unchanged from the last pass.

This resolves the blocker I flagged last round. Approving. component-library-twig#668 and atomic#490 approved separately since this is one work unit across three repos.

@dblanken-yale

PR #1095, opened 2025-09-23
YSP-972: AI: System instructions review and updating

dblanken-yale:
### Description of work
  This PR implements a comprehensive system instructions management module for YaleSites AI that allows administrators to review, update, and manage AI system instructions through the Drupal admin interface with full version control and
   external API synchronization.

  **Key features implemented:**
  - **Complete Drupal module** (`ys_ai_system_instructions`) as a submodule of `ys_ai`
  - **Main management interface** with real-time character counting, AJAX-based loading states, and manual sync capabilities
  - **Bi-directional API integration** for syncing instructions with external AI systems (both push and pull)
  - **Full version control system** with automatic version numbering, author tracking, and rollback capabilities
  - **Settings interface** for API configuration (endpoint, web app name, key selection) and content length controls
  - **Access control** with dedicated permissions and configuration-based feature toggles
  - **Integration with ys_integrations** module providing dashboard card with quick links and status display
  - **Markdown text formatting** support via league/commonmark dependency
  - **Comprehensive error handling** with graceful degradation when API is unavailable

  **Technical implementation:**
  - Added `league/commonmark` composer dependency for markdown processing
  - Created dedicated service classes:
    - `SystemInstructionsManagerService` - Orchestrates all operations with 10-second sync cooldown
    - `SystemInstructionsApiService` - Handles external API communication via Guzzle
    - `SystemInstructionsStorageService` - Database version control with transactions
    - `TextFormatDetectionService` - Markdown formatting utilities
  - Implemented Drupal routing, forms (edit, settings, revert), and controllers (version history, view)
  - Custom access checker validating permissions, feature enablement, and API configuration
  - JavaScript for real-time character counting with remaining character display
  - Comprehensive CSS styling integrated with Gin admin theme
  - Database schema (`ys_ai_system_instructions` table) with proper indexing for version storage
  - Complete logging and audit trail for all operations
  - Integration plugin displaying status card on AI integrations dashboard

  ### Functional testing steps:

  #### **Basic Navigation & Access**
  - [ ] Navigate to `/admin/config/ys-ai/system-instructions` and verify access is granted (user has "manage ys ai system instructions" permission)
  - [ ] Confirm page initially shows loading spinner with "Syncing system instructions, please wait..." message
  - [ ] Verify page auto-refreshes after ~1 second and displays the full form with current instructions
  - [ ] Check that current version number and total version count are displayed at top of form
  - [ ] Verify "View version history" link is present and functional

  #### **Editing & Saving Instructions**
[...]

PR #1375, opened 2026-07-17
1095: Remove unused metatag fields in Posts and Resources

dblanken-yale:
### Description of work
- Hide the metatag groups on Posts and Resources that Pages do not expose, so the Manage Settings > Metadata section shows only **Basic Tags** and **AI Meta data** on all three content types.
- Root cause: the `metatag_firehose` widget config is identical across bundles; the difference lived in `metatag.settings` `entity_type_groups`, which restricts which metatag groups appear per bundle. `post` listed `basic + ai_engine + open_graph` (the extra Open Graph group); `resource` was **absent entirely**, so metatag showed **all** groups. This is in exported config — not prod-only drift.
- Fix (config only): `post` — remove `open_graph`; `resource` — add an explicit `basic + ai_engine` restriction. Both now match the minimal set Pages already show.
- **Data note:** hiding a metatag group makes any already-stored value in that group uneditable via the UI (the value is still output). The local database had **no** Post or Resource nodes with populated metatag values, so nothing is orphaned locally — but please confirm against production before merge that no Posts/Resources rely on per-node Open Graph (or other hidden-group) values.
- Knowledge-base / user-guide update (if it references the fuller Metadata field set) is a post-merge content task.

### Functional testing steps:
- [ ] Edit a Post — the Metadata section shows only Basic Tags and AI Meta data (no Open Graph or other groups).
- [ ] Edit a Resource — same minimal set (previously it showed all metatag groups).
- [ ] Edit a Page — unchanged (still Basic Tags + AI Meta data).
- [ ] Confirm existing Posts/Resources still render their meta tags on the front end (stored values are unaffected).

References yalesites-org/YaleSites-Internal#1095

---
Created with AI


miketullo95 (changes requested):
@dblanken-yale The config fix matches most of #1095's acceptance criteria (Post/Resource now trim down from the full metatag group list), and the root-cause writeup on the config-vs-drift question is solid.

**Change requested:** on multidev, the unused groups are correctly gone, but the **AI Meta data** group is also missing on Post/Resource now. The scope here was to keep Basic Tags + AI Meta data and only remove the extras (Open Graph, etc.) — AI Meta data shouldn't have gone away. Can you take a look at why the `ai_engine` entry in metatag.settings.yml isn't rendering for these bundles?

Low priority, non-blocking: still worth confirming against production at some point that no live Posts/Resources have a populated value in a group being hidden (Open Graph on Post, or any of the groups Resource previously showed) — I'm fairly confident these fields aren't in real use, but better to be safe before hiding them for good.

Note: the knowledge-base/user-guide update (AC #5) will be handled separately as a content task, not blocking this PR.

miketullo95 (approved):
@dblanken-yale Confirmed on multidev — Basic Tags and AI Meta data are both showing correctly on Post and Resource now, and the extra groups (Open Graph, etc.) are gone. That resolves the AI Meta data regression I flagged in my last pass. Approving.

Still worth confirming against production at some point that no live Posts/Resources have a populated value in a group being hidden (Open Graph on Post, or any previously-shown Resource groups) — low priority, non-blocking, doesn't need to hold up merge.

PR #1376, opened 2026-07-18
1412: Add settings icon to YaleSites admin toolbar

dblanken-yale:
### Description of work
- The top-level **Settings** item in the admin toolbar rendered a blank square instead of an icon. It points at the custom `ys_core.admin_yalesites` route, and Gin only ships icons for standard routes, so its toolbar icon fell back to a blank sprite (`--icon: sprite.svg#fallback-view`).
- Vendored the FontAwesome Free 6.7.2 `gear` (solid) SVG (CC BY 4.0) into `ys_toolbar/images/gear.svg` — the same approach `ys_core` already uses for the Dashboard `gauge` icon.
- Added one CSS rule in `ys_toolbar/css/ys_toolbar.theme.css` overriding Gin's `--icon` custom property for the Settings item. Using `--icon` (rather than a `mask-image` override) makes it work in both horizontal and vertical toolbar orientations. A root-relative URL is used because a relative `url()` inside a custom property resolves against Gin's stylesheet, not this file.
- Scoped to the Settings item only. The file has pre-existing repo-wide formatting debt (tracked separately as #1404) that is intentionally left untouched here; the added lines are lint-clean.

### Functional testing steps:
- [ ] In the admin toolbar, the **Settings** item shows a gear icon (no longer a blank square), matching the other top-level items.
- [ ] Collapse the toolbar to vertical (narrow viewport / toolbar toggle) — the gear still shows.
- [ ] Visit a Settings page so the item is active — the gear still shows.
- [ ] Switch to Gin dark mode — the gear remains visible (follows the theme color).

References yalesites-org/YaleSites-Internal#1412

---
Created with AI


miketullo95 (approved):
@dblanken-yale Approving — this is a clean, scoped fix.

Traced the CSS mechanism against the rest of ys_toolbar.theme.css: setting --icon on .toolbar-icon.toolbar-icon-ys-core-admin-yalesites is consistent with Gin's own --icon/mask-image pattern (.toolbar .toolbar-bar .toolbar-icon:before { mask-image: var(--icon) }), so this should be a stable approach rather than a one-off hack.

On the icon style: the linked issue points at FA's "regular" gear, but the vendored icon is "solid" (matching the existing Dashboard gauge icon convention). That's fine — FA's free tier doesn't ship a free regular-weight gear anyway, so solid is the right call.

Manually verified the gear renders correctly in both light and dark mode, and in desktop and mobile view. Good to merge.

PR #1378, opened 2026-07-18
1396: ys_node_access unpublished-content access fixes (partial)

dblanken-yale:
## [1396: ys_node_access unpublished-content access fixes (partial)](https://github.com/yalesites-org/YaleSites-Internal/issues/1396)

> **Scope: PARTIAL.** This PR fixes only the **security-critical `ys_node_access` GAPs** from #1396 (3 of ~18 defects in that ticket). The remaining GAPs across 10 other modules are **not** in this PR — see "Remaining GAPs" below. The remaining GAPs are now filed as **per-module sub-issues of #1396** (#1419–#1429; see below) — each Bug / `tech-debt` / To Do — to be worked one at a time. This ticket is not fully resolved by this PR.

### Description of work
`ys_node_access_node_access_records()` conflated publish status with the module's CAS gating, causing two **confidentiality** issues:
- **Any authenticated user could view any other user's unpublished node.** Unpublished nodes were marked PRIVATE and every authenticated user holds the PRIVATE grant, so drafts were viewable regardless of ownership or the core `view own unpublished content` permission.
- **Unpublished nodes on a content type that never received `field_login_required` were granted a PUBLIC record**, so anonymous users could view them while unpublished.

Fixes:
- Unpublished nodes now get a `grant_view = 0` record, deferring to Drupal core's own unpublished-content access (owner + `view own/any unpublished content`). The empty grant keeps this realm's record set non-empty so core does not add its default public `all` grant, while granting no view access. Published nodes are gated purely on `field_login_required` (PRIVATE = authenticated/CAS only, PUBLIC = everyone), unchanged.
- Guarded `NodeAccessEventSubscriber::on403()` so a missing `cas` module (not a declared dependency) degrades to the default 403 instead of throwing `RouteNotFoundException`.
- Added `hook_update_10001()` calling `node_access_rebuild()` so the fix applies to **existing** content on deploy, not only to nodes saved afterward.
- Un-skipped the three paired GAP tests with real assertions; removed the obsolete characterization tests.

Verified: full `ys_node_access` suite **23 tests pass**; `phpcs` (Drupal + DrupalPractice) clean; security review clear.

### Functional testing steps:
- [ ] As a plain authenticated user (not the author, without `view own unpublished content`), you cannot view another user's **unpublished** node — 403.
- [ ] The node's **author**, with `view own unpublished content`, can still view their own unpublished node.
- [ ] An **anonymous** user cannot view an unpublished node on a content type without `field_login_required`.
- [ ] A **published** login-required page is still hidden from anonymous and shown to authenticated users (CAS gating unchanged); a published non-gated page is still public.
- [ ] After deploy, existing unpublished content is protected (the update hook rebuilds node access).

### Remaining GAPs from #1396 — now filed as per-module sub-issues
The other GAPs are **not** in this PR; each is now a per-module sub-issue of #1396 (type Bug, `tech-debt`, `To Do` on the board) so they can be worked one at a time. Each has a paired `...CurrentBehavior...` test to delete and a `...Should...` stub to un-skip/author:
- yalesites-org/YaleSites-Internal#1419 — **ys_toolbar**: off-canvas `#access` set to a permission *string* (always truthy) instead of a `hasPermission()` check.
- yalesites-org/YaleSites-Internal#1420 — **ys_dcn_field**: `"0"` identifier stripped by `empty()` + hidden by a truthiness check; also missing field-settings config schema (new file).
- yalesites-org/YaleSites-Internal#1421 — **ys_taxonomy_manager**: `type IN ()` query throws when no bundle references the vocabulary.
- yalesites-org/YaleSites-Internal#1422 — **ys_campus_groups**: `CampusGroupsConfig` reads `ys_campus_group.settings` (singular) vs the plural used everywhere else.
- yalesites-org/YaleSites-Internal#1423 — **ys_migrate**: `processImport()` silently drops rows where node creation returns NULL.
- yalesites-org/YaleSites-Internal#1424 — **ys_localist**: `getSourceData()` clobbers `$url` each paging iteration, compounding `&page=`.
- yalesites-org/YaleSites-Internal#1425 — **ys_embed** (5): missing `/s` regex modifiers (GoogleMaps, JWPlayer), `isVideo()` host false-positive, SoundCloud playlist endpoint, `loadPluginById()` NULL instead of Broken fallback.
- yalesites-org/YaleSites-Internal#1426 — **ys_views_basic**: missing `break;` (fall-through returns the wrong default).
- yalesites-org/YaleSites-Internal#1427 — **ys_layouts**: inverted bail condition renders event meta for every bundle.
- yalesites-org/YaleSites-Internal#1428 — **ys_views_content_resources**: uninitialized `$entityValue` for a new/unsaved block.
- yalesites-org/YaleSites-Internal#1429 — **ys_core**: dead `mailto` branch in `getUrlType()` (ordered after `isInternal()`).
- **ys_file_management** — the described GAP is **not a defect** (already fixed on develop, commit 8c099f0d7); no ticket.

References yalesites-org/YaleSites-Internal#1396

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/src/EventSubscriber/NodeAccessEventSubscriber.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/src/EventSubscriber/NodeAccessEventSubscriber.php
@@ -7,6 +7,7 @@ use Drupal\Core\Routing\TrustedRedirectResponse;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Url;
 use Symfony\Component\HttpKernel\Event\ExceptionEvent;
+use Symfony\Component\Routing\Exception\RouteNotFoundException;
 
 /**
  * Overrides the 403 response to perform a CAS redirect for specific pages.
@@ -53,7 +54,15 @@ class NodeAccessEventSubscriber extends HttpExceptionSubscriberBase {
          */
         if ($node->hasField('field_login_required')) {
           if ($node->isPublished() && $node->field_login_required->first()->getValue()['value']) {
-            $casRedirectUrl = Url::fromRoute('cas.login', ['destination' => $node->toUrl()->toString()])->toString();
+            try {
+              $casRedirectUrl = Url::fromRoute('cas.login', ['destination' => $node->toUrl()->toString()])->toString();
+            }
+            catch (RouteNotFoundException $e) {
+              // The 'cas' module is not installed (not a hard dependency), so
+              // there is no login route to redirect to. Leave the default 403
+              // response in place instead of fataling.
+              return;
+            }
             $returnResponse = new TrustedRedirectResponse($casRedirectUrl);
             $returnResponse->getCacheableMetadata()->setCacheMaxAge(0);
             $event->setResponse($returnResponse);
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/tests/src/Kernel/NodeAccessEventSubscriberMissingCasDependencyTest.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/tests/src/Kernel/NodeAccessEventSubscriberMissingCasDependencyTest.php
@@ -13,20 +13,17 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Event\ExceptionEvent;
 use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
-use Symfony\Component\Routing\Exception\RouteNotFoundException;
 
 /**
  * Tests NodeAccessEventSubscriber::on403() when the 'cas' module is absent.
  *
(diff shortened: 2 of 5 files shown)


miketullo95 (approved):
@dblanken-yale Reviewed against #1396. Traced the node_access_records() fix — the grant_view=0 placeholder for unpublished nodes correctly keeps this realm's record set non-empty (so core doesn't fall back to its default public 'all' grant) while handing unpublished-content view decisions back to core's own owner/"view own unpublished content" checks. That closes both the cross-user draft exposure and the missing-field public-while-unpublished gap. The on403() try/catch and the update_10001 rebuild hook are both correctly scoped to what #1396 described, and the 23 passing kernel/unit tests exercise real $node->access() calls for the exact scenarios (non-owner denied, owner+permission allowed, non-owner still denied even with the permission granted).

Approving. A couple of process notes, not blockers:
- Since this only fixes the ys_node_access GAPs and splits the other 10 modules into #1419–#1429, we'll keep #1396 open as the parent epic until those are done rather than closing it now.
- The node_access_rebuild() in the update hook re-walks every node's grants on deploy — worth a one-line mention in release notes/QA about a security/QA improvement, but that's an epic-level note, not something this PR needs to address.

PR #1380, opened 2026-07-18
1427: ys_layouts: EventMetaBlock renders event meta for every bundle (inverted bail condition)

dblanken-yale:
### Description of work
- Fixed an inverted bail condition in `EventMetaBlock::build()` (`ys_layouts`). Because any real `NodeInterface` makes the first OR operand true, the negation was always false, so the block rendered full event meta for **every** node bundle instead of only events. Reordered to `!($node instanceof NodeInterface) || $node->bundle() !== 'event'`, which bails for a null/non-node (short-circuiting before `->bundle()`) and for any non-event bundle.
- Un-skipped the paired GAP unit test `testBuildShouldReturnEmptyForNonEventNode` and removed the obsolete `testBuildRendersEventDataForNonEventNode` that characterized the buggy behavior.

### Functional testing steps:
- [ ] On an event page, confirm the event meta (dates, Add to Calendar, location, etc.) still renders as before.
- [ ] Place the Event Meta block on a non-event page (e.g. a Page) via Layout Builder, save, and view it; confirm no event meta renders.
- [ ] Confirm existing event listings and pages show no regressions.

References yalesites-org/YaleSites-Internal#1427

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Plugin/Block/EventMetaBlock.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Plugin/Block/EventMetaBlock.php
@@ -81,7 +81,7 @@ class EventMetaBlock extends BlockBase implements ContainerFactoryPluginInterfac
 
     /** @var \Drupal\node\NodeInterface $node */
     $node = $this->routeMatch->getParameter('node');
-    if (!($node instanceof NodeInterface || ($node && $node->bundle() !== 'event'))) {
+    if (!($node instanceof NodeInterface) || $node->bundle() !== 'event') {
       return [];
     }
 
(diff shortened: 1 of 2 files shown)


miketullo95 (approved):
@dblanken-yale Reviewed against #1427. The bail-condition fix matches the issue's expected behavior exactly, and I traced event_meta_block's placement — it's only shipped on the event bundle's default layout (core.entity_view_display.node.event.default.yml), locked in place, and layout_builder_restrictions' per-region allowlist keeps it out of the block picker for other bundles (e.g. Page's layout_onecol only allowlists ys_taxonomy_display_block under 'YaleSites Layouts'). So the fix is correct defensively even though the non-event path isn't reachable through the UI today.

Verified on multidev: an existing event page still renders its event meta correctly (no regression), and the block isn't offered on non-event pages. Approving.

PR #1382, opened 2026-07-18
1425: ys_embed: five embed defects (multiline regex, isVideo host check, SoundCloud playlist, Broken fallback)

dblanken-yale:
### Description of work
Fixes five independent defects in the `ys_embed` module. Each defect's paired GAP test (already on develop) is un-skipped and its characterization test removed.

- **GoogleMaps / JWPlayer multi-line iframes** — the iframe validation patterns rejected an embed code whose `src` attribute wraps onto its own line. GoogleMaps now uses `[^>]*` before `src` (matching the sibling GoogleCalendar plugin), which spans newlines within the opening tag but cannot cross into a following element; JWPlayer gains the `s` (DOTALL) modifier.
- **`isVideo()` host check** — an unanchored regex with an unescaped dot classified any URL merely containing `youtube.com` as a video embed. It now parses the URL host and matches it against the allowed video hosts (youtube.com / youtu.be / vimeo.com).
- **SoundCloud playlists** — `getUrl()` and the plugin template hardcoded the `/tracks/` API endpoint, ignoring the captured `track_or_playlist` group, so playlist embeds were rebuilt against the wrong endpoint. Both now use the captured group.
- **Broken fallback** — `EmbedSourceManager::loadPluginById()` read `$instances[BROKEN_ID]` before it was populated, returning `NULL` with an "Undefined array key" warning; it now reassigns to `BROKEN_ID` and lazily instantiates the `Broken` plugin per its docblock.

**Reviewer note (SoundCloud playlists — product question):** the `getUrl`/template fix corrects the rendered endpoint for playlist embeds, which benefits any already-saved/legacy playlist embed. However, a **pre-existing** validation gate — `EmbedConstraintValidator::isTrack()` — still rejects **newly-authored** SoundCloud playlist embeds with an `invalidAudioTrack` violation (this gate and its asserting test predate this PR). So new playlist embeds still cannot be saved. Whether SoundCloud **playlists** should be allowed at all is a product decision outside this ticket's five named defects; if desired, that's a follow-up to relax `isTrack()` and update its test. Flagging so the playlist behavior isn't misread as fully enabled.

### Functional testing steps:
- [ ] Paste a multi-line Google Maps "Embed a map" iframe (with `src` on its own line); confirm it validates and renders.
- [ ] Paste the JW Player documented example embed (multi-line); confirm it validates.
- [ ] Confirm a valid single-line Google Maps / SoundCloud track embed still works (no regression).
- [ ] Confirm an existing SoundCloud track embed still renders correctly.
- [ ] Paste a bare YouTube/Vimeo URL and confirm the "use the Video component" guidance still appears; confirm a non-video URL that merely contains "youtube.com" is no longer misclassified as a video.

References yalesites-org/YaleSites-Internal#1425

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_embed/src/Plugin/EmbedSource/GoogleMaps.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_embed/src/Plugin/EmbedSource/GoogleMaps.php
@@ -21,7 +21,7 @@ class GoogleMaps extends EmbedSourceBase implements EmbedSourceInterface {
   /**
    * {@inheritdoc}
    */
-  protected static $pattern = '/^<iframe.+src=\"https:\/\/www\.google\.com\/maps\/embed(?<map_params>\?.+?)\".*$/';
+  protected static $pattern = '/^<iframe[^>]*src=\"https:\/\/www\.google\.com\/maps\/embed(?<map_params>\?.+?)\".*$/s';
 
   /**
    * {@inheritdoc}
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_embed/src/Plugin/EmbedSource/JWPlayer.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_embed/src/Plugin/EmbedSource/JWPlayer.php
@@ -21,7 +21,7 @@ class JWPlayer extends EmbedSourceBase implements EmbedSourceInterface {
   /**
    * {@inheritdoc}
    */
-  protected static $pattern = '/<iframe.+src="(?<url>https:\/\/(content|cdn).(jwplatform|jwplayer).com\/players\/[^"]+)"/';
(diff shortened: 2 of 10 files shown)


miketullo95 (approved):
@dblanken-yale Reviewed against #1425. All five defects check out: verified isVideo() now anchors on parse_url() host rather than substring-matching, SoundCloud getUrl()/template both correctly use the captured track_or_playlist group, and traced loadPluginById()'s fallthrough in EmbedSourceManager.php to confirm it actually lazy-instantiates the Broken plugin now instead of reading an unpopulated array key. GoogleMaps/JWPlayer multi-line fixes both validated against the documented examples.

Approving. Filed two follow-ups spotted during review, neither blocking this PR:
- #1431 — SoundCloud playlist embeds still can't be newly authored (pre-existing isTrack() gate), separate from this PR's rendering fix for already-saved playlists.
- #1432 — unrelated pre-existing bug in EmbedSourceManager::loadAll() (returns undefined $this->instance instead of $this->instances).

PR #1383, opened 2026-07-18
1419: ys_toolbar: off-canvas Theme Settings button never checks the "yalesites manage settings" permission

dblanken-yale:
### Description of work
- `ToolbarItemsService::buildOffCanvasButton()` set the render array `#access` to the literal permission-name string `'yalesites manage settings'`. Drupal's renderer only denies rendering when `#access` is strictly `FALSE` or an `AccessResultInterface`, so a non-empty string is always truthy — the Theme Settings off-canvas link rendered for every user regardless of permission. It now evaluates `$this->account->hasPermission('yalesites manage settings')`.
- The destination route (`ys_themes.theme_settings`) already enforces the same permission server-side, so this restores the intended display-layer gating (defense in depth).
- Un-skips the paired GAP test (denied path) and adds a granted-path test; removes the characterization test that asserted the old always-truthy string.

**Reviewer note (pre-existing, follow-up):** `#cache.contexts` on both `buildButton()` and `buildOffCanvasButton()` lists only `url.path`. Now that the off-canvas button's `#access` varies by permission, those elements should also declare `user.permissions` for render-cache correctness. Left out of this PR — it's a pre-existing module-wide pattern (the sibling `buildButton()` already gates on route access with the same `url.path`-only cache) and there's no security bypass (the route enforces the permission server-side). Flagging as a follow-up.

### Functional testing steps:
- [ ] As a user WITHOUT the "yalesites manage settings" permission (e.g. a basic editor), confirm the Theme Settings button no longer appears in the off-canvas toolbar.
- [ ] As a user WITH the permission (site admin), confirm the Theme Settings button still appears and opens the off-canvas panel.
- [ ] Confirm the Manage Settings / Layout toolbar links (unchanged) still behave as before.

References yalesites-org/YaleSites-Internal#1419

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_toolbar/src/ToolbarItemsService.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_toolbar/src/ToolbarItemsService.php
@@ -294,7 +294,7 @@ class ToolbarItemsService {
           $route,
           $this->redirectDestination->getAsArray(),
         ),
-        '#access' => 'yalesites manage settings',
+        '#access' => $this->account->hasPermission('yalesites manage settings'),
         '#attributes' => [
           'class' => [
             'use-ajax',
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_toolbar/tests/src/Unit/ToolbarItemsServiceTest.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_toolbar/tests/src/Unit/ToolbarItemsServiceTest.php
@@ -368,40 +368,36 @@ class ToolbarItemsServiceTest extends UnitTestCase {
   }
 
   /**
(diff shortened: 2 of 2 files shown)


miketullo95 (approved):
@dblanken-yale Reviewed against #1419 — the #access fix matches the issue's expected behavior exactly, and I verified the cache-context follow-up you flagged is genuinely pre-existing (buildButton() has the same url.path-only pattern despite also gating on a permission-dependent check), so no objection to leaving that as a separate follow-up.

Also confirmed editor and contributor roles both already carry 'yalesites manage settings' directly in their role config, so this doesn't take Theme Settings access away from any role that currently has it — it only hides the button from roles that never had access to begin with (the destination route already enforced this server-side).

Verified on multidev on both an editor role and a platform admin role — functions correctly. Approving.

PR #1384, opened 2026-07-18
1426: ys_views_basic: missing break; in getDefaultParamValue() falls through for show_current_entity

dblanken-yale:
### Description of work
- `ViewsBasicManager::getDefaultParamValue()`'s `show_current_entity` switch case was missing a `break;`, so execution fell through into the adjacent `pinned_to_top` case and overwrote `$defaultParam`. `getDefaultParamValue('show_current_entity', ...)` therefore always returned the `pinned_to_top` value instead of its own — e.g. the "show current entity" checkbox default in the listing-block widget (`ViewsBasicDefaultWidget`) mirrored `pinned_to_top`. Added the `break;` so each case returns its own default.
- Un-skips the paired GAP test and removes the characterization test that asserted the fall-through behavior.

**Reviewer note:** the automated `/code-review` pass could not complete its verification stage (a session/rate limit was hit mid-run), so please re-run `/code-review` on this PR before merge. The fix is nonetheless a one-line, issue-prescribed `break;`, independently verified by TDD (red → green), the full `ys_views_basic` regression suite (71 tests, no regressions), `composer code-sniff` (exit 0), and `/simplify`.

### Functional testing steps:
- [ ] In a listing block (ys_views_basic), enable "Show current entity" and save; reopen the block form and confirm the checkbox reflects the saved value rather than the pinned-to-top value.
- [ ] Confirm "Pinned to top" and the other listing options still behave correctly.

References yalesites-org/YaleSites-Internal#1426

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_views_basic/src/ViewsBasicManager.php
@@ -781,6 +781,8 @@ class ViewsBasicManager extends ControllerBase implements ContainerInjectionInte
 
       case 'show_current_entity':
         $defaultParam = (empty($paramsDecoded['show_current_entity'])) ? 0 : $paramsDecoded['show_current_entity'];
+        break;
+
       case 'pinned_to_top':
         $defaultParam = (empty($paramsDecoded['pinned_to_top'])) ? FALSE : (bool) $paramsDecoded['pinned_to_top'];
         break;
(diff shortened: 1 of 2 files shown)


miketullo95 (approved):
@dblanken-yale Reviewed against #1426 — the break; is added at the exact line the issue cited, and the test changes (un-skipped GAP test + updated characterization test) match the issue's spec exactly. One-line fix, well-backed by TDD and the full 71-test regression suite.

Noted your reviewer note that the automated /code-review pass didn't finish (rate limit mid-run) — given how small and well-tested this fix is, approving without re-running it.

PR #1385, opened 2026-07-18
1421: ys_taxonomy_manager: AddTermsToNodesForm throws InvalidQueryException when no bundle references the vocabulary

dblanken-yale:
### Description of work
- `AddTermsToNodesForm::buildForm()` built a `type IN ()` node query when no node bundle references the selected vocabulary (`$relevantBundles` was empty), so `$query->execute()` threw `InvalidQueryException` (white screen) before the intended empty-result "No nodes found" message could be reached. It now only runs the query when at least one bundle references the vocabulary; otherwise it falls into the existing "No nodes found" message and returns the form.
- Un-skips the paired GAP test and removes the characterization test that expected the exception.

**Reviewer note (pre-existing, follow-up):** `buildForm()` dereferences its nullable `$taxonomy_vocabulary` parameter (`->id()`) with no null guard, so a NULL vocabulary would fatal. This is pre-existing and out of scope for this ticket; in practice it is gated by the route's entity upcasting (a missing vocabulary 404s before `buildForm()` runs). Flagging as a follow-up.

### Functional testing steps:
- [ ] Create a vocabulary that no content type references (no term-reference field targets it), open its "Add terms to nodes" form, and confirm it shows "No nodes found with fields that reference the selected vocabulary." instead of a white-screen error.
- [ ] For a vocabulary that IS referenced by a content type, confirm the form still lists terms and matching nodes as before.

References yalesites-org/YaleSites-Internal#1421

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_taxonomy_manager/src/Form/AddTermsToNodesForm.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_taxonomy_manager/src/Form/AddTermsToNodesForm.php
@@ -107,13 +107,18 @@ class AddTermsToNodesForm extends FormBase {
       }
     }
 
-    // Get nodes of the relevant bundles.
-    $query = $this->nodeStorage->getQuery()
-      ->condition('status', 1)
-      ->condition('type', array_unique($relevantBundles), 'IN')
-      ->accessCheck(TRUE);
+    // Get nodes of the relevant bundles. Only query when a bundle references
+    // the vocabulary; an empty $relevantBundles would build an invalid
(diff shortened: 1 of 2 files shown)


miketullo95 (approved):
@dblanken-yale Reviewed against #1421 — the empty-bundles guard matches the issue's expected fix exactly, and the test coverage here is Kernel-level (real Vocabulary/query APIs, not mocks), which gives more confidence than a Unit-mocked test would. Verified the pre-existing nullable $taxonomy_vocabulary note against the routing.yml — confirmed the entity:taxonomy_vocabulary param converter 404s before buildForm() runs, so that's correctly out of scope here.

Verified on multidev (as user 1) — confirms the fix. Approving.

PR #1386, opened 2026-07-18
1420: ys_dcn_field: a literal "0" identifier is stripped and never rendered; add field-settings config schema

dblanken-yale:
### Description of work
Fixes three related mishandlings of a literal `"0"` DCN identifier in `ys_dcn_field`:
- **Data loss on save** — `DcnFieldItem::isEmpty()` used `empty()`, so a `"0"` identifier was reported empty and dropped by empty-item filtering on save. It now treats only `NULL`/`''` as empty, so a `"0"` is preserved.
- **Never rendered** — `DcnFieldDefaultFormatter::viewElements()` used a truthiness check on `$dcn_identifier`, so a `"0"` never rendered. It now renders when the identifier is not `NULL` and not `''`.
- **Missing config schema** — added `config/schema/ys_dcn_field.schema.yml` covering the field, formatter, and widget settings, and removed the `$strictConfigSchema = FALSE` opt-outs from all three kernel tests so their config is validated strictly (verified: the suite passes with strict schema on).

Un-skips the paired GAP tests and removes the characterization tests.

**Resource-page render fix (added per review):** the *same* truthiness bug lived in the sibling render path `ResourceMetaBlock::build()` (`ys_layouts/src/Plugin/Block/ResourceMetaBlock.php`), which is the actual resource-page (Layout Builder) render path — so a `"0"` DCN saved correctly but still did not *display* on resource pages. Per @miketullo95's review (fix here rather than split to a follow-up), the same guard is now applied there: render when the identifier is not `NULL` and not `''`. This module is not cited by #1420 but is the editor-facing symptom, so it is included here.

### Functional testing steps:
- [ ] On a resource node, set a DCN Type and enter `0` as the DCN Identifier; save, then reopen the edit form and confirm the `0` is still stored (previously it was stripped).
- [ ] Where the DCN field is rendered via its own formatter, confirm a `0` identifier now displays.
- [ ] On the **resource page itself** (the resource meta block, Layout Builder render path), confirm a `0` DCN Identifier now displays — previously it saved but was dropped from the meta block. (This is the fix added per review in `ys_layouts` ResourceMetaBlock.)
- [ ] Confirm a DCN with a normal (non-zero) identifier still saves and displays exactly as before (no regression).

References yalesites-org/YaleSites-Internal#1420

---
Created with AI


miketullo95 (changes requested):
Requesting changes before this merges.

Tested on the multidev environment. Confirmed the DCN Identifier field now saves and persists a literal `0` (the `DcnFieldItem::isEmpty()` fix works as expected). But once we set a real `0` in the DCN Identifier field and checked the actual resource page, the DCN disappeared from the meta block entirely.

That's because `ys_layouts/src/Plugin/Block/ResourceMetaBlock.php:262` still has the old truthy check:

```php
if ($dcn_type && $dcn_identifier) {
```

Since `dcn_identifier` is the literal string `"0"`, that check treats it as empty and drops the item, so it never renders on the page editors and visitors actually see.

We know the PR description flags this as a known gap with a recommended follow-up ticket, but from an editor's perspective this is the actual symptom in issue #1420 (a `0` DCN not showing up on a resource page), so we'd like it fixed here rather than split off. Can you apply the same fix as `DcnFieldDefaultFormatter::viewElements()` to that check:

```php
if ($dcn_type && $dcn_identifier !== NULL && $dcn_identifier !== '') {
```

Once that's in, we're good to re-test and approve.

@dblanken-yale

miketullo95 (approved):
Approved. We re-tested on the multidev environment and can confirm the DCN field and its value now show correctly on the resource page, including the ResourceMetaBlock fix requested in the last pass. All three truthy-check fixes match the issue, the config schema looks complete, and the unskipped tests cover the "0" identifier case. Nice work.

@dblanken-yale

PR #1387, opened 2026-07-18
1428: ys_views_content_resources: uninitialized $entityValue errors for a new/unsaved block

dblanken-yale:
### Description of work
- `ViewsContentResourcesDefaultWidget::formElement()` assigned `$entityValue` only inside `if (!empty($decodedParams['filters']['types'][0]))` but read it unconditionally when calling `getFormSelectors()`, producing an "Undefined variable $entityValue" warning for a brand-new/unsaved block with empty params. It now initializes `$entityValue = NULL;` before the guard.
- The same new-block path also passed `NULL` params to `json_decode()`, emitting a "Passing null to parameter #1" deprecation (confirmed in the test run). Guarded it with `?? ''` so a new block builds cleanly (no warning, no deprecation).
- Un-skips the paired GAP test and removes the characterization test.

**Reviewer notes (follow-ups, not fixed here):**
- **Dead plumbing:** `$entityValue` is forwarded as `getFormSelectors()`'s third argument, but that method never reads its third parameter — so the value has no effect. The issue explicitly prescribed the `$entityValue = NULL` initialization, so this PR keeps that minimal fix; a cleaner future refactor would drop the unused parameter and the `$entityValue` computation (touches the shared `ViewsContentResourcesManager`).
- **Pre-existing null-deref:** `formElement()` calls `->load('custom_vocab')->label()` with no null check; if that vocabulary is ever missing it would fatal the block edit form. Out of scope for this ticket (unchanged line), low likelihood (custom_vocab is standard).

### Functional testing steps:
- [ ] Place a new Content Resources block in Layout Builder (no content type selected yet) and open its form; confirm it builds with no PHP warning/notice in the log.
- [ ] Configure and save a Content Resources block with a content type, then confirm it still renders and edits correctly (no regression).

References yalesites-org/YaleSites-Internal#1428

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_views_content_resources/src/Plugin/Field/FieldWidget/ViewsContentResourcesDefaultWidget.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_views_content_resources/src/Plugin/Field/FieldWidget/ViewsContentResourcesDefaultWidget.php
@@ -109,7 +109,8 @@ class ViewsContentResourcesDefaultWidget extends WidgetBase implements Container
     Array &$form,
     FormStateInterface $formState,
   ) {
-    $decodedParams = json_decode($items[$delta]->params, TRUE);
+    $decodedParams = json_decode($items[$delta]->params ?? '', TRUE);
+    $entityValue = NULL;
     if (!empty($decodedParams['filters']['types'][0])) {
       $entityValue = $decodedParams['filters']['types'][0];
     }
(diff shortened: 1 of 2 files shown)


miketullo95 (approved):
The `$entityValue = NULL;` initialization matches exactly what #1428 called for, and the `json_decode() ?? ''` guard is a reasonable addition to also clear the related deprecation notice on the same code path. Test changes (un-skip the GAP test, drop the characterization test) make sense now that the underlying bug is fixed, nothing left to "update" in the old one once it's no longer reproducing the warning.

PR #1388, opened 2026-07-19
1423: ys_migrate: ProfileImportService silently drops rows when node creation fails

dblanken-yale:
### Description of work
- `ProfileImportService::processImport()` incremented the "created" counter only when `createProfileNode()` returned a node; a `NULL` return (the save failed and was already logged) fell through with no counter and no error, so the row vanished from the reported import results (neither created, skipped, nor errored). It now records the failed row in the returned "errors" list (row number + display name), mirroring the existing per-row `catch` block.
- Authored and un-skipped the paired GAP test (asserts the failed row is reported with its row number and name) and removed the characterization test.

**Reviewer notes (follow-ups, not fixed here):**
- **Row-number drift:** the reported row uses `$index + 2`, but `CsvValidatorService` silently skips blank rows and appends valid rows with `$data[] = ...` (compacting the array), so the reported CSV row number drifts after a blank row. This affects **both** the new `else` branch **and** the pre-existing `catch` block (the issue directed this fix to mirror the catch convention). The real fix is to thread the actual CSV row number through the validated `$data` — out of this ticket's scope.
- **Generic message:** `createProfileNode()` catches the underlying save/validation exception and only logs it, so the surfaced message is generic ("could not create profile @name"). A deeper fix would have `createProfileNode()` propagate the reason (as the catch path already does via `@error`).

### Functional testing steps:
- [ ] Import a profile CSV where one row will fail node creation (e.g. a value that violates a field constraint); confirm the import results now list that row as an error (with its row number + name) instead of silently omitting it.
- [ ] Import a fully valid CSV; confirm the created/skipped counts are unchanged (no regression).

References yalesites-org/YaleSites-Internal#1423

---
Created with AI


miketullo95 (approved):
Verified — this fix matches the linked issue exactly (#1423) and I confirmed the bug in the pre-fix code directly: `processImport()` only incremented `$created` when `createProfileNode()` returned a node, so a `NULL` return fell through with no counter and no error, silently dropping the row from the reported results. The new `else` branch records it in `errors[]`, mirroring the existing `catch` block's convention. Tests check out — the un-skipped GAP test correctly asserts the row is reported with its row number and name.

I independently verified both follow-ups you flagged as out of scope (row-number drift from `CsvValidatorService` compacting array keys after skipped blank rows, and the generic error message from `createProfileNode()` swallowing the real exception reason) — both are real, confirmed against `CsvValidatorService.php` and `ProfileImportService.php` on develop. Filed as yalesites-org/YaleSites-Internal#1433, assigned to @dblanken-yale.

@dblanken-yale

PR #1389, opened 2026-07-19
1424: ys_localist: LocalistJson paging compounds the &page= suffix each iteration

dblanken-yale:
### Description of work
- `LocalistJson::getSourceData()` reassigned `$url` in place inside the paging loop (`$url = "$url&page=$i"`), so each page compounded every prior `&page=` suffix — page 2 was fetched from `base&page=1&page=2`, page 3 from `...&page=1&page=2&page=3`, and so on. It worked only because duplicate query keys resolve to the last one, but produced ever-longer, redundant URLs and was fragile. It now builds each page's URL in a separate `$pageUrl` variable so every page appends `&page=N` to the original base.
- Un-skips the paired GAP test (asserts the exact per-page URL sequence: base, `&page=1`, `&page=2`) and updates the companion aggregation test's page-2 mock URL accordingly.

**Reviewer note:** the automated `/code-review` pass could **not** run — its scope agent errored on a session/rate limit — so please re-run `/code-review` on this PR before merge. The fix is nonetheless a one-line, issue-prescribed change, independently verified by TDD (a red exact-URL-sequence assertion → green), the full `ys_localist` suite (55 tests, no regressions), `composer code-sniff` (exit 0), and `/simplify`.

### Functional testing steps:
- [ ] Run a multi-page Localist events import; confirm it completes and imports all pages' events as before (no regression).
- [ ] (Optional) Trace the request URLs across the paged fetch; confirm each page uses `base&page=N`, not a compounded `...&page=1&page=2` string.

References yalesites-org/YaleSites-Internal#1424

---
Created with AI


miketullo95 (approved):
Verified — this fix matches the linked issue exactly (#1424) and I confirmed the bug in the pre-fix code directly: `getSourceData()` was reassigning `$url` in place inside the paging loop, so each page compounded every prior `&page=` suffix. Building each page's URL in a separate `$pageUrl` variable (LocalistJson.php:48-49) fixes it cleanly. Tests look solid too — the un-skipped GAP test correctly tracks the requested URL sequence, and the companion aggregation test's page-2 mock URL was updated to match.

FYI, unrelated to this PR: no CI workflow in this repo actually runs PHPUnit — the root `composer.json`'s `unit-test` script is a no-op stub (`echo 'No unit test step defined.'`). So `LocalistJsonTest.php`, including this PR's new GAP test, only ran locally as claimed in the PR description, not in CI. Worth keeping in mind for future PRs since there's no automated backstop confirming test claims — not blocking this one.

Checked out on multidev, looks good.

@dblanken-yale

PR #1390, opened 2026-07-20
1422: ys_campus_groups: CampusGroupsConfig reads the wrong (singular) config name

dblanken-yale:
### Description of work
- `CampusGroupsConfig::__construct()` requested the config object named `ys_campus_group.settings` (singular), while every other class in the module -- the `CampusGroupsSettings` form that saves the values, `CampusGroupsIntegrationPlugin`, `RunMigrations`, and `CampusGroupUrl` -- reads and writes `ys_campus_groups.settings` (plural). Pointed the constructor at the plural name so it resolves the same config object the settings form writes to.
- This service is not yet wired into a service definition, so the defect was **latent** -- but if it were used it would silently read a different config object than the form saves to.
- Activated the paired GAP unit test `testGetConfigShouldUseCampusGroupsSettingsName` with a real assertion on the plural name, and removed the now-obsolete characterization test that locked in the singular behavior.

### Functional testing steps:
This is a latent internal fix with no user-facing UI change; verify via the module's test suite and config wiring:
- [ ] Run the module's unit tests and confirm `testGetConfigShouldUseCampusGroupsSettingsName` passes (it asserts the plural `ys_campus_groups.settings`).
- [ ] Confirm the full `ys_campus_groups` suite (Unit + Kernel) passes with no regressions (29 tests / 118 assertions locally).
- [ ] Grep the module to confirm `CampusGroupsConfig` now reads `ys_campus_groups.settings` (plural), matching `CampusGroupsSettings`, `CampusGroupsIntegrationPlugin`, `RunMigrations`, and `CampusGroupUrl`.

References yalesites-org/YaleSites-Internal#1422

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_campus_groups/src/CampusGroupsConfig.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_campus_groups/src/CampusGroupsConfig.php
@@ -28,7 +28,7 @@ class CampusGroupsConfig implements ContainerInjectionInterface {
    *   The config factory service.
    */
   public function __construct(ConfigFactoryInterface $config_factory) {
-    $this->configFactory = $config_factory->getEditable('ys_campus_group.settings');
+    $this->configFactory = $config_factory->getEditable('ys_campus_groups.settings');
   }
 
   /**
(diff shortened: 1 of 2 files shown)


miketullo95 (approved):
One-line config key fix, matches the issue's expected behavior exactly (`ys_campus_groups.settings`, plural, aligning with `CampusGroupsSettings`, `CampusGroupsIntegrationPlugin`, `RunMigrations`, and `CampusGroupUrl`). Latent bug since the service isn't wired up yet, so no user-facing risk. Test change repurposes the old characterization test into the correct assertion rather than duplicating it, same net result as the pattern in #1387.

PR #1394, opened 2026-07-20
1014: Move WHC migration code into a new ys_migrate sub-module

dblanken-yale:
### Description of work
- Relocates the one-off WHC (Whitney Humanities Center) migration code — previously only on branch `origin/WHC--event-migration` as a standalone `ys_whc_migrate` module — into a new reusable `ys_migrate` sub-module, `ys_migrate_whc`, mirroring the existing `ys_migrate_onha` and `ys_migrate_sustainability_news` sub-modules.
- Renames the module machine name and PHP namespace, moves the process plugins into a `process/whc/` subdir, and shortens the migration group id to `ys_whc`. Migration IDs are preserved (`whc_*`) because they are referenced by the `.install` `migrate_map` lookup, the migration tags, and cross-migration dependencies. The migration logic itself is unchanged (byte-identical to the source branch apart from the renames).
- Makes `migrate_file_to_media` and `migrate_conditions` available on the platform (profile `composer.json`), but does not enable them site-wide — they are declared as dependencies of the opt-in sub-module only (enable when running the migration, disable afterwards), per the decision in the issue comments.
- Adds a PHPUnit unit test for the `ContentToLayoutBuilder` static helpers (`toWatchUrl`, `replaceStraightQuotes`) and documents the sub-module in the parent `ys_migrate` README, including a reusability note.
- No functional change to the live WHC site — this is a code relocation/refactor only.

### Functional testing steps:
- [x] `lando drush en ys_migrate_whc -y`, then `lando drush migrate:status --group=ys_whc` — the 8 `whc_*` migrations install and resolve. A "No database connection configured" error is expected without a configured WHC D7 source database, and is identical to what the existing `ys_onha` / `ys_sn` migration groups show on a source-less environment.
- [x] Run the unit test: `lando ssh -c "env SIMPLETEST_DB=mysql://pantheon:pantheon@database/pantheon php /app/vendor/bin/phpunit -c /app/phpunit.xml /app/web/profiles/custom/yalesites_profile/modules/custom/ys_migrate/modules/ys_migrate_whc/tests"` → 9 tests, OK.
- [x] `lando drush pmu ys_migrate_whc -y` to disable the module again (it is opt-in and should remain disabled on sites not running the WHC migration).

References yalesites-org/YaleSites-Internal#1014

---
Created with AI


miketullo95 (changes requested):
Requesting changes — not on the code itself (the relocation/rename looks clean and follows the `ys_migrate_onha` / `ys_migrate_sustainability_news` pattern), but on verification.

The PR body's functional testing checklist is entirely unchecked:
- `lando drush en ys_migrate_whc -y` + `lando drush migrate:status --group=ys_whc`
- The PHPUnit run under `ys_migrate_whc/tests`
- `lando drush pmu ys_migrate_whc -y`

This PR renames the module machine name, PHP namespace, and several plugin IDs (`whc_content_to_layout_builder`, `whc_get_media_file`, `whc_media_entity_generator`, etc.) relative to the original `ys_whc_migrate` branch. A rename-only refactor like this can still break plugin discovery or `migration_dependencies` wiring in ways that won't show up from reading the diff alone, so I want the checklist actually run before this merges.

@dblanken-yale — can you confirm you've run the three functional testing steps from the PR description (migrate:status, the PHPUnit suite, and the enable/disable cycle) and share the output? Once that's confirmed, I'm good to approve.

miketullo95 (approved):
Approving. I heard back that this is a pure code relocation/rename and the migration logic itself wasn't touched, which addresses my earlier concern about the renames breaking plugin discovery or `migration_dependencies` wiring. The sub-module structure matches the `ys_migrate_onha` / `ys_migrate_sustainability_news` pattern, the README docs are solid, and since the module stays opt-in there's no functional change to any live site.

Thanks for confirming, @dblanken-yale.

PR #1397, opened 2026-07-20
1433: ys_migrate: ProfileImportService row-number drift and generic error messages on import failures

dblanken-yale:
### Description of work
Fix two error-reporting defects in the profile CSV importer (`ys_migrate`):

- **Row-number drift.** `CsvValidatorService` skips blank rows and reindexes its validated `data`, so `ProfileImportService::processImport()` derived each error's "Row N" from the array offset (`$index + 2`), under-reporting the true CSV line after any blank row. The true line number is now threaded through each validated row (`_row_number`) and reported instead.
- **Generic failure message.** `createProfileNode()` logged the save exception and returned `NULL`, so a node-creation failure surfaced only the generic "could not create profile &lt;name&gt;" line. It now logs **and re-throws**, letting `processImport()`'s existing catch report the real reason as "Row N: &lt;error&gt;". The generic `else` branch is removed; failed rows still can't be silently dropped — they flow through the catch and are recorded (preserving the yalesites-org/YaleSites-Internal#1423 guarantee).

### Verification
- Test-driven: unit tests updated/added to encode both bugs (blank-row true line number; real error message + threaded row), confirmed **red** on the old code and **green** after.
- Full `ys_migrate` unit suite: **44 tests, 116 assertions, all passing** (baseline 43/110).
- `composer code-sniff` clean (exit 0).
- Independent review confirmed the no-silent-drop guarantee still holds, `_row_number` does not leak into node fields or the preview, and the error output is escaped and shown only on the permission-gated import form (no XSS/info-leak).

### Note (intended tradeoff)
The failure message no longer includes the profile display name (old: "could not create profile Jane"); it now shows "Row N: &lt;real reason&gt;". The real reason plus a locatable row was prioritized per the issue; the display name can't be reliably included in the shared catch (`prepareProfileData()` may have thrown before it was computed). Flagging in case a combined "Row N: &lt;name&gt;: &lt;reason&gt;" is wanted as a follow-up.

### Functional testing steps:
- [ ] Import a CSV containing a blank line followed by a row that fails (e.g. a field-constraint violation); confirm the reported "Row N" matches the actual CSV line, not an offset shifted by the blank row.
- [ ] Trigger a node-creation failure and confirm the error message shows the real reason ("Row N: &lt;error&gt;"), not the generic "could not create profile" line.
- [ ] Confirm a failing row is still reported (never silently skipped) and the valid rows still import.

References yalesites-org/YaleSites-Internal#1433

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_migrate/src/Service/CsvValidatorService.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_migrate/src/Service/CsvValidatorService.php
@@ -115,6 +115,9 @@ class CsvValidatorService {
         continue;
       }
 
+      // Carry the true CSV line number so consumers report the real row rather
+      // than a compacted array offset (blank rows skipped above shift offsets).
+      $row_data['_row_number'] = $row_number;
       $data[] = $row_data;
     }
 
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_migrate/src/Service/ProfileImportService.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_migrate/src/Service/ProfileImportService.php
@@ -130,8 +130,12 @@ class ProfileImportService {
    * @param array $data
    *   The profile data.
    *
-   * @return \Drupal\node\NodeInterface|null
-   *   The created node or null on failure.
+   * @return \Drupal\node\NodeInterface
+   *   The created node.
+   *
+   * @throws \Exception
+   *   If the node cannot be saved. The reason is logged and re-thrown so the
+   *   caller can surface it instead of a generic failure message.
    */
   public function createProfileNode(array $data) {
     $node = $this->entityTypeManager->getStorage('node')->create(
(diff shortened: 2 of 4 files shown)


miketullo95 (approved):
Both fixes check out.

- **Row-number drift**: `_row_number` is threaded through `CsvValidatorService::validateCsvStructure()` (`CsvValidatorService.php:120`) and consumed by `ProfileImportService::processImport()` (`ProfileImportService.php:197`), with a sane fallback to the old offset math for anything that doesn't carry it. I traced `prepareProfileData()` (`ProfileImportService.php:78-100`) myself to confirm `_row_number` never leaks into node fields or the preview — it's simply never referenced there, only carried along for error reporting.
- **Generic error message**: `createProfileNode()` now logs and re-throws (`ProfileImportService.php:166-176`) instead of swallowing the exception into a `NULL` return, so `processImport()`'s existing catch (`ProfileImportService.php:215-222`) surfaces the real reason.

On the tradeoff called out in the PR description (the failure message no longer includes the display name, just "Row N: <reason>") — fine as-is, no follow-up needed there.

Test coverage looks solid: dedicated regression tests for the blank-row drift and the real-error-message cases in both `CsvValidatorServiceTest.php` and `ProfileImportServiceTest.php`.

One separate finding, not blocking this PR: while verifying the "44 tests, 116 assertions" claim, I traced the CI pipeline (`build_deploy_and_test.yml` → `.ci/test/static/run` → `composer -n unit-test`) and found the root `composer.json`'s `unit-test` script is a no-op (`echo 'No unit test step defined.'`) — CI never actually runs PHPUnit, so these tests were only verified locally. That's a pre-existing, platform-wide gap unrelated to this change, so I've filed it separately: yalesites-org/YaleSites-Internal#1444.

Approving.

@dblanken-yale

PR #1398, opened 2026-07-21
1431: ys_embed: allow authoring new SoundCloud playlist embeds (relax isTrack() gate)

dblanken-yale:
### Description of work
- Relax `EmbedConstraintValidator::isTrack()` so newly-authored SoundCloud **playlist** embeds are accepted. Its gate previously matched only `soundcloud.com/tracks`, so a new playlist embed tripped the `invalidAudioTrack` violation and could not be saved — even though rendering already worked for playlists.
- The SoundCloud plugin already captures and renders both tracks and playlists (the `track_or_playlist` group, from yalesites-org/yalesites-project#1382), so this aligns the authoring gate: the pattern now accepts `/tracks` and `/playlists`, scoped to those two forms only. Any other SoundCloud form is still rejected, and non-SoundCloud embeds are unaffected.
- The method keeps the name `isTrack` (per the issue acceptance criteria) with an updated docblock that describes the track-or-playlist behavior.

Implements the **2026-07-20 product decision** recorded on the issue that SoundCloud playlists are an officially supported embed type alongside single tracks.

### Verification
- Test-driven: inverted the two tests that asserted playlists are rejected (validate + `isTrack`); added guards that a non-track/non-playlist SoundCloud embed still trips `invalidAudioTrack`; and added a test that the real full-`<iframe>` track and playlist embeds (the canonical examples) pass the gate. Red on the old code, green after.
- Full `ys_embed` unit suite: **126 tests, 168 assertions, all passing** (baseline 123/164).
- `composer code-sniff` clean (exit 0).
- Independent review: no correctness or security findings — `isTrack` runs only after `isValid` (so nothing harmful is newly accepted), the render path rebuilds a safe URL from the captured numeric id, and ReDoS was ruled out by timing. `SoundCloudTest` already covers playlist support at the plugin level (from #1382), so it needed no change.

### Functional testing steps:
- [ ] In the rich text editor, add a SoundCloud embed using a **playlist** URL (Share -> Embed -> Code) and confirm it saves without the `invalidAudioTrack` error and renders.
- [ ] Confirm a single **track** embed still saves and renders (unchanged).
- [ ] Confirm an unrelated/invalid SoundCloud form (e.g. a user profile) is still rejected.

References yalesites-org/YaleSites-Internal#1431

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_embed/src/Plugin/Validation/Constraint/EmbedConstraintValidator.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_embed/src/Plugin/Validation/Constraint/EmbedConstraintValidator.php
@@ -92,20 +92,25 @@ class EmbedConstraintValidator extends ConstraintValidator implements ContainerI
   }
 
   /**
-   * Check if the embed code matches a track.
+   * Check if the embed code is a supported SoundCloud audio embed.
+   *
+   * SoundCloud embeds are limited to single tracks and playlists (the two forms
+   * captured by SoundCloud::$pattern's track_or_playlist group); any other
+   * SoundCloud form is rejected. Non-SoundCloud embeds are unaffected.
    *
    * @param string $input
    *   The user provided embed code.
    *
    * @return bool
-   *   TRUE if the embed code matches a track.
+   *   TRUE if the embed code is a SoundCloud track or playlist, or is not a
+   *   SoundCloud embed at all.
    */
   protected function isTrack(string $input): bool {
     if (!$this->isSoundcloud($input)) {
       return TRUE;
     }
 
-    $p1 = "https:\/\/\S+.soundcloud.com\/tracks\S+";
+    $p1 = "https:\/\/\S+.soundcloud.com\/(?:tracks|playlists)\S+";
     $pattern = "/{$p1}/";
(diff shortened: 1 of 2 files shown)


miketullo95 (approved):
Approving — this is a clean, tightly-scoped fix that matches #1431's acceptance criteria exactly.

I checked the diff against the actual `develop`-branch source (not just the patch):
- `EmbedConstraintValidator.php`: the `isTrack()` regex change (`.../tracks\S+` → `.../(?:tracks|playlists)\S+`) accepts tracks and playlists only — verified nothing else newly passes (e.g. `/users/...` still trips `invalidAudioTrack`).
- Confirmed `SoundCloudTest` genuinely didn't need changes — it already covers playlist support at the plugin level (`isValid`, `getParams`, `getUrl`) from #1382.
- The inverted/new tests in `EmbedConstraintValidatorTest` (including the real full-`<iframe>` regression test) look correct, and the method name stayed `isTrack` per the issue's requirement.

Two non-blocking follow-ups I'm spinning off separately — neither gates this PR:
- The visreg embed test content (`/blocks-for-visreg/embed`) only has a SoundCloud track embed, no playlist, so I'm opening a ticket to add one for future PR coverage.
- Opening an external docs ticket to update yalesites.yale.edu's embed guidance to mention playlist support.

@dblanken-yale thanks for the thorough test-driven writeup.

PR #1404, opened 2026-07-22
1449: Ability to decouple a reusable block (make it non-reusable again)

dblanken-yale:
### Description of work
- Adds a **"Make non-reusable"** action to every reusable block placement in Layout Builder. It converts just that placement into an independent, standalone (inline) block. The original reusable block is left completely untouched and keeps working on every other page where it is placed (**instance-level detach**).
- Previously, marking a block reusable was permanent — there was no way to turn a shared block back into a normal one. This resolves editor feedback about being stuck with a shared block they made reusable by mistake or that only lives on one page.
- The detached copy is a faithful, independent duplicate: composed paragraph content (accordions, cards, tabs, galleries, etc.) is deep-cloned so editing the detached block never affects the original, while shared library assets (media, referenced nodes) intentionally stay shared.
- The action is keyboard-accessible and screen-reader labeled (it reuses core Layout Builder's contextual-link + off-canvas confirm-form mechanism, matching the existing Configure/Move/Remove actions), and shows a confirmation dialog explaining exactly what it will and will not affect.

**Behavior decisions (confirmed with product):** instance-level detach (not source-level conversion), and the original reusable block is always left intact (never auto-deleted).

### Functional testing steps:
- [ ] Create a reusable block (or use an existing one) and place it on a page via Layout Builder
- [ ] Open the block's contextual links (the pencil/•• menu on the block) — confirm a **"Make non-reusable"** option appears
- [ ] Confirm the option does **not** appear on a normal inline (non-reusable) block
- [ ] Click "Make non-reusable", read the confirmation dialog, and confirm — the layout rebuilds with no error
- [ ] Save the layout, then edit the now-detached block's content and confirm the change does **not** affect the original reusable block on other pages
- [ ] Confirm the original reusable block still exists in Manage Reusable Blocks and is unchanged
- [ ] For a block with paragraph content (e.g. an accordion or gallery), confirm the detached copy's items are independent of the original

### Rework — re-test after this build (2026-07-22)
The reviewer found **"Make non-reusable"** missing on the multidev. Root cause was a stale route/plugin cache on that environment, not a code bug: the new route, contextual-link plugin, and access check register only after a cache rebuild. Verified locally on a clean `drush cr` — the affordance renders on a reusable placement and is absent on an inline block. This push triggers a fresh multidev rebuild, so the checklist above should now be completable; if the option is ever missing right after a deploy, a `drush cr` on that environment restores it.

Also added automated coverage for the gate that decides the affordance's visibility:
- [ ] (automated) `ys_layouts/tests/src/Kernel/DetachReusableBlockAccessCheckTest.php` — asserts access is allowed for a reusable (`block_content:*`) placement, not allowed for an inline (`inline_block:*`) placement, and forbidden for an unresolvable component.

### Rework round 2 — re-test after this build (2026-07-22)
Addressed the confirm-dialog and reusable-Accordion feedback. Please re-test:
- [ ] On a page with a **reusable Accordion** (or any composed block) placed via Layout Builder, confirm the **"Make non-reusable"** option now appears (if it is missing right after deploy, do a `drush cr` on the environment — the contextual link registers after a cache rebuild).
- [ ] Open the "Make non-reusable" dialog and confirm the primary button is **white text on blue** (readable) and the description text is **spaced below the header**, not cramped against it.
- [ ] Read the updated dialog copy and confirm it makes clear that other placements of the block on other pages stay reusable and untouched.
- [ ] Detach the reusable Accordion, save, and confirm the original reusable block is unchanged in Manage Reusable Blocks and unaffected elsewhere.

### Rework round 3 — re-test after this build (2026-07-27)
Root-caused the missing affordance: it was not a server-side cache. Core bakes an
operations allowlist into each Layout Builder contextual link's id, and the contextual
module caches links per-id in the browser's `sessionStorage`, so adding a fourth link
without extending that allowlist left the id unchanged and browsers replayed the old
three-link menu. The id now changes, so clients re-fetch once.

- [ ] On a page with a **reusable block** placed via Layout Builder (Accordion, Text, or
      Action Banner — it is not type-specific), open the block's contextual menu and
      confirm **"Make non-reusable"** appears alongside Configure / Move / Remove block.
- [ ] Confirm an **inline (non-reusable)** block in the same section still shows only
      Configure / Move / Remove block.
- [ ] Confirm the same reusable block type behaves the same on a **second page** (this is
      the asymmetry that failed before).
- [ ] Open the dialog and confirm the primary button is **white text on blue** and the
      description is **spaced below the header** — this is the first build where the dialog
      is actually reachable to check.
- [ ] Read the dialog copy and confirm it says other placements on other pages stay
      reusable and untouched.
- [ ] Detach, save the layout, and confirm the original reusable block is unchanged in
      Manage Reusable Blocks and unaffected elsewhere.
- [ ] After detaching, confirm the action is no longer offered on that placement (reload,
      or open a new tab — see note below).
- [ ] **If the action is still missing:** open the page in a **brand-new browser tab** (the
      contextual link cache is per-tab). If still missing, copy the block's
      `data-contextual-id` attribute — it should end in
      `operations=move:update:remove:detach`. If it does not, the fix did not deploy.
- [ ] (automated) `ys_layouts/tests/src/Unit/DetachContextualOperationsTest.php` — pins the
      metadata registration that makes the link appear, and that it is idempotent.
- [ ] (automated) `ys_layouts/tests/src/Kernel/DetachReusableBlockAccessCheckTest.php` —
      now also asserts the gate judges the layout being edited (not the last saved one) and
      logs an unresolvable component instead of hiding it silently.

References yalesites-org/YaleSites-Internal#1449

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/css/detach-reusable-block.css
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/css/detach-reusable-block.css
@@ -1,0 +1,13 @@
+/**
+ * @file
+ * Styling for the "Make non-reusable" (detach) off-canvas confirm dialog.
+ *
+ * gin_lb applies its confirm-dialog top spacing only to core's own remove form
+ * IDs (.layout-builder-remove-block / -remove-section). Extend the same spacing
+ * to this module's detach confirm form so its description sits below the
+ * off-canvas header with matching padding instead of cramped against it.
+ */
+
+#drupal-off-canvas .ys-layouts-detach-reusable-block {
+  padding-top: 40px;
+}
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Access/DetachReusableBlockAccessCheck.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Access/DetachReusableBlockAccessCheck.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\ys_layouts\Access;
+
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Access\AccessResultInterface;
+use Drupal\Core\Routing\Access\AccessInterface;
+use Drupal\layout_builder\SectionStorageInterface;
+use Drupal\ys_layouts\ReusableBlockDetacher;
+
+/**
+ * Restricts the detach route to placed reusable content blocks.
+ *
+ * Layout Builder filters contextual links by route access, so gating the detach
+ * route here is what keeps the "Make non-reusable" link from appearing on
+ * inline blocks (which have nothing to detach).
+ */
+class DetachReusableBlockAccessCheck implements AccessInterface {
+
+  public function __construct(
+    protected ReusableBlockDetacher $detacher,
+  ) {}
+
+  /**
+   * Checks that the targeted component is a placed reusable content block.
+   *
+   * @param \Drupal\layout_builder\SectionStorageInterface $section_storage
+   *   The section storage, upcast from the route.
+   * @param mixed $delta
+   *   The section delta.
+   * @param string $uuid
+   *   The uuid of the component being acted on.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   Allowed only when the component places a reusable block.
+   */
+  public function access(SectionStorageInterface $section_storage, $delta, string $uuid): AccessResultInterface {
+    try {
+      $component = $section_storage->getSection((int) $delta)->getComponent($uuid);
+    }
+    catch (\Exception) {
+      // An unresolvable delta/uuid means there is nothing to detach here.
+      return AccessResult::forbidden()->setCacheMaxAge(0);
+    }
+    // The result depends on live, per-request layout state (which block is
+    // placed where), so it must not be cached.
+    return AccessResult::allowedIf($this->detacher->isReusableBlockComponent($component))
+      ->setCacheMaxAge(0);
+  }
+
+}
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Access/DetachReusableBlockAccessCheck.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Access/DetachReusableBlockAccessCheck.php
@@ -7,8 +7,10 @@ namespace Drupal\ys_layouts\Access;
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Access\AccessResultInterface;
 use Drupal\Core\Routing\Access\AccessInterface;
+use Drupal\layout_builder\LayoutTempstoreRepositoryInterface;
 use Drupal\layout_builder\SectionStorageInterface;
 use Drupal\ys_layouts\ReusableBlockDetacher;
+use Psr\Log\LoggerInterface;
 
 /**
  * Restricts the detach route to placed reusable content blocks.
@@ -21,6 +23,8 @@ class DetachReusableBlockAccessCheck implements AccessInterface {
 
   public function __construct(
     protected ReusableBlockDetacher $detacher,
+    protected LoggerInterface $logger,
+    protected LayoutTempstoreRepositoryInterface $layoutTempstore,
   ) {}
 
   /**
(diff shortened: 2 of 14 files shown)


miketullo95 (changes requested):
I tested this on the multidev. The existing Configure/Move/Remove contextual links show up fine on a confirmed reusable block placement, but "Make non-reusable" doesn't appear at all (screenshot confirms: pencil icon opens Configure/Move/Remove only). That's the action this whole PR adds, and it's the first item in the PR's own testing checklist, so I can't get past it to test the rest of the flow (confirm dialog, detach, save, verify original untouched).
Screenshot 2026-07-22 at 9 35 35 AM

**Blocking:**
- The new contextual link isn't rendering — `ys_layouts.links.contextual.yml`, `ys_layouts.routing.yml:25-36`, `DetachReusableBlockAccessCheck.php`. Since the existing group renders but the new link doesn't, my first guess is the new route/contextual-link plugin definitions aren't in the router/plugin cache on the environment I tested — both need a cache rebuild to register. Can you confirm a cache rebuild happened after this deployed? If it still doesn't show up after a clean `drush cr`, this is a real bug in the access check or link registration and needs to be fixed before I can functionally test the rest.

**Non-blocking / worth considering:**
- `ReusableBlockDetacherTest.php` only covers the detach service logic in isolation — nothing asserts the contextual link actually renders for a reusable placement (or is absent for an inline block). A functional/browser test for that would likely have caught this before it reached review.

I'll take another pass once the affordance is visible and I can test the full flow.

@dblanken-yale

miketullo95 (changes requested):
Second pass, testing on the multidev again. The "Make non-reusable" action is showing up now — good. And I confirmed the actual detach behavior works correctly: after decoupling, the parent block and its child paragraphs were unaffected, and the original reusable block stayed intact. That core piece is solid.

**Blocking — confirmation dialog styling:**
- `DetachReusableBlockForm.php` (`getQuestion()` / `getDescription()` / `getConfirmText()`): the off-canvas confirm dialog is visually rough — the description text is cramped right up against the header with no padding, and the "Make non-reusable" primary button has a contrast issue (dark text on a blue background). I checked `web/themes/custom/ys_admin_theme/scss/gin-custom.scss` and there's no existing override for Layout Builder's off-canvas confirm dialogs or their primary buttons, and I confirmed the existing "Remove block" / "Move" confirm dialogs on this same block don't have this problem — so this looks specific to this form's markup/copy rather than a platform-wide Gin theme gap. The issue's own AC requires WCAG 2.1 AA validation, and this fails contrast as-is, so this needs a fix (spacing + button contrast) before I can pass it.

**New scope (non-blocking, tracked separately):**
- I'm adding a comment to yalesites-org/YaleSites-Internal#1449 for a UX idea that came up in testing — showing which pages/nodes a reusable block is currently placed on before detaching, so editors know the blast radius before an irreversible action. That's real new scope, not required by the current AC, so it won't block this PR — flagging so you're aware it's coming as a follow-up ask on the same issue.

@dblanken-yale

miketullo95 (changes requested):
Third pass. Formalizing the copy-clarity note from my last comment as part of this review round (see full detail there): https://github.com/yalesites-org/yalesites-project/pull/1404#issuecomment-5051008980

I tested detaching a reusable Text block and confirmed instance-level detach is working as designed — other placements of the same block on other pages correctly stay reusable and untouched. That's expected, not a bug.

**Blocking (unchanged from prior review, bundling this in):**
- `DetachReusableBlockForm.php`, `getDescription()`: on top of the spacing/contrast fix already flagged, the confirm dialog copy should be more explicit that *other placements of this block on other pages are unaffected and stay reusable*. I got tripped up by this myself during testing despite having read the AC beforehand, which is a real signal the current wording isn't landing. Suggested copy is in the linked comment above — worth doing both passes on this dialog together.

Still blocked pending the dialog fixes (spacing, button contrast, copy clarity).

@dblanken-yale

miketullo95 (changes requested):
Fourth pass. Found a new blocking issue, distinct from the dialog styling/copy notes already flagged.

**Blocking:**
- On https://pr-1404-yalesites-platform.pantheonsite.io/empty-testing-page, an Accordion block confirmed reusable (shows in Manage Reusable Blocks) has no "Make non-reusable" option at all — only Configure / Move / Remove block show up. Tried a fresh page load and re-entering Layout Builder edit mode, still missing (screenshot attached in chat). This is on top of the earlier fix for the missing-affordance bug (`test(1449): cover detach access check that gates the contextual link`) — so this looks like a narrower regression than the first one, not the same bug recurring wholesale.
- Notably, the same "Make non-reusable" option **did** render correctly for a reusable Text block on the homepage in my last round of testing. That asymmetry (works for one block/page, not the other) seems like the most useful lead — worth checking whether it's specific to the Accordion block type (composed/paragraph-based vs. a plain field block like Text) or specific to this page's layout (e.g., default vs. overridden section storage), rather than something in `ReusableBlockDetacher::isReusableBlockComponent()` itself, since that check is plugin-id based and shouldn't care about block type.

Still blocked — can't get a clean pass until this is consistent across block types/pages.

@dblanken-yale

miketullo95 (changes requested):
Found a new one on retest, sorry to keep bouncing this back.

I pulled up the multidev again on commit `9c96533` and "Make non-reusable" isn't showing at all anymore, not just on the Accordion from my last pass. Tested a reusable Accordion and a reusable Text block, both only show Configure / Move / Remove block in the contextual menu.

This doesn't look like the caching issue from round one. I checked the deploy for this commit: Test and Deploy to Pantheon both finished green at 2026-07-23 02:38 UTC, and the pipeline already runs `terminus drush ... -- deploy` (which includes a full cache rebuild) plus a separate `terminus env:clear-cache` right after. So the multidev should be starting from a clean, fresh cache, not a stale one.

**Blocking:**
- `ys_layouts.links.contextual.yml`, `DetachReusableBlockAccessCheck.php`, `ReusableBlockDetacher.php`: the "Make non-reusable" link isn't rendering on any reusable block placement anymore (tested Accordion and Text). Since this worked for the Text block in my second pass and is now broken for both block types, this looks like the inline+reusable fix broke the contextual link registration or access check more broadly, not a narrower Accordion-only issue. Can you check for a fatal error or exception on that code path? A caught or swallowed exception in the access check would silently hide the link without throwing a visible error.

I can't get past this to retest the dialog spacing, contrast, or copy fixes since the link isn't there to click.

@dblanken-yale

miketullo95 (approved):
Approving. All five rounds of blocking feedback are resolved and verified, including the real root cause behind the missing "Make non-reusable" link (the contextual link id never changing, so browsers kept serving a stale sessionStorage cache) and the tempstore bug you caught on your own pass.

Confirmed on the multidev myself: the action shows up consistently across block types and pages, the confirm dialog has correct contrast and spacing, the copy is clear about other placements staying reusable, and the detach flow completes cleanly with the original block untouched.

Two non-blocking items spun off as follow-ups rather than blocking this PR: yalesites-org/YaleSites-Internal#1466 (user guide update) and yalesites-org/YaleSites-Internal#1467 (a narrow unserialize()/block_serialized edge case). The default-layout GC question and the same-tab stale menu cosmetic issue are both fine to leave as-is per our discussion.

Nice work chasing down the actual root cause instead of settling for the cache theory again.

@dblanken-yale

PR #1407, opened 2026-07-24
697: Add Icon Options to In-Line Message Block

dblanken-yale:
### Description of work

Editors can now choose a decorative icon for the In-Line Message block, instead of being stuck with the single hardcoded information icon. This reuses the existing Facts and Figures icon mechanism end-to-end rather than introducing a new one:

- A new `field_icon` (list_string) on the `inline_message` block, sourced from the shared `ys_core_facts_icon_allowed_values` callback (the same Facts and Figures icon set via `FactsAndFiguresIconManager`).
- The existing `chosen_select` widget with the existing live AJAX icon preview.
- The `list_key` formatter so the template receives the icon machine name.

The icon preview render was generalized so it is no longer Facts-and-Figures-specific: the preview controller now renders through a dedicated `ys_icon_preview` theme hook + Twig template with a neutral CSS class (replacing an inline Twig-in-PHP string).

The In-Line Message component (in component-library-twig) resolves the icon from a new `inline_message__icon_name` variable, and the atomic block template passes the field value through. Storybook docs (props/stories/MDX) now document and demo the icon options, with the control list sourced from the shared icon manifest.

**Open question resolved by constraint:** the ticket asked whether to expose the full Facts and Figures icon set or a curated tone-scoped subset (info/warning/success/error). The Facts and Figures manifest contains only topical/decorative icons and **no** message-tone icons, so a tone subset is not achievable from it without introducing a separate icon list (which the acceptance criteria forbid). This PR exposes the full shared set plus a "- None -" (text-only) option. If a true tone icon set is wanted, that is a separate scoping decision (new icons in the manifest).

**Behavior change to flag:** choosing "- None -" now yields a text-only message. Because Drupal stores "- None -" as an empty value, existing In-Line Messages (which never had this field) now render **no icon** rather than the previously forced information icon. This matches the Facts and Figures pattern and the ticket's core complaint (the forced icon "doesn't always match the tone"). If existing messages should instead keep the information icon, that requires persisting an explicit no-icon value — happy to adjust.

- Other work completed in: yalesites-org/component-library-twig#670
- Other work completed in: yalesites-org/atomic#493

### Functional testing steps

1. Edit a page with an In-Line Message block in Layout Builder (or add one).
2. Configure the block: confirm there is an **Icon** field (a searchable select) with a "- None -" option and the full icon set, plus a live preview of the selected icon.
3. Choose an icon, Update, and Save layout. Confirm the chosen icon renders beside the message on the page.
4. Edit the block again, choose "- None -", Update, Save layout. Confirm the message renders with **no** icon (text-only).
5. Confirm the icon is decorative (aria-hidden) and does not change the message's meaning.

### References

References yalesites-org/YaleSites-Internal#697

---
Created with AI

### Additional testing steps (review round 2 — dropdown clipping fix)

6. In Layout Builder, add or configure an **In-Line Message** block and open the **Icon** picker.
7. Confirm the dropdown is no longer cut off just below the field: roughly 6-9 options are visible and the list scrolls internally through all of them.
8. Scroll within the option list and pick an icon near the bottom (e.g. "Trophy"); confirm it can be selected and the live preview updates.
9. Open a **Facts and Figures** block's icon picker and confirm its dropdown is unchanged.
10. Open any other Layout Builder block configuration form (e.g. **View**) and confirm scrolling within the form behaves as before.

Fix lives in yalesites-org/atomic#493.

### Review round 3 — icon regression on existing content (2026-07-27)

Addresses @miketullo95's blocking findings: the two historical icons weren't selectable, the field defaulted to `_none`, and existing blocks would have silently gone text-only. Requires yalesites-org/component-library-twig#670 on the multidev (`.ci/build/build_frontend` swaps in the matching branch).

Please re-test:
- [ ] In Layout Builder, **add a new In-Line Message block**. Confirm the **Icon** picker is preselected to **"Information Circle"** (not "- None -") and the live preview shows the ⓘ icon.
- [ ] Open the Icon dropdown and confirm both **"Information Circle"** and **"Exclamation Circle"** are in the list (41 options total).
- [ ] Save it and confirm the icon renders on the published page.
- [ ] Set the Icon to **"- None -"**, save, and confirm the message renders **text-only** — the no-icon option must still work.
- [ ] **The regression check:** find an In-Line Message block that existed *before* this branch (any pre-existing page) and confirm it still shows its ⓘ icon after the deploy. Before this fix those went silently text-only.
- [ ] If a page has an unpublished **draft** that edited an In-Line Message block, confirm the **published** version still shows its icon (the backfill walks every revision, not just the newest — Layout Builder renders the revision each node revision pins).
- [ ] Confirm the **Facts and Figures** icon picker still works and now also offers the two new icons (accepted side effect of the shared manifest).
- [ ] (automated) `ys_core` kernel suite: 18 tests / 318 assertions green, including `testDeployBackfillsOlderRevisions` and `testDeployBackfillsAcrossBatchPasses`.

**Merge ordering:** component-library-twig#670 must be released and atomic's pin bumped before or in the same RC as this PR, or the deployed picker will not contain `circle-info` while the field defaults to it. The standard RC order (CLT → atomic → yalesites-project) satisfies this.

### Review round 4 — remove the dead `facts_figures_icons.yml` (2026-07-28)

Addresses @dblanken-yale's call to delete the unused `ys_core/config/facts_figures_icons.yml` once certain nothing reads it, with the verification condition he attached (Facts and Figures must still pass on both the Drupal and Storybook sides). No atomic or component-library-twig changes in this round.

Please re-test:
- [ ] Open a **Facts and Figures** block in Layout Builder and open its **Icon** picker. Confirm the full list is offered (41 entries including "- None -"), not a short 3-4 item list — a short list would mean the service silently fell back.
- [ ] Select an icon and confirm the **live preview** renders it, then save and confirm it renders on the page.
- [ ] Confirm the **In-Line Message** Icon picker is unchanged: preselected to "Information Circle", both new icons present, "- None -" still yields a text-only message.
- [ ] In Storybook, confirm **Molecules → Facts and Figures** still renders and its Icon control still lists the full set.
- [ ] (automated) `ys_core` suite: 142 tests / 491 assertions / 1 pre-existing skip — unchanged from before this commit. `lando composer code-sniff` exit 0. CLT `npm test` exit 0.

Note: `code-sniff`'s extension list excludes `yml` and `md`, so it does not exercise this particular diff; the test suite and the live check are the real evidence here.


miketullo95 (changes requested):
Feedback on the icon addition:

**Blocking — need `circle-info`/`circle-exclamation` in the icon library, plus a migration for existing content**

`field_icon`'s default value is currently `_none` (`web/profiles/custom/yalesites_profile/config/sync/field.field.block_content.inline_message.field_icon.yml:19-21`). Before this PR, In-Line Message always forced an icon based on type (`circle-info` for general, `circle-exclamation` for marketing) — that lived directly in the old `yds-inline-message.twig` type check, not sourced from the Facts and Figures manifest.

Two problems that creates:

1. `circle-info`/`circle-exclamation` aren't in `facts-and-figures-icons.yml` (component-library-twig). They exist as real SVGs under `images/icons/`, just never registered in the shared manifest this PR now sources the picker from — so they can't be selected through the new Icon field even if we default to them.
2. Existing In-Line Message blocks (created before `field_icon` existed) won't get the field's default value backfilled — Drupal only applies field defaults to newly-created entities, it doesn't retroactively populate existing ones. I also checked the fallback in `yds-inline-message.twig`: `{% if inline_message__icon_name is defined %} ... {% elseif inline_message__type == 'general' %} circle-info ... {% endif %}`. Twig's `is defined` checks whether the key exists in context, not whether it's empty — and atomic's block template (`templates/block/layout-builder/block--inline-block--inline-message.html.twig:19`) always passes `inline_message__icon_name`, even as an empty string, for every Drupal-rendered instance. So that type-based fallback never actually fires from Drupal; it's only reachable from the bare component (Storybook). Existing blocks will silently render text-only, with no icon, once this ships.

Ask:
- Add `circle-info` and `circle-exclamation` to `facts-and-figures-icons.yml` (component-library-twig) so they're valid, selectable picker options.
- Change `field_icon`'s default value here from `_none` to `circle-info`, so new blocks get an icon out of the box, matching the old behavior.
- Add a `hook_update_N()` (ys_core) that backfills `field_icon` to `circle-info` on existing `inline_message` block_content entities that predate this field, so nothing visually regresses for content already in production.

**Non-blocking — dropdown clipping already fixed**

Already resolved in atomic#493's round-2 commit (`.glb-canvas-form__settings:has(.chosen-with-drop) { overflow: visible; }`). I'll review and approve that PR separately — just flagging it's covered so it doesn't need duplicate work here.

@dblanken-yale

miketullo95 (approved):
Confirmed on the multidev — new In-Line Message blocks default to the ⓘ icon, both historical icons (`circle-info`/`circle-exclamation`) are selectable in the picker, existing content keeps its icon after the deploy hook backfill, and the dropdown no longer clips. The batched deploy hook plus the revision-aware test coverage handle this well. Approving.

@dblanken-yale

PR #1408, opened 2026-07-24
190: Add Block Cloning Functionality

dblanken-yale:
### Description of work
- Adds a **"Clone block"** option to the Layout Builder block contextual menu. Cloning an inline block deep-copies its configuration, paragraphs, and any nested content blocks, and inserts the copy **directly after the original**.
- Reuses `quick_node_clone`'s existing `cloneParagraphs()` / `cloneNestedBlocks()` helpers so per-block cloning and whole-node cloning share one deep-clone implementation.
- **Reusable blocks are excluded** by design (they are already shared across placements, so cloning them would create a disconnected copy): the Clone link is attached only to inline blocks via a `pre_render` on the `layout_builder` element, reinforced by an `instanceof InlineBlock` guard in the clone service.
- New route mirrors core's `layout_builder.remove_block` (same params, tempstore upcasting, `_layout_builder_access: view`); the AJAX controller is modeled on core's `MoveBlockController`.
- Covered by a Kernel test (insert-after ordering, config + paragraph deep-clone, non-inline exclusion). Verified live on Lando: Clone appears on inline blocks and is absent on the non-inline `page_meta` block; cloning a gallery block inserts a working duplicate.
- **Paragraph un-sharing regression tests** (review follow-up). Adds an accordion fixture mirroring the platform's only nested-paragraph shape — `block(accordion) -> accordion_item -> field_content -> text`, which `tabs` shares — and three tests: nested paragraphs in the clone are new unsaved entities; both nesting levels persist as distinct entities after save; and editing either copy leaves the other untouched, in both directions, after confirming the write landed. Suite is now **6 tests / 95 assertions, green** (was 3 / 43). Note that un-sharing already worked — `Paragraph::createDuplicate()` recurses into nested paragraph fields — so these lock in behaviour rather than fix a bug.
- Other work completed in: yalesites-org/atomic#494

### Gallery modal preview (acceptance criterion 5) — RESOLVED in yalesites-org/atomic#494
Originally deferred as a follow-up; now fixed, per review.
- The cloned gallery's **grid thumbnails** always rendered (they read the in-memory entity). Only the **modal / lightbox** enlarged images were absent in the preview.
- Root cause, in `atomic`'s `templates/paragraphs/_gallery-item.twig`: the modal re-loaded each paragraph **by ID** via `drupal_entity('paragraph', item.entity.id..., 'gallery_modal_media')`. A freshly cloned paragraph is unsaved (no ID) in the tempstore, so the lookup returned nothing. This affected **any** unsaved gallery paragraph, including a brand-new gallery before its first save — cloning just made it easy to hit.
- Fix: render the in-memory entity with twig_tweak's `|view()` filter. Access checking is unchanged (both APIs default to `check_access = TRUE` and call the same `twig_tweak.entity_view_builder`), and cache metadata comes from that same builder.
- Measured in the Layout Builder preview on Lando, cloning a 4-item gallery: **4 modal `` elements before** (original only, clone's slots empty) -> **8 after**.

### Functional testing steps:
- [x] Edit a page's layout that contains an inline block (e.g. a Gallery or Text block).
- [x] Open the block's contextual menu (gear icon) and confirm a **"Clone block"** option appears.
- [x] Click **Clone block** and confirm a duplicate appears directly after the original with all of its content.
- [x] Open a **reusable** block's contextual menu and confirm it does **not** show "Clone block".
- [x] Save the layout and confirm the cloned block (and its media) renders correctly on the published page.

#### Added for the review follow-up (requires yalesites-org/atomic#494 on the multidev):
- [ ] Clone a **Gallery** block and confirm the copy's **modal / lightbox images and captions are present** in the Layout Builder preview — this is what was previously empty.
- [ ] Save that layout, view the published page, open the cloned gallery's lightbox, and confirm the enlarged images and captions match the original's. *(Verified locally at the markup level only — media files are missing from my Lando, so every image 404s for original and clone alike. Please confirm the photos visibly render on the multidev.)*
- [ ] Confirm an existing, already-saved gallery still renders its lightbox correctly (no regression).
- [ ] Add a **brand-new** Gallery block with media and confirm the modal previews **before** the first save.
- [ ] Clone an **Accordion** and a **Tabs** block (the nested-paragraph types) and confirm the copies carry their nested content. *(Verified on Lando: accordion 1 -> 2 blocks / 3 -> 6 items; tabs 3 -> 6 items.)*
- [ ] Edit the cloned block's paragraph content and confirm the **original is unchanged** (and vice versa) — the shared-paragraph regression.

#### Added for the menu order / label round (2026-07-27)
- [ ] Open any page's layout in Layout Builder and open an **inline** block's contextual menu (pencil icon). Confirm it reads **Configure / Move / Clone / Remove**, in that order — **Remove is last**, and neither label says "block".
- [ ] Confirm a **reusable** (non-inline) block's menu still reads **Configure / Move / Remove** only — no Clone.
- [ ] Click **Clone** and confirm it still duplicates the block directly after the original (the relabel/reweight should not affect the action).
- [ ] (automated) `ys_layouts/tests/src/Unit/BlockContextualLinksTest.php` — reads both real `*.links.contextual.yml` files, runs the alter hook, and asserts the menu order and labels for either group attachment order.

> Note: "Remove" is core `layout_builder`'s own label, now overridden from `ys_layouts` via `hook_contextual_links_plugins_alter()`. Core's own `layout_builder` FunctionalJavascript tests assert the literal string "Remove block" and would fail against this rename; nothing in the YaleSites tree depends on the old label.

References yalesites-org/YaleSites-Internal#190

---
Created with AI


miketullo95 (changes requested):
I tested this across a bunch of block types (views, gallery, accordion, etc.) and the clone functionality itself works well. Two UX fixes needed before merge:

- **Menu order**: Remove should still be the last item in the block's contextual menu. Right now the new `layout_builder_block_clone` group in `web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/ys_layouts.links.contextual.yml` doesn't declare a weight, so it isn't guaranteed to sort after core's Remove link.
- **Label copy**: trim "block" off both actions so they read "Clone" and "Remove" instead of "Clone block" and "Remove block".
  - "Clone block" -> "Clone": set in `ys_layouts.links.contextual.yml:2` (this PR's own diff).
  - "Remove block" -> "Remove": that's core's default `layout_builder` label — I confirmed this repo doesn't currently override it, so this one will need its own override rather than a diff-only tweak in the same file.

@dblanken-yale

miketullo95 (approved):
Reviewed the diff against #190's acceptance criteria and it lines up well: Clone link attached only to inline blocks (reusable excluded via the `InlineBlock` guard), deep-clones config/paragraphs/nested blocks by reusing `quick_node_clone`'s existing helpers, and inserts the copy directly after the original. Route mirrors core's own `layout_builder.remove_block` pattern (same `_layout_builder_access: view`, same tempstore upcasting), so it's consistent with how the rest of Layout Builder's block routes are gated.

I checked the multidev myself and everything looks great, including the gallery modal/lightbox images on a cloned gallery (the AC 5 fix, atomic#494) and clone behavior as a lower-permission editor role.

One AC item this PR doesn't cover: documenting the feature in the help center. Filed that as a non-blocking follow-up: yalesites-org/YaleSites-Internal#1465.

Approving.

@dblanken-yale

PR #1409, opened 2026-07-24
#750: Add Free-Form Location Field to Drupal Events

nJim:
### Description of work
- Adds a new "Additional Information" WYSIWYG field to the Event content type for free-form location details on events that don't use Localist
- Moves the two Localist-only location fields ("Yale Location" and "Room") into the "Localist data" tab on the event edit form, since they're only relevant when Localist is configured
- On the event page, shows the free-form field only when no Localist location is set; if both are filled in, Localist always takes precedence and the free-form field is suppressed
- Adds `schema.org` markup to the location display for improved semantics/screen reader support

### Functional testing steps:
- [ ] Log in to the [multidev for testing](http://pr-1409-yalesites-platform.pantheonsite.io) `terminus drush yalesites-platform.pr-1409 -- uli`
- [ ] Test field changes
  - [ ] Edit the event [Office Hours with Mike](https://pr-1409-yalesites-platform.pantheonsite.io/node/44/edit)
  - [ ] Confirm "Yale Location" and "Room" now appear under the "Localist data" tab 
  - [ ] Confirm "Additional Information" appears in the "Basic info" tab
- [ ] Test new content visibility
  - [ ] In the "Additional Information", enter some sample text (e.g. "Meet in the lobby of the Schwarzman Center, 168 Grove St.") and Save
  - [ ] [Visit the saved node](https://pr-1409-yalesites-platform.pantheonsite.io/events/office-hours-with-mike) and confirm the Location section on the page shows your free-form text
- [ ] Test Localist backed events
  - [ ] Since this environment has no existing "Yale Location" terms, first [create one in the admin interface](https://pr-1409-yalesites-platform.pantheonsite.io/admin/structure/taxonomy/manage/event_place/add) (e.g. name it "Test Hall" with any Yale address).
  - [ ] Edit the event [Office Hours with Mike](https://pr-1409-yalesites-platform.pantheonsite.io/node/44/edit)
  - [ ] Click on the "Localist data" tab. , then select the new place in "Yale Location" on the event and Save.
  - [ ] Revisit the event page and confirm it now shows the Localist location(with address and map) instead of your free-form text. This confirms Localist takes precedence when both are filled in


dblanken-yale (changes requested):
I think this may need a little more work.  Below are a few things I ran into:

**Location and Room in the Localist data tab.** When I go into "Office Hours with Mike" I see the location and room moved to Localist data, however they look editable. [Mike's grooming note on #750](https://github.com/yalesites-org/YaleSites-Internal/issues/750#issuecomment-4329159479) says "lock and move 2 location fields into localist tab" — looks like the lock half didn't make it in. That matters because both are in [`overwrite_properties`](https://github.com/yalesites-org/yalesites-project/blob/develop/web/profiles/custom/yalesites_profile/modules/custom/ys_localist/migrations/localist_events.yml#L204-L219) in `localist_events.yml` (`field_event_room` L210, `field_event_place` L218), so anything typed there gets wiped on the next sync.

What makes it read oddly is the ordering. In [the form display](https://github.com/yalesites-org/yalesites-project/blob/feat-750/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.node.event.default.yml) the two moved fields land at weights 6 and 7, which puts them above all ten of the actually-synced fields (weights 29–37) *and* above [`field_localist_info`](https://github.com/yalesites-org/yalesites-project/blob/develop/web/profiles/custom/yalesites_profile/config/sync/field.field.node.event.field_localist_info.yml#L22) at weight 28--the markup that reads "These fields are disabled because they are synchronized from Localist." So they sit above that notice while being the only editable things in the tab.

**Microdata.** I'm not as familiar with the microdata aspect--looks neat. When I put it in [the validator](https://validator.schema.org/) I do see the Place entries. But I think that's the issue: per the [HTML microdata spec](https://html.spec.whatwg.org/multipage/microdata.html), an `itemprop` only becomes a property of the nearest ancestor carrying `itemscope`. Nothing establishes a [`schema.org/Event`](https://schema.org/Event) for it to attach to, so the `Place` comes through as a standalone item and `Event.location` never gets expressed — meaning there's no Event for search engines to pick up. You can confirm nothing establishes one with:

```
grep -rn --include='*.twig' -E 'itemscope|itemtype' web/ atomic/
```

That returns only the two lines the PR adds, at [L422](https://github.com/yalesites-org/component-library-twig/blob/feat-750/components/02-molecules/meta/event-meta/yds-event-meta-localist.twig#L422) and [L424](https://github.com/yalesites-org/component-library-twig/blob/feat-750/components/02-molecules/meta/event-meta/yds-event-meta-localist.twig#L424), and there's no `schema_metatag` or `rdf` enabled in `core.extension.yml` either. The free-form branch at [L453](https://github.com/yalesites-org/component-library-twig/blob/feat-750/components/02-molecules/meta/event-meta/yds-event-meta-localist.twig#L453) has only `itemprop="location"` with no enclosing `itemscope`, so that one drops entirely.

I tested this against two microdata extractors to be sure I wasn't misreading it. As shipped, the place branch yields a detached `Place` and the free-form branch yields nothing at all; adding `itemscope itemtype="https://schema.org/Event"` to the meta block wrapper makes both come out correctly as `Event` → `location`. That attribute pair on its own is enough — nothing else needed. Could we add it?

**`check_markup()` and cacheability.** Looking into [`check_markup()`](https://api.drupal.org/api/drupal/core%21modules%21filter%21filter.module/function/check_markup/10), it uses [`renderInIsolation()`](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Renderer.php/function/Renderer%3A%3ArenderInIsolation/11.x), which would disregard cacheability and attachments, right? My worry is MathJax and Linkit — and I think both are actually live on this field, since [`filter_ys_mathjax`](https://github.com/yalesites-org/yalesites-project/blob/develop/web/profiles/custom/yalesites_profile/config/sync/filter.format.basic_html.yml#L67) and [`linkit`](https://github.com/yalesites-org/yalesites-project/blob/develop/web/profiles/custom/yalesites_profile/config/sync/filter.format.basic_html.yml#L80) are both enabled on `basic_html`, and the field pins `allowed_formats: [basic_html]`.

MathJax attaches its library via [`setAttachments()`](https://git.drupalcode.org/project/mathjax/-/blob/4.1.0/src/Plugin/Filter/MathjaxFilter.php#L35), so math typed into this field wouldn't get the library and would render as literal `$$…$$`. Linkit calls [`addCacheableDependency($entity)`](https://git.drupalcode.org/project/linkit/-/blob/7.0.15/src/Plugin/Filter/LinkitFilter.php#L187), so we'd lose the linked-entity dependency and the text format's own cache tag — meaning edits to the format or the linked entity's label wouldn't invalidate event pages. Core's docblock on `check_markup()` recommends `#type: processed_text` instead for exactly this reason, so at [MetaFieldsManager.php L247](https://github.com/yalesites-org/yalesites-project/blob/feat-750/web/profiles/custom/yalesites_profile/modules/custom/ys_localist/src/MetaFieldsManager.php#L247) something like:

```php
$locationAdditionalInfo = [
  '#type' => 'processed_text',
  '#text' => $field_value['value'],
  '#format' => $field_value['format'] ?? 'basic_html',
];
```

Twig prints that identically and the metadata bubbles. This matches the existing `ResourceMetaBlock` pattern, so it's pre-existing and I'm fine if you'd rather handle it separately.

**One thing that will fail locally.** [`EventMetaBlock::build()` L100](https://github.com/yalesites-org/yalesites-project/blob/feat-750/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Plugin/Block/EventMetaBlock.php#L100) now reads `$eventFieldData['location_additional_info']`, but the fixture in [`EventMetaBlockTest.php` L58](https://github.com/yalesites-org/yalesites-project/blob/develop/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/tests/src/Unit/EventMetaBlockTest.php#L58) wasn't updated, so the existing test errors with `Undefined array key "location_additional_info"`:

```
lando php /app/vendor/bin/phpunit -c /app/phpunit.xml \
  /app/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/tests/src/Unit
```

On develop that's `OK (88 tests, 146 assertions)`; on this branch it's `88 tests, 142 assertions, 1 error`. CI won't catch it since `composer unit-test` is a stub. One line after `'place_info' => [],` fixes it:

```php
      'location_additional_info' => NULL,
```

dblanken-yale (changes requested):
Image

It looks as if the new Yale Location is actually in the Localist tab, when it probably shouldn't be right?  Also, when selecting a value in it, it doesn't save.  But even with it set to None, the additional info does not display instead; my understanding was that priorities would go to:

1.  localist room
2. yale location
3. additional info
4. nothing if all not set/blank

Image

Image

nJim:
@dblanken-yale I believe this is ready for review. I made the changes we discussed today including the change in always display the "Additional" information if it is populated.

miketullo95 (approved):
Looks good to me. The commits since your two review rounds address all four points:

- **Field locking**: `field_localist_event_experience`, `field_event_place`, and `field_event_room` are all in `fieldsToDisable` now ([ys_core.module:1120-1122](https://github.com/yalesites-org/yalesites-project/blob/feat-750/web/profiles/custom/yalesites_profile/modules/custom/ys_core/ys_core.module#L1120-L1122))
- **Cacheability**: `field_address_additional_info` renders via `#type: processed_text` now ([MetaFieldsManager.php:261](https://github.com/yalesites-org/yalesites-project/blob/feat-750/web/profiles/custom/yalesites_profile/modules/custom/ys_localist/src/MetaFieldsManager.php#L261)), so MathJax/Linkit keep their attachments and cache tags
- **Microdata**: the companion twig PR adds `itemscope itemtype="https://schema.org/Event"` to the wrapper, so the Place now attaches to an actual Event
- **Test fixture**: `EventMetaBlockTest.php` matches the current `getEventData()` keys

One dependency worth flagging: #1409 alone won't render the new Address/Additional Info fields on the front end — the actual display logic lives in component-library-twig#671, which hasn't been reviewed by anyone yet. These need to land together.

@dblanken-yale — given the timing, should this (and #671) go into the next RC cut, or wait and merge after the v2.26 branch is cut? #750 is milestoned to the 09-16 release, so want to make sure we're not pulling future-release work into an in-progress RC.

@nJim

dblanken-yale:
Review of the location field changes, grouped by priority.

## Blocking

### 1. Duplicate address rendering in the locked Layout Builder section
`config/sync/core.entity_view_display.node.event.default.yml:94`

The two new field blocks (`field_block:node:event:field_event_address` at weight 2, `field_block:node:event:field_address_additional_info` at weight 3) sit in the same "Title and Metadata" section as `event_meta_block`, which already renders both fields in its non-Localist branch. Any native event with an address renders the address and the additional info twice.

The section's `layout_builder_lock` value is 2 (`LOCKED_BLOCK_DELETE`) and no role holds `bypass lock settings on layout overrides`, so editors cannot remove the duplicates per node.

The field blocks also have no `is_localist_event` gate. On a Localist event where an editor filled the native fields, native address data renders next to the Localist location. Empty fields render nothing, so that case requires populated fields.

Suggest removing both components and letting `EventMetaBlock` own the rendering. Worth confirming they were intentional: no other node type's default display contains a plain `field_block:node:*`, and `field_event_place` / `field_event_room` have none.

### 2. field_event_place and field_event_room locked with no backfill
`modules/custom/ys_core/ys_core.module:1233`

Adding these to `$fieldsToDisable` locks them on every site, because `_ys_core_disable_event_fields()` has no `enable_localist_sync` check. Combined with the new `{% if is_localist_event %}` gate in the template, events with `field_event_place` set but `field_localist_id` empty lose their rendered location.

That covers hand-authored events on Localist-enabled sites, plus any event whose Localist ID was cleared. The location stops rendering, and the editor cannot clear or re-select the place term to work out why. The address has to be re-entered in the new field, per node, with no notification.

Stored data is safe: `FormBuilder` skips input processing for disabled elements and reuses `#default_value`. Sites that never ran the Localist migration are unaffected, since no role can create `event_place` terms.

Two options: gate the lock on Localist sync being enabled, or add a `hook_deploy_NAME` that copies `field_event_place` / `field_event_room` into the new fields. The diff currently has no update, post_update, or deploy hook.

## Worth fixing

### 3. Native address never renders on Localist events
`modules/custom/ys_localist/src/MetaFieldsManager.php:372`

`field_event_address` and `field_address_additional_info` are not in the disable list and sit in the unconditional `group_place` fieldset, so they stay editable on Localist events. Rendering is gated on `is_localist_event`, and the Localist branch only reaches `place` / `room` behind `{% if place or room %}`.

`localist_events.yml:147` maps `field_event_place` via `entity_lookup`, which resolves to nothing for a feed item with no matching place term, so a Localist event with no place or room is reachable. An editor who fills in Additional Address Information on one sees it render nowhere, with no feedback.

Either disable the native fields alongside the Localist ones, or fall back to them when `place` and `room` are both empty.

### 4. City/state/postal line assumes a US address format
`modules/custom/ys_localist/src/MetaFieldsManager.php:305` plus the CLT template

The field sets `available_countries: {}`, so any country is selectable, but the template renders `city, state postal` unconditionally. Countries whose address format omits `administrativeArea` or `postalCode` render empty spans.

Angola uses the generic format (no admin area, no postal code) and renders `Luanda, ` with a trailing comma and empty microdata. Belgium renders `Brussels,  1000`.

Suggest guarding each component individually in the template.

## Minor

### 5. No test coverage for the new logic
`modules/custom/ys_layouts/tests/src/Unit/EventMetaBlockTest.php:58`

The three new fixture keys are never asserted; they only keep `EventMetaBlock::build()` from hitting undefined keys. `MetaFieldsManagerTest` covers `stripEmptyParagraphs` only, and nothing covers `getEventData()`. A unit test across Localist/native and address present/absent would cover items 3 and 4.

### 6. Field config appears hand-written
`config/sync/field.field.node.event.field_event_address.yml:17`, `field.field.node.event.field_address_additional_info.yml:19`

Both use `translatable: null`. None of the 381 existing `field.field.*.yml` files use `null` (246 are `false`, 135 are `true`). `field_event_address` also uses `langcode_override: null` where `field.field.taxonomy_term.event_place.field_address.yml:22` uses `''`. Both values are falsy and pass schema validation, so this is cosmetic. Re-exporting through `npm run confex` would normalise them.

### 7. Handler change looks droppable
`config/sync/field.field.node.event.field_event_place.yml:20`

`default:filter_existing_terms` behaves identically to `default:taxonomy_term` unless a `filter` setting is present, and none is written here. 8 existing configs already use this handler without declaring `taxonomy_manager`, so the undeclared dependency is consistent with the rest of the repo. Probably safe to revert the line unless the switch was deliberate. The `Yale Location` to `Localist location` relabel in the same hunk reads as intentional.

## PR description

The testing steps say to confirm "Yale Location" and "Room" appear in the Basic info tab, but the form display moves them into the Localist data group and they are now read-only. The later step asking the tester to select a Yale Location term will not be possible. Worth updating before QA runs it.

dblanken-yale:
**TL;DR:** all six items from the [8 Sep review](https://github.com/yalesites-org/yalesites-project/pull/1409#issuecomment-5589990620) are fixed and pushed, plus the description and testing steps. Nothing is waiting on a decision. The template half landed on component-library-twig#671 (`dac8a72f`), which changed after its approval, so it needs another look too.

**One call I made on item 1:** the meta block now owns both fields, and I removed the two field blocks. Existing events carry layout overrides that never render default-layout blocks, so only the meta block reaches every event. Additional Address Information follows the 27 Aug always-display note. If you would rather keep the field blocks, the change to reverse is the view display file in `8e1792525`.

<details>
<summary><b>Per-item resolution</b></summary>

1. **Duplicate field blocks.** Both blocks are gone from `core.entity_view_display.node.event.default.yml`. The template now renders Additional Address Information on Localist and native events. It is labelled "Location" with `itemprop="location"` when nothing else rendered, otherwise "Additional Address Information". The native address still only shows on non-Localist events.
2. **No backfill.** New `ys_localist_deploy_10001` in `ys_localist.deploy.php`. For events with no `field_localist_id` and a place or room, it copies the place term's address into `field_event_address` and writes "place name, room" into `field_address_additional_info`. It never overwrites a filled field, covers the default and latest draft revision, is batched, and is idempotent. Kernel test: `EventLocationBackfillDeployTest` (6 tests).
3. **US-only address line.** New CLT partial `_yds-event-meta--locality.twig` prints city, region and postcode only when present. Both the Localist and native branches use it. The Belgium case now renders `Brussels 1000`.
4. **Coverage.** The new logic moved into `isLocalistEvent()`, `getEventAddress()` and `getAddressAdditionalInfo()` on `MetaFieldsManager`, with 5 new tests in `MetaFieldsManagerTest`.
5. **Field config.** Both new fields now have `translatable: true`, and `langcode_override: ''`. Stable through `confim` then `confex`.
6. **Handler.** Reverted to `default:taxonomy_term`. The "Localist location" label stays.
7. **Description and testing steps.** Rewritten for the always-display behavior and the locked fields.

</details>

<details>
<summary><b>How it was checked</b></summary>

- On local Lando after `confim` and the deploy hook, I checked six events: native US, native Belgium, additional info only, Localist with place and room (plus a native address that must not show), Localist with info only, and a hand-authored event with place and room that the hook backfilled. On each page every value appears once, with no empty microdata spans.
- `MetaFieldsManagerTest` 24/24, `EventMetaBlockTest` 11/11, `EventLocationBackfillDeployTest` 6/6.
- `composer code-sniff` clean on both standards. CLT `npm run lint` clean.
- Not checked: the Storybook render of the template (only exercised through Drupal).

</details>

dblanken-yale:
**TL;DR:** Ran the 13 functional testing steps on the multidev. 12 pass. 1 needs a human: "Office Hours with Mike" cannot be compared to develop because people have edited it by hand on this multidev.

| # | Step | Result | Evidence |
|---|---|---|---|
| 1 | Log in to the multidev | pass | Logged in with a one-time `uli` link |
| 2 | Create a new event | pass | [node/add/event](https://pr-1409-yalesites-platform.pantheonsite.io/node/add/event) |
| 3 | "Address" and "Additional Address Information" in the "Location" fieldset on "Basic info" | pass | Both are in the Location fieldset |
| 4 | "Localist location", "Localist room" and "Event Experience" under "Localist data", not editable | pass | All three are in the Localist data tab and disabled |
| 5 | Fill in a US address and additional info, then save | pass | [node/85](https://pr-1409-yalesites-platform.pantheonsite.io/node/85) |
| 6 | Address and additional info each show once | pass | Each appeared exactly once on the whole page |
| 7 | Change the address to Belgium and save | pass | [node/85](https://pr-1409-yalesites-platform.pantheonsite.io/node/85) |
| 8 | Second line reads "Brussels 1000" with no stray comma | pass | Rendered as `Brussels 1000` |
| 9 | Create an event with only additional info | pass | [node/86](https://pr-1409-yalesites-platform.pantheonsite.io/node/86) |
| 10 | It shows once, labelled "Location" | pass | [node/86](https://pr-1409-yalesites-platform.pantheonsite.io/node/86) |
| 11 | Mark the first event as Localist-synced | pass | Set `field_localist_id = test-1` on node 85 with the terminus command from the PR |
| 12 | Native address hidden, additional info still shows, labelled "Location" | pass | [node/85](https://pr-1409-yalesites-platform.pantheonsite.io/node/85) now shows only "Location: Meet in the lobby" |
| 13 | [Office Hours with Mike](https://pr-1409-yalesites-platform.pantheonsite.io/events/office-hours-with-mike) looks the same as on develop | **needs a human** | The page differs, but the cause is the data, not the code. See below. |

<details><summary>Step 13 detail</summary>

On the multidev, node 44 shows a Location block ("123 Main Street, Hamden, CT 06518") and an "Additional Address Information" block whose content is the same words. [On dev](https://dev-yalesites-platform.pantheonsite.io/events/office-hours-with-mike) it shows neither.

The node's revisions on the multidev show people editing it by hand from July through September. Revision 332 (uid 1) set the Localist location to "Example Place", and revision 343 (uid 7, 2026-09-01) typed in the native address. On dev the node was last changed in 2024 and has no location data. Given this data, the PR renders node 44 the way the description says it should: no Localist ID, so the native address and additional info show.

As a substitute, the three existing events nobody edited on the multidev render the same text on the PR as on dev: `/events/dinner-with-dad`, `/events/2036-07-03-michaelcon`, `/events/2036-12-05-event-for-visual-regression-testing`.

To check this by hand: clone content from dev into the multidev and compare Office Hours with Mike again, or clear the test values on node 44.
</details>

Side notes, not failures:
- Belgian addresses usually put the postcode first ("1000 Brussels"). The step expects "Brussels 1000", and that is what renders.
- The location markup is `itemprop="location"` typed directly as `PostalAddress`, with no `Place` wrapper. Schema.org allows that, but it is not quite the Event/Place/PostalAddress nesting the description mentions.

<sub>Proven against `8e17925` on https://pr-1409-yalesites-platform.pantheonsite.io via playwright-cli (run through `npx @playwright/cli`). Content created by this run is prefixed `prove-pr 1409:` (nodes 85 and 86).<br>
Run history: `8e17925` 12 pass 1 unprovable. Strikes: 0.</sub>

PR #495, opened 2023-11-28
YALB-1644 - Dial: Center-align callout with content-width content

joetower:
### Description of work
- Adds alignment variable for `callout__alignment`
- Adds a default of `center` - existing callouts will render the same
- Adds a `field_style_alignment` field in Drupal, default is `center`
- Wires the `callout__alignment` variable to `field_style_alignment`
- You can choose between `center` and `left`. `center` uses the `site` width. `left` uses `content` width to restrict the callout inner wrapper width.

### Functional testing steps:
- [x] Review this page with two callouts: https://pr-495-yalesites-platform.pantheonsite.io/welcome/customize-with-confidence
- [x] Verify the callout container is left-aligned and the content area of the callouts is limited to the left-aligned content
Screenshot 2023-11-28 at 11 53 48 AM

- [x] Login and review another test page here - single callout: https://pr-495-yalesites-platform.pantheonsite.io/testing-all-of-the-things 
- [x] Verify it is set to `left` and toggle between `left` and `center` and verify the callout changes.

https://github.com/yalesites-org/yalesites-project/assets/366413/7ffa0d73-fc07-4d37-bb65-814f98585148

PR #1413, opened 2026-07-27
1435: Update Storybook to latest version and move to Vite

nJim:
### Description of work
- No changes needed in this repo for the migration itself as atomic consumes component-library-twig directly. This PR exists solely to give the branch-matching CI build
  - [Component Library Changes](https://github.com/yalesites-org/component-library-twig/pull/672)
  - [Atomic Theme Changes](https://github.com/yalesites-org/atomic/pull/495)

### Functional testing steps:
- [ ] Confirm the multidev preview builds successfully
- [ ] Spot-check a few pages against production for visual parity
- [ ] See the component-library-twig and atomic PRs for their detailed test plans


dblanken-yale:
Brought this branch current with its base and pushed it last, so the multidev builds against the companion work.

### Reconcile

This branch was 78 commits behind `after-2260`. Per Dave, `after-2260` was first fast-forwarded to `develop` (it was a strict ancestor — no merge commit, no rewritten history), then merged in here. Zero conflicts, and no `hook_update_N` collisions since this branch touches no `.install`/`.post_update.php`. The base relocated `ys_core/images/favicons/*` and `YaleSitesMediaManager` into a new `ys_media` module; checked that against this branch's `CoreTwigExtension` change — no interaction, that diff is docblock-only.

No code changes were needed here. Two notes for the review:

### The description is slightly out of date

It says no changes are needed in this repo, but the branch does carry four: `.ci/build/build_frontend` (dropping the dead `webpack-cli` install), `config/sync/addtoany.settings.yml` (sprite path), and `ys_core/src/CoreTwigExtension.php` plus its unit test. All four verified:

- Dropping `npm install webpack-cli` is safe — `@emulsify/core` is a production dependency and brings `vite`, `sass`, and `vite-plugin-svg-sprite` transitively, so `npm ci --omit=dev` still yields a working toolchain.
- The addtoany path is correct in all 5 occurrences and matches the sprite's real location (`dist/assets/icons.svg`, confirmed in a fresh build). Existing sites pick it up on deploy — it's plain profile config and `drush deploy` runs `config:import` after `updatedb`, so no update hook needed. One thing to expect rather than be surprised by: the import overwrites `additional_html` on any site where an editor hand-edited it through the AddToAny admin UI. That's necessary here to fix the path.
- `composer code-sniff` (Drupal + DrupalPractice) is clean on both changed PHP files.

### Asset versioning is a real, unfixed regression — flagging, not fixing

`getAssetPath()` read webpack's `dist/manifest.json` for cache-busting. The Vite build emits no manifest, so it's now a guaranteed passthrough. Scope is narrower than "assets are unversioned": everything in `atomic.libraries.yml` still gets Drupal's `?<system.css_js_query_string>`, bumped on every cache rebuild. The exception is `icons.svg`, referenced from `_yds-icon.twig` as a raw `<use xlink:href>` that Drupal never rewrites — so that URL is now permanently stable, and it's on essentially every page.

After a release that adds or changes an icon, any browser or Pantheon CDN node holding the old sprite renders new icons as empty space (a `<use>` pointing at a missing `#id` fails silently — no console error, no fallback) and shows old artwork for changed ones. Hard-refresh fixes it for one user, which makes it a "works for me" support ticket.

A cheap fix needing no build-side support is to append the existing `system.css_js_query_string` state value to the returned filename, injected via the state service. I left it alone because it changes runtime output on every page and the mitigation call is yours. Two smaller related notes: the docblock says restoring hashing "is tracked separately" but cites no issue number, and `testGetAssetPathReturnsVersionedNameFromManifest` / `testGetAssetPathFallsBackToInputWhenAssetNotInManifest` are now permanently skipped everywhere, leaving that branch at zero coverage while reading as covered in reports.

### Storybook parity, since that's the acceptance bar

Compared the rebuilt Storybook against the published production build in a browser, expanding every sidebar node and reading the tree in DOM order: **138 items identical with Visreg off, 769 identical with Visreg on**, after fixing a story-ordering regression in component-library-twig #672. Toolbar controls match production: "Site: Global Theme (lever)", "Headings: YaleNew (Old-Style Numerals)", "Show/Hide Visreg Pages". Details in the #672 comment.

Merge last, after component-library-twig #672 and atomic #495.

PR #1419, opened 2026-07-29
1479: AI Tester Compare: UX Improvements & Export for AI Analysis

dblanken-yale:
### Description of work

- Builds on yalesites-org/YaleSites-Internal#1470 (PR #1418) — this branch is stacked on that sub-issue and targets its branch, so please review/approve that PR first. GitHub will retarget this one to `551-beacon-drupalai` once #1418 merges.

**Accessibility fixes** — every ratio below was measured on the rendered page in both colour modes, not estimated:

- **Dark-mode caveat contrast.** The yellow cross-assistant caveat rendered at **1.02:1** in dark mode (light grey `rgb(210,211,211)` on yellow `rgb(239,207,100)`) — effectively invisible. Cause: it paired Gin's `--gin-color-text`, which flips light for dark mode, with a background that does not flip. The text colour is now pinned and dark mode gets its own dark surface. Now **10.6:1**.
- **Strikethrough removed** from the Run A diff highlights. The background alone carries the signal now.
- While in that stylesheet, three further pairs were below AA and are fixed. Two of them the ticket does not name — flagging that explicitly since it is slightly beyond the literal ACs, but AC 8 asks for the view's contrast to be validated against WCAG 2.1 AA and these are colour values in the same file:
  - diff highlights: 3.74:1 light / 2.99:1 dark → **7.4:1 / 8.3:1**
  - `differs` badge: 4.12:1 → **6.3:1**
  - only-in-Run-A/B badge: ~4.0:1 once its translucent background composites → **8.0:1**

**Additions:**

- **Help text + colour legend** explaining that a highlighted span is wording absent from the other run.
- **"Export for Clarity" modal**, built on core's `use-ajax` dialog so the focus trap and Escape handling come from the platform rather than hand-rolled JS. Contains the upload instructions, a link to Clarity, a research prompt, a primary **Download comparison JSON** button and a secondary **Copy prompt** button. The prompt interpolates this comparison's real run ids, each side's assistant label and the question count; its copy lives in a Twig template since it is editorial text refined independently of the code.
- **Reuses the existing `ys_ai_tester.compare_json` route** for the download rather than adding a second export path (AC 6).
- **Small-screen block.** Below `768px` the four-column table is replaced by a notice asking for a larger screen. `768px` is the breakpoint the sibling `ys_beacon` admin stylesheets already use — there is no admin breakpoint token in this repo. Scoped to the compare view alone; the rest of the AI Tester stays usable at small widths.

### Not implemented — one AC deliberately skipped

The ticket also asks for a **hover/keyboard tooltip on each diff highlight**. That is **not** in this PR, by decision during scoping. A paragraph-length LLM answer produces dozens of changed-word spans per cell, so making each one focusable would add tens of tab stops per table row and make keyboard traversal of the comparison materially worse than a repeated per-word tooltip is worth. The help text and legend give that explanation once for the whole table instead. Happy to revisit if you'd rather have the literal per-span tooltip.

Two smaller notes:

- The small-screen block hides the comparison from screen-reader users on narrow viewports too, since a screen reader would otherwise linearise the table fine at any width. That is the direct consequence of the AC as written; say the word if it should be a collapse rather than a replace.
- Docs: no user-facing doc updates, per the ticket's assumption that this is an internal admin tool tied to a temporary feature.

### Functional testing steps:

- [ ] Open a cross-assistant comparison at `/admin/config/yalesites/ys-beacon/tester/compare/{run_a}/{run_b}` (one Beacon run, one legacy run) and confirm the yellow caveat copy is readable.
- [ ] Switch the admin theme to dark mode (Gin `Dark mode`, or set your OS to dark since the setting is `auto`) and confirm the caveat is still clearly readable rather than washed out.
- [ ] On a row marked **Differs**, confirm the Run A highlights are **not** struck through and are comfortably readable in both light and dark mode.
- [ ] Confirm the help text and the two-swatch legend appear above the table and describe the highlighting.
- [ ] Click **Export for Clarity**. Confirm the modal opens with instructions, the prompt, and two buttons.
- [ ] In the modal, confirm the prompt names the correct run numbers, both assistant labels, and the question count.
- [ ] Press <kbd>Tab</kbd> repeatedly and confirm focus stays inside the modal and cycles; press <kbd>Esc</kbd> and confirm it closes and focus returns to the **Export for Clarity** button.
- [ ] Click **Copy prompt**, confirm the "Prompt copied to the clipboard." status appears, and paste somewhere to confirm the full prompt came through.
- [ ] Click **Download comparison JSON** and confirm a `compare-{a}-{b}.json` file downloads containing `run_a`, `run_b`, `summary`, `pairs`.
- [ ] Narrow the browser below 768px (or open on a phone) and confirm the comparison is replaced by the "needs a larger screen" notice.
- [ ] At that same narrow width, open `/admin/config/yalesites/ys-beacon/tester` and confirm the rest of the AI Tester is **not** blocked.

**Question tabs** (added after review — these replace the four-column table):

- [ ] On the same comparison, confirm the four-column table is gone, replaced by a vertical list of question tabs on the left with the selected question's two runs side by side on the right.
- [ ] Confirm each tab shows the question number, a truncated preview of the question text, and its status as a visible badge (Identical / Differs / Only in Run A / Only in Run B) rather than by colour alone.
- [ ] Click a tab and confirm only that question's panel is shown, and that each column heading names that run's source file with its run id and assistant.
- [ ] Put focus on a tab and press <kbd>↓</kbd> and <kbd>↑</kbd>: confirm selection moves one question at a time and wraps at both ends. Press <kbd>Home</kbd> and <kbd>End</kbd> and confirm they jump to the first and last question.
- [ ] Press <kbd>Tab</kbd> while a tab has focus and confirm focus leaves the tab list rather than stepping through every tab.
- [ ] Reload and confirm the page does not visibly flash a fully expanded list of every answer before collapsing to a single panel.
- [ ] Switch the admin theme to dark mode and confirm the tab list, the selected tab and the panel are all clearly readable.
- [ ] Reload with JavaScript disabled and confirm every question's answers are still readable as a stacked list, rather than the right-hand column going blank.
- [ ] Open a comparison where a question was only asked in one run and confirm the other column still reads "— not asked in this run —".

**AI analysis export** (added after review — replaces the Clarity-specific export):

- [ ] On a comparison, confirm the third header button now reads **Export for AI analysis** rather than "Export for Clarity".
- [ ] Click it and confirm the modal still opens, traps focus, and closes on <kbd>Esc</kbd>.
- [ ] Confirm the modal copy no longer instructs you to upload to Clarity: it should say to attach the file to an AI assistant, note that the export carries published site content only, and mention Clarity as one option with the caveat that a comparison of any size may exceed what it accepts.
- [ ] Confirm the prompt textarea now contains a "How the file is structured" section listing `run_a`/`run_b`, `summary` and `pairs`, and that `citations` is described as nested inside each side (`a`/`b`) rather than on the question.
- [ ] Press **Copy prompt** and paste it somewhere: confirm the whole prompt including the structure section arrives, and that the run ids, both assistant labels and the question count match the comparison you opened.
- [ ] Press **Download comparison JSON** and open the file. Confirm no citation carries a `content` key, and that each one still has `number`, `title`, `url`, `excerpt` and `cited`.
- [ ] If you have a download of the same comparison from before this change, compare sizes: the new file should be substantially smaller, and its size should no longer track how long the cited pages happen to be.
- [ ] Confirm the compare view's plain **Download JSON** button produces the same leaner file — it is deliberately the same route, so this is expected rather than a regression.
- [ ] Open a comparison where a question was asked in only one run and confirm that side is `null` in the JSON rather than an empty object.
- [ ] End-to-end, worth doing once: attach the downloaded JSON to an AI assistant with the copied prompt and confirm it parses the file and answers the four numbered tasks without asking where the citations are.

**Run colours and module-wide accessibility** (added after review — Run A green / Run B blue, plus the axe pass):

- [ ] On a question marked **Differs**, confirm Run A's highlights are now **green** and Run B's are **blue** (this is the reverse of what the earlier steps above described, and no red is left in the palette).
- [ ] Confirm the legend matches what the panels show: a green swatch beside "Only in Run A" and a blue swatch beside "Only in Run B". If those are ever swapped relative to the answer columns, that is the bug the new unit test guards.
- [ ] Switch to dark mode and confirm both highlight colours are still comfortably readable, and still green for A / blue for B.
- [ ] On `/admin/config/yalesites/ys-beacon/tester`, turn on a screen reader (VoiceOver: <kbd>⌘F5</kbd>) and move to a Run History checkbox. Confirm it now announces which run it selects — e.g. "Update Run #1, questions.txt, 07/29/2025 - 10:40" — rather than an unlabelled checkbox. This was the critical finding.
- [ ] Confirm the Run History table looks unchanged: adding that label must not have introduced an extra column.
- [ ] Select two runs and compare, to confirm labelling the checkboxes did not break selection.
- [ ] Trigger a validation error on that page (e.g. submit with both a file and typed questions) with runs ticked, and confirm the ticks survive the rebuild.
- [ ] On a comparison, put keyboard focus on a question tab (click a tab, then press <kbd>↓</kbd>) and confirm there is now a clearly visible blue **outline** around the focused tab, not just a faint grey halo. Check in light mode especially — that was the weak case.
- [ ] Press <kbd>Tab</kbd> from a tab into the open panel and confirm the panel also shows a visible focus outline.
- [ ] Known and deliberately not fixed: the Run History table's checkbox column has an empty header cell. That is core's output for any table without a "select all", and is a best-practice note rather than a WCAG 2.1 AA failure — see the PR comment if you would like it changed.

**Run B blue and the question heading** (added after review — the two items from the latest round):

- [ ] On a question marked **Differs**, confirm Run B's highlights are **blue** and that **no red appears anywhere** in the comparison. Run A stays green.
- [ ] Confirm the legend's second swatch is blue and still reads "Only in Run B", matching the colour in the right-hand panel.
- [ ] Switch the admin theme to dark mode and confirm both highlight colours are still comfortably readable — green for Run A, blue for Run B, no red.
- [ ] Confirm the question at the top of each panel (e.g. "How do I apply for financial aid at Yale?") now reads as a headline: clearly larger than the "Run A: ..." / "Run B: ..." headings beneath it, rather than a pixel bigger.
- [ ] Sanity check the two blues are comfortably distinguishable *to your eye*: Run B's highlight blue is intentionally not the same blue as the "Only in Run A/B" status chips in the tab rail or the selected tab's background. If they read as the same colour, say so — widening the gap is a two-value change.
- [ ] Narrow the browser to roughly 800-950px and confirm the larger question heading still sits below the page title in the visual hierarchy and does not wrap or overflow its panel.

**Two-column layout after removing the `columns` knob** (added after review):

- [ ] Open a comparison of two runs. Confirm each question's panel still shows Run A and Run B **side by side in two equal columns**, not stacked one above the other. This is the one thing the change could plausibly have broken.
- [ ] Narrow the browser to roughly 1000px. Confirm the two answers now **stack vertically** while the question tab list stays vertical on the left.
- [ ] Narrow below 768px and confirm you still get the "use a larger screen" notice instead of the tabs.
- [ ] Click through several question tabs and confirm each panel's two columns are still even, with neither side overflowing.
- [ ] With focus on a tab, press <kbd>↑</kbd>/<kbd>↓</kbd>, <kbd>Home</kbd> and <kbd>End</kbd> and confirm tab selection still moves and the matching panel still swaps in. Nothing about the keyboard behaviour changed, but the template did, so it is worth one pass.
- [ ] No visual change is expected anywhere from this commit. If the panels look different from the previous build, that is a bug.

References yalesites-org/YaleSites-Internal#1479

---
Created with AI

**The single-run view now uses the question tabs** (added after review — this is the layout change, plus the always-on scrollbar and the "Question X of N" heading):

- [ ] Open a single run at `/admin/config/yalesites/ys-beacon/tester/{run_id}`. Confirm the old **Question / Answer / Sources** table is gone, replaced by the same vertical question rail on the left and one answer panel on the right — the comparison layout minus the Run B column.
- [ ] Confirm the answer now gets the **full panel width** rather than a third of it, and that the sources are a readable list rather than a narrow column wrapping one character of the URL per line. That crushed table was the defect this fixes, not just an inconsistency.
- [ ] Confirm each panel heading reads **"Question X of N: [the question]"** — e.g. "Question 1 of 19: What are the Yale library hours?" — and that the number and total are correct for the run.
- [ ] Confirm the same heading format now also appears on the **comparison** view (both views share the widget), and that the full question text is still shown in the panel even when the tab preview is truncated.
- [ ] Confirm each tab carries a status badge: **Answered** (green), **No answer** (yellow), or **Error** (red). Red appears only for a recorded backend error — the diff palette still has no red in it.
- [ ] Open a run that has a failed question and confirm the panel shows **"Error: {message}"** in place of an answer, and that a question that came back blank shows **"Empty answer"**.
- [ ] Switch the admin theme to dark mode and confirm that error/empty line is clearly readable. It previously rendered at 3.7:1 in dark mode, which was below AA — that is fixed here.
- [ ] Confirm the panel shows a **Sources** heading with every retrieved source and whether each was cited. The comparison only lists sources unique to one side, so this is the run view's own information and must not have been dropped.
- [ ] Confirm the run header above the rail is unchanged: **Run #N — date — File — Assistant — Status**, plus **Download JSON**, **Download CSV**, **Download questions (.txt)** and **Back to tester**. Click each download and confirm it still works.
- [ ] Put focus on a tab and press <kbd>↓</kbd> / <kbd>↑</kbd> (wrapping at both ends), then <kbd>Home</kbd> and <kbd>End</kbd>. Confirm the keyboard model matches the comparison's exactly.
- [ ] Confirm the tab rail now sits on its own **bounded surface with a border**, in both colour modes, on both the run and compare views.
- [ ] On a run with more questions than fit the rail (roughly 15+), confirm a partially-clipped tab at the bottom edge makes it obvious the list continues. **Note:** on macOS with a trackpad the scrollbar itself still fades when idle — that is an OS behaviour no CSS can override, which is why the container edge and the "of N" heading carry the cue. On Windows/Linux (or macOS set to "always show scroll bars") the scrollbar should be visible at rest.
- [ ] Confirm adding or removing questions never shifts the tab text sideways (the scrollbar's space is reserved).
- [ ] Narrow the browser below 768px on the **run** view and confirm the rail moves **above** the panel at full width and stays usable — the run view must not show the comparison's "needs a larger screen" notice, since a single answer column works fine there.
- [ ] Confirm the **comparison** still replaces itself with that notice below 768px, and still shows two equal side-by-side columns at full width.
- [ ] Open a run whose batch has not finished and confirm it still says "No results yet — batch may still be processing." rather than rendering blank.
- [ ] Reload the run view with JavaScript disabled and confirm every question's answer is readable as a stacked list rather than the panel area going blank.

**Scrollbar visibility and the per-side sources list** (added after review — the corrections):

- [ ] On a run or comparison with more questions than fit the rail (roughly 15+), confirm the tab rail now shows a **visible scrollbar at rest**, without needing to scroll first. This is the item my earlier measurement got wrong, so it is worth a real look rather than a glance.
- [ ] Confirm the scrollbar thumb is clearly distinguishable from its track in **both** light and dark mode.
- [ ] Confirm the tab text does not shift sideways when a run has few enough questions that the rail does not overflow (the scrollbar's space is reserved either way).
- [ ] In the **comparison**, open a question where the two runs retrieved different sources. Confirm each panel now lists **every** source that run retrieved, not just the ones unique to it.
- [ ] Confirm each source is flagged either *cited* or *retrieved, not cited*, so a document that came back but went unused is visible rather than absent.
- [ ] Confirm sources no other run retrieved carry an **"only in this run"** chip, and that a source both runs retrieved carries **no** chip in either column.
- [ ] Confirm the per-side meta line still agrees with the list — e.g. "2 of 3 sources cited" above a list of three sources with two marked *cited*.
- [ ] Confirm the old "Sources only here:" summary line is gone and that you are happy with its information living inside the list instead. If you want the summary line back as well, that is additive — say so.
- [ ] On the **single-run** view, confirm the Sources list shows every retrieved source with the same cited/not-cited flags, and **no** "only in this run" chips (there is no other run to be unique against).
- [ ] Switch to dark mode and confirm the source links, the flags and the chips are all readable in both views.

**Generic AI export copy** (added after review — the Clarity naming is gone):

- [ ] On a comparison, click **Export for AI analysis**. Confirm the modal's second paragraph reads only that the export carries published site content and that any assistant approved for work use is a reasonable choice, followed by a tool-neutral caveat that a comparison of any real size may exceed what a single conversation accepts.
- [ ] Confirm there is **no** mention of Clarity anywhere in the modal and **no** "Open Clarity" link. **This supersedes the earlier step above** that asked you to confirm Clarity was mentioned as one option with a caveat — that step is now wrong, and the copy naming no tool at all is the expected result.
- [ ] Confirm the modal still has exactly two actions, **Download comparison JSON** and **Copy prompt**, and that removing the link did not disturb their layout.
- [ ] Press <kbd>Tab</kbd> through the modal and confirm focus moves from the prompt textarea straight to the two buttons, with no leftover link stop between them.
- [ ] Press <kbd>Esc</kbd> and confirm the modal still closes and focus returns to the **Export for AI analysis** button.
- [ ] Click **Copy prompt**, paste it somewhere, and confirm the prompt itself is unchanged — in particular that its third numbered task still reads "accuracy, completeness, clarity". That lowercase word is the plain English one for answer quality and was deliberately kept.
- [ ] Switch the admin theme to dark mode and confirm the reworded paragraph is comfortably readable. No colour changed, but it is one glance.


miketullo95 (changes requested):
@dblanken-yale Answering your two questions first, then the bigger thing.

## Your two open questions

**The per-span diff tooltip: your call stands, don't build it.** Your reasoning holds. A paragraph-length answer produces dozens of changed-word spans, and turning each into a tab stop makes keyboard traversal of the comparison materially worse than a repeated per-word tooltip is worth. The help text and the legend explain it once for the whole table, which is the better trade. I'm treating that AC as satisfied by the explanation rather than unmet.

**The small-screen block: keep it as a replace.** That is the AC as written and I'm happy with it. The screen-reader consequence on narrow viewports is real but acceptable for an internal admin tool, and I'd rather not add a collapse variant now given the layout is about to change (see below).

So neither of those is what's holding this up.

## Why I'm requesting changes

The compare view this PR polishes is about to be re-skinned. Kevin designed a full AI Tester experience in Figma, and we want to adopt its main compare UI: vertical question tabs on the left, two answer panels side by side on the right, source chips that open a popover saying whether a source was cited or only retrieved.

Rather than merge polish on the four-column table and then immediately replace that table, I'd like Phase 1 below folded into this branch so the layout lands once.

**Design:** [YaleSites AI Tester, Figma](https://www.figma.com/design/GuOZcTkDIHpd5iwxHHpb50/YaleSites-AI-Tester?node-id=138-2175)

### Important context on the design

Kevin drew these when AI Tester was very basic. You kept building on a separate track and the prototype has moved well ahead of the designs on features. **This is not a fresh design handoff. It's us bringing a slightly outdated design forward into the shape AI Tester actually has today.** Where the design and the code disagree, it's almost always because the code moved after the design was drawn. Assume your code is current unless I say otherwise below.

### Almost none of this PR is wasted

Worth saying plainly, because "we're replacing the table" reads worse than it is. What carries forward untouched:

- every measured contrast pair and the `html.gin--dark-mode` override block
- the `--ys-diff-removed-*` / `--ys-diff-added-*` custom properties, which are exactly the hook the toggle in Phase 4 needs
- the `.ys-compare-badge--*` classes, which move onto the tabs
- `.ys-compare-mobile-notice` and the 768px media query
- the entire Export for Clarity modal, route, template, and JS
- `crossAssistantCaveat()`, `diffHelp()`, `legendItem()`, `sideCell()`

What changes is the wrapper around `sideCell()`, not the logic inside it.

---

## The plan

Five phases. Phase 1 is the one that has to land. Everything after it is additive and can ship separately.

### Phase 1: tab shell and side-by-side panels

Replace the four-column table in `AiTesterController::compare()` with the tabs-plus-panels layout.

**Theme hook and template.** Add `ys_ai_tester_question_tabs` to `ys_ai_tester_theme()`, with `templates/ys-ai-tester-question-tabs.html.twig`. Variables: a `tabs` array (label, status, preview, panel id) and a `panels` array. Make the panel count a variable so the run view can use one panel and compare can use two.

**Markup and ARIA.** Hand-roll it. Core's `#type => vertical_tabs` is built for form `details` groups and will fight the design.

```
<div class="ys-qtabs">
  <div class="ys-qtabs__list" role="tablist" aria-orientation="vertical">
    <button role="tab" id="qtab-0" aria-controls="qpanel-0" aria-selected="true"  tabindex="0">…</button>
    <button role="tab" id="qtab-1" aria-controls="qpanel-1" aria-selected="false" tabindex="-1">…</button>
  </div>
  <div class="ys-qtabs__panels">
    <div role="tabpanel" id="qpanel-0" aria-labelledby="qtab-0" tabindex="0">…</div>
    <div role="tabpanel" id="qpanel-1" aria-labelledby="qtab-1" tabindex="0" hidden>…</div>
  </div>
</div>
```

Roving tabindex, Up and Down move between tabs, Home and End jump to first and last. Standard APG tabs pattern. New behavior in `js/question-tabs.js`, same `Drupal.behaviors` and `once` shape as `js/compare-export.js`.

**No-JS fallback.** Render every panel visible and unhidden, let the JS add `hidden` on attach. If the behavior never runs it degrades to a readable stacked list instead of a blank right column.

**Tab labels carry the status.** The table had a Status column and the tab layout loses it, so status moves onto the tab: question number, a truncated preview of the question text, and the pair status as a visible badge reusing your existing `.ys-compare-badge--*` classes. Keep the text label, not just color.

Kevin's compare tabs read only "Question 1". Adding the preview and the badge is a deliberate divergence and the right call. Someone scanning 25 questions needs to find the ones that differ without clicking through every tab.

**Panels.** Reuse `sideCell()` as-is. It already handles the null side, the error case, the empty answer, the meta line, and the unique-sources list. Keep the panel heading as `Run A: {source_filename}` per the design, and keep the run id and assistant label in it since cross-assistant comparison is the point of #1470.

**What stays above the tabs.** Summary line, cross-assistant caveat, both run meta blocks, download buttons. `crossAssistantCaveat()` in particular is load-bearing for #1470.

**CSS.** New `css/question-tabs.css`, new `ys_ai_tester/question_tabs` library. Keep it separate from `compare.css` so the run view can attach it without pulling in the diff colors. Kevin's design is drawn light-mode only, which is fine, our admin theme follows the user's preference. Just keep the pattern `compare.css` already uses and carry on recording measured ratios in comments.

### Phase 2: source chips and citation popover

Build once, use in both `run()` and `compare()`. This replaces `buildCitationsCell()`.

**Good news: this is already fully backed by stored data.** `CitationFormatter::format()` writes `['number', 'title', 'url', 'content', 'excerpt', 'cited']` per citation. That `cited` boolean is exactly the Cited vs Retrieved distinction the popover needs, and `excerpt` gives you 300 characters of source text for the card. Nothing new to plumb.

**Where it lives.** New `src/CitationRenderer.php` service. The controller is already 887 lines and this is worth unit testing on its own.

**Chip markup.** A `<button type="button" aria-expanded="false" aria-controls="...">` per citation, labelled with `number`. The `cited` boolean drives the variant and icon. Add a visually-hidden suffix (`", cited"` / `", retrieved, not cited"`) so the distinction isn't carried by icon and color alone.

**Popover behavior.** Kevin's design says hover or click. WCAG 2.1 SC 1.4.13 requires dismissible, hoverable, persistent, and the card contains a link, so hover-only is out:

- click, Enter, or Space toggles (primary path)
- hover opens after a short delay and stays open while the pointer is inside the card
- Escape closes and returns focus to the chip
- click outside closes
- one popover open at a time

**This is not a reversal of your tooltip call.** Different thing: a bounded number of citations per answer, each one a useful destination, and only the active tab panel is ever in the tab order. Your reasoning about dozens of per-word spans still stands.

**View Site** reuses your existing `citationLink()` helper, which already allowlists http and https, sets `noopener noreferrer`, and appends the visually-hidden new-window cue. Move it onto the new service so both callers get it.

**Edit Site is deliberately not in scope**, see roadblocks.

**Chip labels:** the design says "Page 1". Citations can be media files, a PDF for instance, not just pages. Use "Source N".

### Phase 3: inline `[docN]` markers as citation badges

Kevin shows small circled numbers inline in the answer. We have the data and **right now it renders badly**, so this is a real fix rather than decoration.

`CitationFormatter` reads the `[docN]` markers to work out which sources were cited but never strips them, and both backends store the answer raw. Confirmed in `AiTesterBatch::processQuestion()` and `BeaconAnswerBackend::answer()`. So the compare and run views are currently printing literal `[doc1]` mid-sentence.

Fix: regex-replace `/\[doc(\d+)\]/` in the rendered answer HTML with a badge pointing at the matching chip. See roadblock 1 below for the catch.

**Ordering against the diff.** Do the replacement *after* `WordLevelDiff` runs, on the final HTML string. `WordLevelDiff` splits on whitespace and `[doc1]` contains none, so a marker always survives as one contiguous token, though it may land inside a `.diffchange` span. `Html::escape()` doesn't touch `[`, `]`, or digits, so the regex still matches. Please add a unit test covering a marker inside a changed span, since that assumption is what the approach rests on.

### Phase 4: diff highlighting toggle

Keep everything this PR built. Add a **Highlight differences** checkbox above the panels, default **off**, so the clean side-by-side loads first.

Cheapest correct version: keep rendering the diffed markup unconditionally and have the toggle add or remove a class on `.ys-compare` that neutralizes `.diffchange`. No re-render, no second copy of the answer in the payload, no round trip. Your custom properties already give you the single place to do it.

Show `diffHelp()` and the legend only when the toggle is on. They explain the highlighting, so they're noise when it's off.

Persist in `localStorage` so nobody re-toggles on every page load, and announce with `Drupal.announce()` the way `compare-export.js` does.

### Phase 5: copy and header reconciliation

- **Header buttons.** Keep what exists: Download JSON, Download CSV, Export for Clarity, Back to tester. Restyle to Kevin's button treatment. Don't collapse the three into one "Export Results" dropdown yet, since Export for Clarity is a modal and not a download.
- **Page copy.** Kevin's subhead says "Import a YAML test file to run the test." We don't use YAML. Write copy describing the actual `.txt` format.
- **Status vocabulary.** The design shows "Published". Real values are `processing`, `complete`, `failed`.
- **Tests.** Extend `tests/src/Unit/AiTesterCompareRenderTest.php`. New coverage: tab and panel structure with ARIA wiring, status badge on the tab, chip cited vs retrieved variants, popover contents, marker-to-chip mapping including the duplicate-URL case and the unresolvable marker, and the toggle defaulting to off.

---

## Two real roadblocks

Only two things here actually need your attention as problems.

### 1. `[docN]` marker numbering doesn't survive deduplication

`CitationFormatter::format()` sets `cited` from the citation's **pre-deduplication** index (`$index + 1`), but assigns `number` **after** deduplicating by URL. When two retrieved sources share a URL, every marker after that point is off by one. The stored citations array is the post-dedup output, so the original ordering is gone and the mapping can't be reconstructed at render time.

Suggested handling:

1. **Additive change to `CitationFormatter::format()`:** collect a `markers` array on each citation listing the original `[docN]` numbers that resolved to it. Purely additive, no schema change, nothing downstream breaks. New runs then carry an exact mapping.
2. **Fallback for existing runs:** where `markers` is absent, assume marker N maps to citation number N. Correct whenever no duplicate URLs were retrieved, which is the common case.
3. **Never drop a marker.** If one resolves to nothing, render an inert badge rather than deleting it or leaving raw `[docN]` on screen.

Open to a better approach if you see one.

### 2. "Edit Site" can't always be built, so we're not building it

Kevin's popover has a primary Edit Site button. `EntityCitationResolver` derives title and absolute URL from an entity, and `CitationFields` stores only those two on the indexed document. No entity id, no node id. Reversing a URL to a local node works for same-site content, but Beacon can query a shared collection spanning multiple sites, so a citation URL may point at content this install can't edit at all.

**Ship View Site only.** Edit Site gets its own ticket once we've worked out cross-site citations. I'd rather not render a button that works sometimes.

---

## Why the design says things we don't do anymore

Context only, none of this needs a decision and none of it should change what you build. It's here so nobody reads the Figma later and thinks we missed something.

**YAML import.** Several frames reference a YAML format with import rules. This is the clearest example of the design predating the build. We dropped YAML shortly after Kevin started, in favour of plain text one question per line, because it's far easier for site builders. `AiTesterForm::parseQuestionLines()` is the current reality. Ignore the YAML copy.

**"Select up to 3 runs."** The Compare Run Results card says three. We cut it to two to keep it simple, and the code is two-sided end to end: the route takes `{run_a}/{run_b}`, `RunComparator::compareResults()` is a two-way pairing algorithm, `AiTesterForm::compareSubmit()` rejects any other count. Two runs.

**The File 1 / File 2 dropdowns.** Skipping them. Run selection stays in the Run History checkboxes on the landing page. Not worth reworking a selection model that works.

**Named, editable Tests.** The landing table lists "Import test name 1" with Edit and Delete, implying a Test entity separate from a Run. There is no Test, only Runs, with questions in the `source_content` blob. `AiTesterRerunForm` covers Run Again via `source_run_id`. If we want editable Tests later that's a schema change and its own ticket. Not this work.

**No assistant awareness.** The design has no cross-assistant caveat and no assistant labels, because multi-backend support didn't exist when it was drawn. Both are required by #1470 and both stay.

---

## Suggested split

| Scope | Size |
|---|---|
| Phase 1: vertical question tabs and side-by-side panels | M |
| Phase 2: citation source chips with cited / retrieved popover | M |
| Phase 3: inline `[docN]` markers as citation badges, incl. the `CitationFormatter` fix | S |
| Phase 4: diff highlighting toggle | S |
| Phase 5: copy and header reconciliation | XS |
| Follow-up: Edit Site in the popover | ? |
| Follow-up: named, editable Tests | L |

Phase 1 is the one I care about landing. Happy for the rest to ship separately if that's cleaner than one big branch, your call on how to split the PRs.

Two last things. If any of this reads as more than the sizes above once you're into it, say so before starting rather than after. And this branch is still stacked on #1418, so that one goes first regardless.

miketullo95 (changes requested):
Close on this one — two UI tweaks before merge, both in the compare view styling.

**1. Question heading is too small** — `web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/modules/ys_ai_tester/css/question-tabs.css:143-146`

`.ys-qtabs__question` is `font-size: 1rem` — the same size as the tab preview text and barely bigger than the side headings below it (`0.9375rem`). It's the actual headline of each panel (e.g. "How much does YaleSites cost?") and should read like one. Bump it to `1.5rem`.

**2. Swap the Run A/Run B diff colors — drop red entirely** — `css/compare.css:20-34` (the `--ys-diff-removed-*`/`--ys-diff-added-*` custom properties, light + dark), `css/compare.css:106-134` (the legend swatches and `.ys-diff--removed`/`.ys-diff--added` rules that consume them)

Run A currently renders in the red family and Run B in green. Red reads as "this is wrong," which isn't the message here — Run A and Run B are just two valid answers being compared, not a right one and a wrong one.

- Run A → green
- Run B → blue

Same variables drive both the diff highlight and the legend swatch, so swapping the values covers both in one place. Whatever hex values you land on, keep measuring against WCAG 2.1 AA (4.5:1) in both light and dark mode the way the rest of this file already documents each pair's ratio — the `html.gin--dark-mode .ys-compare` overrides will need updating to match.

One thing worth a glance once it's in front of you: the "Identical" status badge (`compare.css:163-166`) is already the same green family (`#e6f4ea`/`#14663f`). With Run A also green there'll be two different elements using near-identical greens for different meanings. Probably fine since they don't appear side by side in a way that's ambiguous, but flagging it.

Everything else here looks solid — the contrast fixes, the export modal, and the question tabs all read well. No other notes.

@dblanken-yale

miketullo95 (changes requested):
One thing to fix before merge: this still names Clarity specifically as the export destination in a few spots, and since Clarity can't actually handle a comparison of this size, singling it out is misleading rather than helpful. Generalize the language to any AI/LLM instead. A few spots to update:

- `templates/ys-ai-tester-ai-export.html.twig:36-39` — the visible modal copy ("Yale's own Clarity is one option...") plus the "Open Clarity" link to ai-chat.yale.edu. Drop this paragraph and link entirely and keep the surrounding sentence generic ("any assistant approved for work use is a reasonable choice").
- `templates/ys-ai-tester-ai-export.html.twig:6` — the file-level doc comment still frames the design rationale around Clarity ("Clarity was the original target, but it cannot ingest..."). Update this once the visible copy no longer names it, so the comment doesn't reference UI text that's gone.
- `src/Controller/AiTesterController.php:573` and `:576` — same pattern in the `compareExport()` docblock ("Clarity was the original target... offers Clarity as one destination rather than the only one"). Generalize to "any LLM" / "any AI assistant."
- `tests/src/Unit/AiTesterCompareExportTest.php:184` — test docblock also calls out "not a Clarity-specific upload." Reword without naming Clarity for consistency.

Left `templates/ys-ai-tester-ai-export.html.twig:72` alone ("accuracy, completeness, clarity") since that's the plain English word describing answer quality, not a reference to the tool — no change needed there.

Minor/optional, not blocking: the PR title itself still says "Export to Clarity," might be worth a rename for anyone scanning PR history later.

@dblanken-yale

miketullo95 (approved):

PR #1441, opened 2026-08-10
1493: RC: MathJax multi-row LaTeX renders flattened — typogrify strips the \\ row separator before MathJax runs

dblanken-yale:
### Description of work

- `ys_mathjax` now supplies a `TypogrifyFilter` subclass that masks each MathJax-delimited region (`$$…$$`, `\[…\]`, `\(…\)` — exactly the delimiters in `mathjax.settings`) with an inert placeholder, lets typogrify do its real work on the surrounding prose, and restores the math verbatim. `hook_filter_info_alter()` swaps it in for the `typogrify` plugin.
- Applies to `basic_html`, `heading_html` and `restricted_html` at once, with **no text-format config change** — so there is no config export in this PR and nothing to re-import.
- Fixes more than the row separator. Confirmed damage on `develop`, all three formats: `\\` → `&#92;`, `\,` → `&#x2c;`, and `f''(x)` → `f&rdquo;(x)` (the `''` quote conversion enabled in our formats destroys double-prime notation — not catalogued in the issue).
- Removes the README note that told editors to work around this ("prefer `\prime` and `\ldots` rather than `'` and `...`"), replacing it with the multi-row example that now works.

#### Two acceptance criteria in the issue rest on a wrong premise — please sanity-check my reasoning

1. **"Run `filter_ys_mathjax` before `typogrify`" cannot work.** `mathjax`'s filter does no LaTeX processing server-side — it wraps the text in a div and attaches the JS library (`web/modules/contrib/mathjax/src/Plugin/Filter/MathjaxFilter.php`). TeX→MathML happens in the browser, after every text filter. So the relative order of the two filters is irrelevant; typogrify rewrites the raw LaTeX either way. That route is rejected, and no filter weights are changed.
2. **`filter_ys_mathjax` is only enabled in `basic_html`.** `heading_html` and `restricted_html` have `typogrify` but no MathJax filter. Math in those formats still renders client-side *if* MathJax was loaded by a `basic_html` field elsewhere on the page, so fixing typogrify in all three is still right — but enabling the MathJax filter in the other two formats is #982 scope and is **not** done here.

#### Why this shape

- **Not a contrib patch to `typogrify`**: it would hard-code our math delimiters into a contrib module and need re-rolling on every update. The subclass lives in our module and is unit tested.
- **Not repairing the entities afterwards** in `YsMathjaxFilter` (weight 100): that only reverses `processEscapes()`, leaving the quote, dash and ellipsis conversions that also corrupt math.
- **Not a second "protect" filter at a low weight**: that puts placeholder state across filter boundaries and needs config changes in three formats. Here the placeholder never leaves the filter's own `process()` call.
- `smartypants_enabled` is untouched — typogrify keeps doing its real work.

### Verification performed

Local Lando, all three formats, via `check_markup()` on 11 cases each:

| | `develop` | this branch |
|---|---|---|
| cases preserved | 18 / 33 | **33 / 33** |
| `bmatrix` / `pmatrix` / `array` / `aligned` | `\\` → `&#92;` | intact |
| `\,` thin space | → `&#x2c;` | intact |
| `f''(x)` | → `f&rdquo;(x)` | intact |
| prose (`"yes"`, `--`, `...`, `$5`) | typogrified | **identical, unchanged** |

Rendered page, real Text block, MathJax MathML (the accessibility criterion):

| construct | `develop` | this branch |
|---|---|---|
| bmatrix | 1 `<mtr>` / 3 `<mtd>` | **2 `<mtr>` / 4 `<mtd>`** |
| pmatrix | 1 `<mtr>` / 3 `<mtd>` | **2 `<mtr>` / 4 `<mtd>`** |
| array | 1 `<mtr>` / 3 `<mtd>` | **2 `<mtr>` / 4 `<mtd>`** |
| aligned | 1 `<mtr>` / 3 `<mtd>` | **2 `<mtr>` / 4 `<mtd>`** |

`develop` reproduced the issue's claim exactly (one row of three cells), so assistive tech was announcing a mathematically different matrix. No `merror` in any expression; single-line inline and display equations unchanged.

- 33 Unit tests pass (was 9). Each math case also asserts the *stock* filter corrupts that fragment, so a case cannot pass for the wrong reason.
- `composer code-sniff` (Drupal + DrupalPractice) and `composer lint:php` exit 0.
- `npm run lint:styles` fails with 1536 pre-existing errors, all in `web/themes/contrib/atomic/_yale-packages/tokens/build/scss/tokens.scss` — a generated file in the local gyst'd atomic clone that is absent in CI. This diff contains no style files.

### Reviewer notes

- **A code review caught a real bug in my first attempt.** The placeholder was `ysmathmask0ysmathmask`; `SmartyPants::smartNumbers()` matches a bare `\d+` with **no word boundary**, so with the format's "Digit grouping in numbers" option enabled (off today, but a checkbox on the settings form) it wrapped the `0` in `<span class="number">`, the restore missed, and the math was **deleted** and replaced by visible gibberish — worse than the bug being fixed. Verified live: `ysmathmask<span class="number">0</span>ysmathmask`. The index is now spelled with letters, and `testMathSurvivesEveryTypogrifyOption` turns on each optional typogrify transformation in turn as a regression guard.
- **Not spot-checked per-field.** The issue asks for two rich-text surfaces beyond the Text block. I verified at the `check_markup()` layer for all three formats instead, which is the code path every rich-text field funnels through — exhaustive over formats rather than a sample of fields. Say the word if you want the per-field checks anyway.
- **Security**: masking restores the exact bytes typogrify was handed, and `filter_html` runs at weight `-10` in all three formats, so it cannot reintroduce anything sanitisation removed. Typogrify's incidental `<`→`&lt;` is redundant here because `Xss::split()` already returns `&lt;` for a stray `<`. No config change means no `full_html`-style format is affected.
- **Uncovered**: the `preg_replace_callback` returning `NULL` fallback (a PCRE limit) has no test — it silently reverts to today's broken behaviour. I left the logger out rather than add dependency injection to this filter for one line; happy to add it if you'd rather have it diagnosable.

### Functional testing steps:

- [ ] On a page, add a **Text** block containing `$$\begin{bmatrix} a & b \\ c & d \end{bmatrix}$$` and save.
- [ ] View the published page — the matrix renders as **two rows**, not one flat row.
- [ ] Repeat with `pmatrix`, `array`, and an `aligned` system of equations (`\[\begin{aligned} x + y &= 2 \\ x - y &= 0 \end{aligned}\]`) — all stack correctly.
- [ ] Check `$$\int_0^1 x\,dx$$` — the `\,` renders as a thin space, not a comma.
- [ ] Regression-check single-line math still renders: `\(E = mc^2\)` inline and `$$a^2 + b^2 = c^2$$` display.
- [ ] Confirm typography elsewhere on the page is unaffected — straight quotes still become curly, `--` still becomes a dash, `...` still becomes an ellipsis.
- [ ] Confirm ordinary prose is untouched: "Tickets are $5 each" shows no math treatment.

References yalesites-org/YaleSites-Internal#1493

---

Created with AI

Screenshot 2026-08-10 at 3 29 05 PM

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_mathjax/README.md
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_mathjax/README.md
@@ -29,10 +29,13 @@ page loads.
 
 - A single dollar sign (`$`) does **not** start math, so ordinary text such as
   "Tickets are $5" is unaffected.
-- Because the Text block also applies typographic replacement (Typogrify),
-  prefer LaTeX macros over literal punctuation inside math — e.g. use `\prime`
-  and `\ldots` rather than `'` and `...` so the source is not altered before
-  MathJax renders it.
+- Punctuation inside the delimiters is left exactly as typed. Multi-row
+  constructs work as they do in standard LaTeX — `\\` ends a row and `&`
+  separates columns:
+
+  ```
+  $$\begin{bmatrix} a & b \\ c & d \end{bmatrix}$$
+  ```
 
 ## For developers
 
@@ -40,5 +43,11 @@ page loads.
   decide whether to load the library (unit tested).
 - `Plugin\Filter\YsMathjaxFilter` — extends the contrib `MathjaxFilter`;
   attaches the library via the parent only when `hasMath()` is TRUE.
+- `Plugin\Filter\YsTypogrifyFilter` — extends the contrib `TypogrifyFilter` and
+  masks math regions so typographic replacement cannot rewrite LaTeX before
+  MathJax sees it. `ys_mathjax_filter_info_alter()` swaps it in for the
+  `typogrify` plugin, so the protection applies to every text format that runs
+  typogrify (`basic_html`, `heading_html`, `restricted_html`), not just the ones
+  with the MathJax filter enabled.
 - Delimiters and MathJax options are configured in `mathjax.settings`
   (`config_type: 0`, single-dollar inline math disabled).
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_mathjax/src/MathDelimiterDetector.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_mathjax/src/MathDelimiterDetector.php
@@ -24,6 +24,19 @@ class MathDelimiterDetector {
   const MATH_PATTERN = '/\\\\[([]|\$\$|<math[\s>\/]/i';
 
   /**
+   * Matches a complete math region, delimiters included.
+   *
+   * The same delimiter pairs as MATH_PATTERN, but matched closed rather than
+   * open, so a caller can pick math out of surrounding prose. Keep the two in
+   * step: both describe the delimiters configured in `mathjax.settings`.
+   *
+   * MathML is deliberately absent. `<math>` is already in SmartyPants'
+   * skip-tag list, and Basic HTML's filter_html strips the element long before
+   * any of this runs.
+   */
+  const MATH_REGION_PATTERN = '/\$\$.*?\$\$|\\\\\[.*?\\\\\]|\\\\\(.*?\\\\\)/s';
+
+  /**
    * Checks whether the given text contains math notation.
    *
    * @param string $text
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_mathjax/src/Plugin/Filter/YsTypogrifyFilter.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_mathjax/src/Plugin/Filter/YsTypogrifyFilter.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace Drupal\ys_mathjax\Plugin\Filter;
+
+use Drupal\typogrify\Plugin\Filter\TypogrifyFilter;
+use Drupal\ys_mathjax\MathDelimiterDetector;
+
+/**
+ * Applies typographic refinements to everything except math notation.
+ *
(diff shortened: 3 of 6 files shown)


miketullo95 (approved):
@dblanken-yale This is a solid fix — masking math regions before typogrify runs, rather than reordering filters or patching contrib, is the right call given MathJax renders client-side and never sees filter order. The regression coverage is thorough (33/33 unit tests, each asserting the stock filter would actually corrupt the fragment) and the digit-grouping bug you caught in your own first attempt is a good example of why that guard matters.

I tested on the multidev: bmatrix and the aligned system both stack into the correct rows, the thin space in the integral renders correctly (not a comma), and inline/display single-line equations are unaffected. Looks good.

On the two premises you flagged:
- Agreed the filter-reorder route in the issue doesn't apply — MathJax does no server-side LaTeX processing, so filter weight is irrelevant.
- Fine deferring the MathJax-filter-in-heading_html/restricted_html scope to #982 — this PR's job is making sure typogrify doesn't corrupt math wherever it runs, not guaranteeing MathJax renders in every format on its own.

Also fine with check_markup()-layer verification standing in for manual per-field spot checks, and leaving the PCRE-limit fallback unlogged for now — neither is a blocker.

Posted a comment on yalesites-org/YaleSites-Internal#982 flagging that its "Known limitations" section and Release Testing Step 4 still reference the old `\prime`/`\ldots` workaround this PR makes unnecessary — worth cleaning up before the still-pending editor-facing doc gets written.

Approving.

PR #1442, opened 2026-08-10
1494: RC: Retroactively fix pre-#683 document links that still render double-encoded (%2520)

dblanken-yale:
### Description of work

- Adds `ys_core_update_10019()`, a one-time update hook that repairs link URIs stored with their percent-encoding still intact from before yalesites-org/YaleSites-Internal#683. Drupal encodes the path a second time when it builds the href (`UnroutedUrlAssembler::buildLocalUrl()`), so a stored `%20` rendered as `%2520` and the document 404'd. The #683 fix only decodes at save time, so nothing self-healed.
- Writes the field tables directly instead of loading and saving entities, so **no new revisions are created and no `changed` timestamps move** into the Authoring Dashboard's "Latest Edits". Both the field data table and the field revision table are rewritten, because Layout Builder renders an inline block from a specific revision id (`InlineBlock::getEntity()` calls `loadRevision()`).
- Repairs only the **path** of an `internal:`/`base:` URI. `Url::fromUri()` splits the query and fragment off before that encoding happens, and external URIs are returned untouched, so neither can double-encode. A value is rewritten only when its stored path is exactly the canonical encoding of its decoded form **and** decoding leaves no percent behind — which keeps a file named `100% off.pdf` intact and makes the hook strictly idempotent.
- Discovers link fields from the field map rather than a hardcoded list, so a site's own link fields are covered too. Link fields kept in shared table storage (the menu link base field, redirects) are **logged as skipped** rather than silently implied to be covered.
- Invalidates the affected entities' cache tags and resets the entity storage cache, so corrected links appear without a manual cache rebuild.
- Adds Unit coverage for the match rule (22 cases, including the values that must not be touched) and Kernel coverage proving both tables are repaired, no revisions are created, `changed` does not move, the run is idempotent, and a dry run writes nothing.

### Verification performed locally

Run against a real pulled database, on the visual-regression content that reproduces the issue:

- `drush updatedb` reported: `Repaired 7 double-encoded link URI(s) across 32 field table(s); left 0 value(s) untouched.`
- `/blocks-for-visreg/button` went from **3** `%2520` hrefs to **0**, with no manual cache rebuild.
- The corrected URL returns **HTTP 200** with the spreadsheet content type; the double-encoded URL returned **404**.
- `block_content_revision` row count unchanged (450 to 450); all 8 captured `changed` timestamps byte-identical.
- Second run repaired 0 rows; a rescan found 0 remaining candidates.
- **No queue items were created**, confirming Beacon/Search API indexing is not triggered. Direct table writes invoke no entity hooks at all, so this is structural rather than incidental.
- ys_core Kernel suite on MySQL: 17 tests, 0 failures. ys_core Unit suite: 153 tests vs a 131-test baseline (the 22 added), 0 failures.
- `composer code-sniff` exits 0.

### Functional testing steps:

- [ ] Visit a page with a Button Link (or any link field) pointing at a document whose file name contains spaces, saved before the #683 fix
- [ ] Confirm the link's href contains `%20` rather than `%2520`
- [ ] Click the link and confirm the document opens rather than 404ing
- [ ] Confirm the content's "Changed" date has not moved and no new revision was created
- [ ] Confirm links that were already correct, and external links, are unchanged

Added for the `+` rework (`ys_core_update_10020()` and the patch change):

- [ ] Upload a document whose file name contains a `+` and no space, e.g. `annual+report.pdf`
- [ ] Paste that file's path into a Button Link with a literal `+` (not `%2B`) and save
- [ ] Confirm the stored link keeps the `+` (it used to be saved with a space), the rendered href is `annual%2Breport.pdf`, and clicking it opens the document
- [ ] For the retroactive half: with such a document in place, set a link's stored URI to the space form (`annual report.pdf`) and run `drush updatedb`; confirm the `+` is restored and the link resolves
- [ ] Confirm a link to a document whose name genuinely contains a space is **not** rewritten
- [ ] Confirm a link whose document no longer exists is left alone, and appears in the dblog warning naming what could not be decided

### Notes for the reviewer

- **A correction to the ticket's example.** The ticket motivates the revision-table handling with block 126 / revision 217. Querying the layout API shows revision 217 is rendered by no node — it is an orphaned historical revision, and block 126's current revision was already correct. The genuinely load-bearing cases are revisions 218, 220 and 225, which are exactly the three broken hrefs that were rendering. Writing the revision table is still required, but for those rather than for 217.
- **The `+` problem is now fixed here too**, per review feedback. The patch uses `rawurldecode()` so `+` is no longer eaten at save time, and `ys_core_update_10020()` repairs values already stored with a space, deciding from the filesystem which of the two file names exists. The pre-#683 form (`a%2Bb.pdf`) needed no new logic — the 10019 rule already repairs it — and now has test coverage.
- **A separate pre-existing bug, NOT fixed here and not a regression from this PR:** the patch decodes the *whole* URI rather than just the path, and `LinkitHelper::uriFromUserInput()` returns `mailto:` and external input verbatim. So a stored `?token=a%26b` becomes `?token=a&b` (splitting the parameter) and `?q=one%23two` becomes a fragment. `urldecode()` did the same, so this is already live on `master`. It wants its own ticket; see the PR comment for the suggested fix.
- Link fields in shared table storage are not repaired (0 affected rows were found on the test database); the hook logs exactly which tables it skipped so this is visible rather than assumed.

References yalesites-org/YaleSites-Internal#1494

---
Created with AI

### DIFF
--- a/patches/contrib/linkit/3436733-5.patch
+++ b/patches/contrib/linkit/3436733-5.patch
@@ -2,13 +2,16 @@ diff --git a/src/Plugin/Field/FieldWidget/LinkitWidget.php b/src/Plugin/Field/Fi
 index c9e4620..0585d09 100644
 --- a/src/Plugin/Field/FieldWidget/LinkitWidget.php
 +++ b/src/Plugin/Field/FieldWidget/LinkitWidget.php
-@@ -216,7 +216,8 @@ class LinkitWidget extends LinkWidget {
+@@ -216,7 +216,11 @@ class LinkitWidget extends LinkWidget {
     */
    public function massageFormValues(array $values, array $form, FormStateInterface $form_state) {
      foreach ($values as &$value) {
 -      $value['uri'] = LinkitHelper::uriFromUserInput($value['uri']);
 +      $uri = LinkitHelper::uriFromUserInput($value['uri']);
-+      $value['uri'] = !empty($uri) ? urldecode($uri) : NULL;
++      // rawurldecode() rather than urldecode(): a URL path is not form data, so
++      // a "+" in it is a literal plus and not an encoded space. urldecode()
++      // turned a link to "a+b.pdf" into one to "a b.pdf", which then 404s.
++      $value['uri'] = !empty($uri) ? rawurldecode($uri) : NULL;
        $value += ['options' => $value['attributes']];
      }
      return $values;
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/tests/src/Kernel/DoubleEncodedLinkUriRepairTest.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/tests/src/Kernel/DoubleEncodedLinkUriRepairTest.php
@@ -1,0 +1,219 @@
+<?php
+
+namespace Drupal\Tests\ys_core\Kernel;
+
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\block_content\Entity\BlockContent;
+use Drupal\block_content\Entity\BlockContentType;
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+
+/**
+ * Tests the retroactive repair of double-encoded link URIs in field tables.
+ *
+ * Issue #1494: links stored before #683 kept their percent-encoding and render
+ * double-encoded. The repair writes the field tables directly, so this covers
+ * what an entity save would have got wrong: the revision row Layout Builder
+ * renders from, the absence of new revisions, and the "changed" timestamp
+ * staying put.
+ *
+ * @group ys_core
+ * @group yalesites
+ */
+class DoubleEncodedLinkUriRepairTest extends KernelTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'system',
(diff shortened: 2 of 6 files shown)


miketullo95 (approved):
@dblanken-yale This is a thorough, well-documented data migration. Writing directly to the field tables instead of loading/saving entities is the right call, and both the revision-row handling for inline Layout Builder blocks and the match rule's precision (round-tripping the decode through rawurlencode() to confirm it's genuinely double-encoded, not just percent-containing) hold up well against the acceptance criteria.

On the three open items:
- The `+` scope extension — glad this got folded in given how directly related it is (same forward-fix, same repair mechanism), and the verification on it is just as thorough as the main fix.
- The whole-URI decode bug — agreed this should be its own ticket rather than riding along here. Opened yalesites-org/YaleSites-Internal#1513 with your suggested fix, assigned to you, Medium/M.
- Production-copy verification — the structural argument (Beacon/Search API never reached for block_content, not just observed empirically) plus the dry-run/idempotency/cache-invalidation coverage is enough for me; not asking for an explicit prod-copy run before merge.

Documentation AC (release note on the auto-heal, editor guidance for un-repairable links) — the release note half will get picked up in the normal release-notes pass rather than a separate ticket; not blocking this PR either way.

Approving.

PR #1443, opened 2026-08-10
1508: RC: Sitewide alert block shows a contextual edit button on the front end

dblanken-yale:
### Description of work
- Adds `hook_block_view_BASE_ID_alter()` in `ys_alert` for the `alert_block` plugin, dropping the `block` contextual-links group so Drupal's contextual edit pencil no longer renders on the sitewide alert on the front end.
- Core attaches a "Configure block" link to *every* placed block. On the alert that pencil opened the block instance form — block label and visibility conditions only — not the alert content, which is managed in Alert Settings at `/admin/yalesites/alert`.
- This removes the control rather than changing any permission: `entity.block.edit_form` stays gated by "administer blocks", and platform admins can still reach the block instance form from `/admin/structure/block`.
- Because core only renders the contextual placeholder for a non-empty `#contextual_links`, emptying the group suppresses the pencil for **every** role rather than relying on a per-role permission check.
- Keying the hook to the single `alert_block` plugin (rather than a global `hook_block_view_alter()` plus a condition) means other placed blocks are structurally unaffected.
- **Answers the open grooming question on the issue:** of the roles on the platform, only `platform_admin` holds "administer blocks", so it was the only role that actually saw the pencil. `site_admin`, `editor`, and `contributor` have "access contextual links" but not "administer blocks", so they received the placeholder attribute but no rendered pencil. After this change the placeholder is not emitted for any of them.
- Covered by a Unit test for the hook body plus a Kernel test that runs real hook discovery and reads the plugin id out of the shipped `block.block.alert_block.yml`, so renaming the `alert_block` plugin id cannot silently break the hook-name coupling and quietly restore the pencil.

### Functional testing steps:
- [ ] As a platform admin, confirm a sitewide alert is active (set one at `/admin/yalesites/alert`).
- [ ] Visit any front-end page where the alert renders and hover over the alert — **no** round edit pencil appears at the upper right.
- [ ] Hover a different placed block (e.g. the footer block) — its contextual pencil **still** appears, with "Configure block" / "Remove block".
- [ ] Confirm Alert Settings at `/admin/yalesites/alert` still loads with the current alert values and still saves.
- [ ] Log in as a site admin (or any non-platform-admin role) and confirm the alert shows no pencil there either.
- [ ] The alert itself is unchanged: headline, message, link, and the dismiss ("X") button all still render and work.
- [ ] If a pencil still shows immediately after deploy, rebuild caches — the block render cache is keyed `entity_view:block:alert_block`, so a pre-fix cached entry can survive until a rebuild. `drush deploy` covers this.

References yalesites-org/YaleSites-Internal#1508

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_alert/tests/src/Kernel/AlertBlockContextualLinksTest.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_alert/tests/src/Kernel/AlertBlockContextualLinksTest.php
@@ -1,0 +1,120 @@
+<?php
+
+namespace Drupal\Tests\ys_alert\Kernel;
+
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\block\BlockViewBuilder;
+use Drupal\block\Entity\Block;
+use Symfony\Component\Yaml\Yaml;
+
+/**
+ * Tests that a placed alert block renders with no contextual links.
+ *
+ * The Unit test covers what the alter hook does to a render array. This one
+ * covers the part that only real hook discovery can: that the hook is *named*
+ * so core actually calls it for this plugin. The name embeds the plugin's base
+ * id (`block_view_<base_id>_alter`), so renaming the `alert_block` plugin id
+ * would silently stop the hook firing and bring the pencil back. Reading the
+ * plugin id out of the shipped block placement instead of hardcoding it here is
+ * what makes that regression fail this test.
+ *
+ * @see \Drupal\block\BlockViewBuilder::buildPreRenderableBlock()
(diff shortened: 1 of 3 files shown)


miketullo95 (approved):
@dblanken-yale Clean, tight fix — keying the hook to the alert_block plugin id rather than a global alter, and the Kernel test that reads the plugin id from the shipped config so a rename can't silently bring the pencil back, are both good calls. Covers all four of #1508's acceptance criteria.

Tested on the multidev, looks good.

Separate note, unrelated to this PR: noticed the Emergency alert's expand/collapse chevron doesn't show a pointer cursor on hover. Confirmed it's untouched by this diff — opened yalesites-org/YaleSites-Internal#1514 for it (Low/XS), not blocking here.

Approving.

PR #1446, opened 2026-08-11
1497: RC: Quick Links: several malformed URLs save silently as broken or dead links

dblanken-yale:
### Description of work

All three reported cases were reproduced on local Lando before anything was changed, and each was fixed.

- **`//example.com` (protocol-relative) saved as a dead link.** Confirmed mechanism: the core link widget rewrites it to the unresolvable `internal://example.com`, which *passes* validation, saves, and renders as `<a href="">`. Note the issue says "no `href` attribute is emitted" — measured, it emits an **empty** `href=""`. Dead either way.
- **`https:/example.com` (one slash) saved as a broken link.** It already declares a scheme, so it was stored verbatim and rendered as a literal `<a href="https:/example.com">`.
- Both are now **auto-corrected** to a valid absolute URL. **This is the AC's "pick one behavior" call:** auto-correct, not reject, applied identically to both shapes — consistent with the bare-domain case the same function already corrects silently (from yalesites-org/YaleSites-Internal#1158). Rejecting one shape while silently fixing its sibling would be the inconsistent option.
- **The AC's third development bullet asked whether the fix belongs in a constraint scoped to Quick Links or in a broader Linkit/core fix. It is neither.** There is already a platform-wide normaliser (`_ys_core_normalize_bare_domain_uri()`, hooked onto *every* core LinkWidget by `ys_core_field_widget_single_element_form_alter()`), so the correct place was to extend that one function: every link field on the platform gets the fix, no new constraint, plugin, or file. Quick Links needing no custom plugin is exactly why this works.
- The scheme is **lowercased** during correction. Verified empirically that this is required, not cosmetic: core compares `parse_url()`'s scheme against its allowed-protocol list with a case-sensitive `in_array()`, so `HTTPS://example.com` is rejected as invalid — without lowering, the "correction" would produce a value that still fails.
- **`htp://example.com` was rejected with no visible message.** The root cause is more specific than the ticket assumed, and worth flagging: **the submit is not an Ajax request at all.** Measured in the browser — `Drupal.ajax.instances` has 55 entries, none bound to the Update button; fetching the route with `_wrapper_format=drupal_dialog.off_canvas`, `drupal_ajax`, and none all return the submit as a plain `<input type="submit">` with no ajax settings entry and no `ajaxSubmit` anywhere. So core's own error-path message injection (`AjaxFormHelperTrait::ajaxSubmit()`) never runs. The editor is thrown out of the off-canvas dialog onto the standalone Configure block page, and Gin's `page--layout-builder.html.twig` prints only `pre_content` and `content` — never the `highlighted` region where `block.block.ys_admin_theme_messages` lives. So the message stayed queued and appeared later on an unrelated page.
- Fixed by adding a `status_messages` element to both Layout Builder configure-block forms, which renders the message wherever the form renders. It deliberately reuses the key core's Ajax error path assigns, so if that path ever does run it **overwrites** this element instead of rendering and announcing the error twice.
- Scoped to all Layout Builder block forms rather than just Quick Links: any block's configure form can reject any field, and the missing-message defect is not Quick Links-specific.
- Adds unit cases for the normaliser (it previously had **no** test coverage at all) and 3 kernel tests for the form alter.

### Update — a mistyped scheme is now auto-corrected too (per review)

Per @dblanken-yale's review comment, `htp://example.com` is now **auto-corrected** rather than merely rejected-with-a-visible-message. It is handled in the same platform-wide normaliser as the other three shapes, so it applies to every link field on the platform.

- The correction is bounded to a scheme within **one edit** (`levenshtein()`) of `http`/`https`, and can only ever emit `http` or `https`.
- **That bound is the safety property, not a heuristic.** Measured against this site's real `UrlHelper::getAllowedProtocols()` — `http, https, ftp, news, nntp, tel, telnet, mailto, irc, ssh, sftp, webcal, rtsp` — every one of them is **at least 2 edits** from both (`ftp` and `rtsp` are the closest, at 2). `javascript` is 9 and `data` is 3. So no scheme an editor deliberately typed can be rewritten, and no executable scheme can be manufactured. A dedicated test pins this, so widening the bound to 2 fails the suite rather than silently rewriting `ftp://`.
- Ties resolve to `https`, so a secure address is never downgraded.
- Two edits is treated as a different word, not a typo: `hxxp://example.com` (a defanged URL — the realistic case) is still rejected, now with the visible message this PR already added.

**This does NOT make PR #1444 redundant — recommend keeping it open.** #1444 patches gin_lb's `preprocessFieldMultipleValueForm`, which runs for **every multi-value field on the platform**; its own regression steps include an Accordion block submitted with empty required fields, which has nothing to do with URLs. This change removes exactly one input that can trigger a validation-error re-render on one field of one block type — every other trigger survives, including `hxxp://example.com` above. That matches @miketullo95's read in the approval above that the two are complementary.

**Known limits, stated plainly:**
- A bare domain carrying a **port** (`example.com:8080/a`) is still not normalised — the pre-existing "already declares a scheme" guard matches `example.com:`. That is a real pre-existing gap rather than intended behaviour; it is now pinned by a test asserting current behaviour so the change above provably does not make it worse. Out of scope here, worth a follow-up.
- **Links already saved in the database are not fixed** — only new saves are corrected. No backfill update hook is included. This matches #1158's own precedent (it shipped no backfill either), but existing broken links stay broken until an editor re-saves them. Worth a follow-up ticket.
- The editor being ejected from the off-canvas dialog is itself broken behaviour and is the *reason* the message was lost. Restoring the missing `#ajax` is core/gin_lb territory and would change the configure-block submit flow for every block type — **deliberately not attempted here.** The alternative deeper fix (overriding `page--layout-builder.html.twig` in `ys_admin_theme` to print `{{ page.highlighted }}`) would also close it generically, and would additionally cover `layout_builder.remove_block` / `remove_section` / `choose_*`, which have the same message-dropping defect and are **not** covered by this PR.
- Pre-existing and untouched: a well-formed but upper-cased `HTTPS://example.com` (double slash) is still rejected by core for the same case-sensitivity reason. Not in this ticket's scope, but it is a real defect, not a quirk.
- **Accessibility:** the rejection now meets WCAG 3.3.1 with visible text (not colour alone) — the box renders an `<h2>Error message</h2>` plus the error text, with `role="contentinfo"`/`aria-labelledby`, measured contrast 5.16:1 (AA for normal text), and `aria-invalid="true"` on the field. **Gap:** the field's `aria-describedby` points at its description, not the error text, so this is a form-level message rather than a per-field programmatic association. Closing that means enabling core's Inline Form Errors module platform-wide — out of scope here.

### Verification already done on local Lando

- Cases 1 and 2, end to end through the real contextual **Configure** dialog on node 36's Quick Links block: `//example.com` and `https:/example.com` both now store `https://example.com` and the published page renders `<a href="https://example.com">`. Controls `/about` (→ `internal:/about`) and `https://yale.edu` save unchanged, and the block's other 8 links render unaffected. Test fixture restored to its original value afterwards.
- Case 3: the rejection message is now visible on screen at the moment of failure, and appears **exactly once** (measured in the DOM, not assumed).
- Security: the message now renders editor-supplied input where it previously rendered nothing, so this was checked at runtime rather than reasoned about — entering `htp://x"><script>alert(1)</script>` renders as literal escaped text with no script execution (escaped at `FormattableMarkup::placeholderFormat()`'s `@` → `Html::escape()`). No protocol-filter bypass: both new branches can only emit `http`/`https`, `javascript:`/`data:` fall through untouched, and `https:/javascript:…` becomes *more* strictly rejected than before.
- `composer code-sniff` (the real CI gate): **exit 0**. `composer lint:php`: exit 0.
- ys_core **Unit**: `Tests: 158, Assertions: 211, Skipped: 1` — 0 failures (baseline before the change: 131 tests, 0 failures).
- ys_core **Kernel** (full): `OK (20 tests, 327 assertions)`, including the pre-existing `GrandHeroReusableBlockFormAlter` regression test for this same `ys_core_form_alter()` branch.
- The tests were written first and confirmed failing for the right reason; the kernel test was additionally mutation-checked (renaming the form key back makes 2 of 3 fail), so it cannot pass vacuously.
- Note this project's CI runs no PHPUnit, so these suites are local-only and do not gate the PR.

### Functional testing steps:

- [ ] Edit a page, open **Layout Builder**, and **Configure** an existing Quick Links block.
- [ ] Set one link's URL to `//example.com` and save the block, then **Save layout**. Confirm the link on the published page goes to `https://example.com` — not a link-styled element that does nothing when clicked.
- [ ] Repeat with `https:/example.com` (one slash). Confirm it also resolves to `https://example.com`.
- [ ] Repeat with `HTTPS:/example.com`. Confirm it saves and resolves to `https://example.com`.
- [ ] Repeat with `htp://example.com` (mistyped scheme). Confirm it now **saves** and the published page resolves it to `http://example.com` — it is no longer rejected.
- [ ] Configure the block again, set a link's URL to `hxxp://example.com` (a defanged URL: two characters wrong, so deliberately not guessed at), and click **Update**. Confirm a visible red error box reading `The path 'hxxp://example.com' is invalid.` appears immediately, and that it appears only once. Before this fix the field was outlined in red with no message, and the text turned up later on an unrelated page.
- [ ] Regression on the scheme correction: confirm `ftp://example.com`, `mailto:someone@yale.edu` and `tel:+12035551212` all save **exactly as typed** and are not rewritten to `http`.
- [ ] Regression: confirm normal links still work — an internal link picked from autocomplete, `/about`, `https://yale.edu`, `mailto:someone@yale.edu`, and a `tel:` link all save and render as before.
- [ ] Regression: confirm a validation error on a **different** block type's configure form (e.g. leave a required Grand Hero field empty) also now shows its message in the form.

References yalesites-org/YaleSites-Internal#1497

---

Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/tests/src/Kernel/LayoutBuilderBlockFormMessagesTest.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/tests/src/Kernel/LayoutBuilderBlockFormMessagesTest.php
@@ -1,0 +1,126 @@
+<?php
+
+namespace Drupal\Tests\ys_core\Kernel;
+
+use Drupal\Core\Form\FormState;
+use Drupal\KernelTests\KernelTestBase;
+
+/**
+ * Tests that Layout Builder block forms render their own validation messages.
+ *
+ * Regression test for a rejected value leaving the editor with no explanation.
+ * The Layout Builder configure-block submit button carries no #ajax, so a
+ * server-side validation failure re-renders the form as a standalone Layout
+ * Builder page, and Gin's page--layout-builder.html.twig never prints the
+ * highlighted region that holds the site's messages block. The error text ("The
+ * path 'htp://example.com' is invalid.") therefore stayed in the messenger
+ * queue and surfaced later on an unrelated page load, leaving only a
+ * red-outlined field with no visible text — which also gives screen reader
+ * users no error at all.
+ *
+ * Placing a status_messages element in the form itself makes the message render
+ * wherever the form is rendered, in the dialog or as a standalone page.
+ *
+ * @group ys_core
+ */
+class LayoutBuilderBlockFormMessagesTest extends KernelTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'system',
+    'user',
+    'ys_core',
+  ];
+
+  /**
+   * Builds a Layout Builder block form array and its form state.
+   *
+   * The admin label default must be non-empty: ys_core_form_alter() only
+   * reaches into the form object for the current component when it is missing,
+   * which a bare form state cannot supply.
+   */
+  protected function blockForm(string $plugin_id): array {
+    $form = [
+      'settings' => [
+        'label' => ['#default_value' => 'Quick Links'],
+        'block_form' => [],
+      ],
+    ];
+    $form_state = new FormState();
+    $form_state->setBuildInfo(['args' => [NULL, NULL, NULL, $plugin_id]]);
+    return [$form, $form_state];
+  }
+
+  /**
+   * Both configure-block forms carry a status_messages element.
+   */
+  public function testConfigureBlockFormCarriesMessagesElement(): void {
+    foreach (['layout_builder_add_block', 'layout_builder_update_block'] as $form_id) {
+      [$form, $form_state] = $this->blockForm('inline_block:quick_links');
+
+      $this->assertArrayNotHasKey(
+        'status_messages',
+        $form,
+        "The messages element must come from the alter, not the fixture ($form_id)."
+      );
+
+      ys_core_form_alter($form, $form_state, $form_id);
+
+      $this->assertArrayHasKey(
+        'status_messages',
+        $form,
+        "A rejected value must be explained inside the block form ($form_id)."
+      );
+      $this->assertSame('status_messages', $form['status_messages']['#type']);
+      // It has to outrank the form's own fields so the editor sees it without
+      // scrolling the off-canvas panel.
+      $this->assertLessThan(0, $form['status_messages']['#weight']);
+    }
+  }
+
+  /**
+   * Core's own Ajax error element replaces ours instead of joining it.
+   *
+   * The element has to reuse the key core assigns in
+   * AjaxFormHelperTrait::ajaxSubmit(). Two elements would otherwise reduce to
+   * the same lazy-builder placeholder, and the renderer substitutes every
+   * occurrence of it — so a second element means the editor sees the error
+   * twice and a screen reader announces it twice.
+   */
+  public function testCoreAjaxMessagesElementDoesNotDuplicateOurs(): void {
+    [$form, $form_state] = $this->blockForm('inline_block:quick_links');
+    ys_core_form_alter($form, $form_state, 'layout_builder_update_block');
+
+    // Exactly what core does when an Ajax submit fails validation.
+    $form['status_messages'] = [
+      '#type' => 'status_messages',
+      '#weight' => -1000,
+    ];
+
+    $messages = array_filter(
+      $form,
+      fn ($element) => is_array($element)
+        && ($element['#type'] ?? NULL) === 'status_messages'
+    );
+
+    $this->assertCount(
+      1,
+      $messages,
+      'The form must carry one status_messages element, not two.'
+    );
+  }
+
+  /**
+   * The messages element is not added to unrelated forms.
+   */
+  public function testUnrelatedFormIsNotAltered(): void {
+    [$form, $form_state] = $this->blockForm('inline_block:quick_links');
+
+    ys_core_form_alter($form, $form_state, 'user_login_form');
+
+    $this->assertArrayNotHasKey('ys_core_status_messages', $form);
+  }
+
+}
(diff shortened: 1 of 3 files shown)


miketullo95 (approved):
@dblanken-yale Thorough work across all three cases. Extending the existing platform-wide normaliser rather than adding a new constraint is the right call — same reasoning as #1444's gin_lb fix, go to the actual contract-holder instead of bolting on something narrower. The case-sensitivity catch on the scheme lowering, and the dead-link mechanism analysis for protocol-relative URLs, are good catches.

On the open items:
- Accessibility gap (message not tied to the field via aria-describedby) — the partial fix (visible text, 5.16:1 contrast) is a real improvement over silent failure; enabling Inline Form Errors platform-wide is a bigger call for another day, not blocking here.
- No retroactive backfill for already-broken links — opened yalesites-org/YaleSites-Internal#1515 for it, same shape as #1494's repair.
- The broader off-canvas message-loss bug on remove_block/remove_section/choose_* — opened yalesites-org/YaleSites-Internal#1516, not blocking here either.

This also resolves the open question holding #1444 back — htp:// stays a validation-rejection case with a now-visible message rather than something auto-corrected, so that PR's crash fix and this one are complementary, not conflicting. Moving #1444 to ready to merge as well.

Approving.

miketullo95 (approved):
Re-confirming approval on the current head (dd60e44), which was pushed after my last review. That commit extends the auto-correction to mistyped schemes like htp://, which goes beyond what #1497's AC specified for that case (visible rejection message) and beyond what my prior approval's reasoning assumed when I said #1444 stays complementary rather than redundant. I looked at it and I'm fine treating this as an intentional improvement — approving as-is.

@dblanken-yale

PR #1450, opened 2026-08-11
1503: RC: Decide whether to clean up orphaned blocks left behind by Layout Builder

dblanken-yale:
## [1503: RC: Decide whether to clean up orphaned blocks left behind by Layout Builder](https://github.com/yalesites-org/YaleSites-Internal/issues/1503)

> **Read this first: the issue's central premise did not hold up, so this PR needs a product decision before it needs a code review.** The issue is a "decide whether to build this" ticket carrying both `forming` and `claude`. Details below.

### The issue's five "known orphans" are not orphans

The issue names `block_content` **341, 342, 344, 346, 347** as orphans left behind by a QA run and asks the mechanism to confirm it identifies them as deletion candidates. Traced through every node revision, none of them is:

| block | type | actually referenced by | verdict |
|---|---|---|---|
| **341** | image_banner | node 95 vid **389 = current revision** | **live content on a published page** |
| **342** | video_banner | node 95 vid **389 = current revision** | **live content on a published page** |
| 344 | cta_banner | node 77 vids 393, 397 (old revisions) | revision-only reference |
| 346 | grand_hero | node 77 vid 397 (old revision) | revision-only reference |
| 347 | image_banner | node 77 vid 397 (old revision) | revision-only reference |

Confirmed three independent ways: an entity-API walk of all 245 node revisions; SQL showing node 95's current `vid` is 389 with three component rows in `node__layout_builder__layout`; and `GET /qa1157-banner-width-test` returning HTTP 200 with `image-banner` and `video-banner` markup rendered.

The issue inferred orphanhood from `/block/341/edit` returning 404 — but that is true of **all 334** non-reusable blocks on the site, orphaned or not, because non-reusable blocks have no canonical route.

**Consequence: the issue's two Verification criteria contradict each other.** It asks the mechanism to flag 341/342 *and* to never flag a block referenced by a current layout. No mechanism can do both. **This PR implements the safety criterion and deliberately fails the other**, because satisfying it means shipping a command that deletes live page content.

### The underlying concern is real

The sweep found **21 genuine orphans** the issue never identified. **Update (post-review):** re-run against a fresh copy of the Pantheon `dev` database, the current count is **4**, not 21 — the same four listed below. The other 17 were parent-deleted cases that core's own `layout_builder_cron()` collects, and `dev` has since cleaned them. Blocks 341/342/344/346/347 do not exist on `dev` at all. See the verification comment on this PR. Four match its described scenario exactly — block removed from a layout, parent node still exists, so core's cron will never reach them:

| block | type | label | parent (still exists) |
|---|---|---|---|
| 10 | pull_quote | quote1 | node 7 |
| 13 | text | test_2 | node 7 |
| 23 | button_link | button link 1 | node 22 |
| 32 | grand_hero | Grand Hero 1 | node 31 |

The core-behaviour half of the issue is confirmed: `layout_builder_cron()` collects only rows whose `inline_block_usage` parent columns are both NULL, and `InlineBlockEntityOperations::removeUnusedForEntityOnSave()` early-returns for `RevisionableInterface`, which nodes always implement.

### Description of work
- Adds `OrphanedInlineBlockCleaner` (+ interface) and the `ys-layouts:orphaned-blocks` Drush command to **`ys_layouts`** — not `ys_core`, because it needs `layout_builder` and `block_content`, which `ys_layouts` already declares.
- **Reports by default; deletes only with `--delete`. Deliberately not wired into cron** — auto-deleting editor content on a schedule is a much riskier default than leaving inert rows in place. This is a decision to ratify, not an oversight.
- Answers the issue's open revision-history question with **full safety**: a block counts as referenced if ANY revision of ANY layout-bearing entity points at ANY of its revisions. Blocks referenced only by a non-default revision are reported separately and never deleted, because rolling a node back would make them live again.
- Also checks **Layout Builder default layouts stored in config**. They belong to no entity, so an entity-only sweep would report their blocks as orphans and delete live content.
- `deleteOrphans()` takes no ID list and derives its own, so deleting a block still on a page is structurally impossible rather than merely guarded against.
- Discovers layout-bearing entity types from the field map instead of hardcoding `node`, and tests revisionability on the **entity type definition**, not the storage class — `SqlContentEntityStorage` implements `RevisionableStorageInterface` even for non-revisionable types, and Section Library's `section_library_template` is a real layout-bearing type with no revision table (it crashed the first version).
- 7 kernel tests, documentation in the module README.

### Verification already done
- `composer code-sniff` (both standards, whole custom tree): **exit 0, zero findings**.
- `ys_layouts` kernel suite: **OK, 31 tests / 334 assertions**, zero failures.
- `ys_layouts` unit suite unchanged at 93 tests with 1 **pre-existing** failure (`BlockContextualLinks`), proven pre-existing by reverting this branch's only tracked change and re-running.
- Report run on a real database cross-checked against an independently written throwaway probe: **both agree on the same 21 IDs**, with 341/342 correctly excluded.
- **`--delete` HAS now been run against a real `dev` database copy**, at @miketullo95's request. 4 orphans deleted (10, 13, 23, 32); the revision-only block survived; a second run reports empty; all four affected pages render byte-identically before and after. Full results, including an independent cross-check of the orphan set, are in the verification comment on this PR.

### Functional testing steps:
- [ ] `lando drush ys-layouts:orphaned-blocks` — reports a table of orphans and changes nothing.
- [ ] Confirm no block listed is visible on any published page (spot-check a few IDs at `/node/{parent}`).
- [ ] Create a test page, add an inline block, save; remove that block, save again; re-run the report and confirm the block now appears.
- [ ] Add an inline block to a page and leave it in place; confirm it never appears in the report.
- [x] `lando drush ys-layouts:orphaned-blocks --delete` on a throwaway DB, then re-run the report and confirm it is empty and pages still render. **Done on a fresh `dev` copy — see the verification comment.**
- [ ] Decide whether report-only-plus-opt-in-delete is the wanted shape, or whether this should be cron-driven / not built at all.

**Re-verification steps added after review (@miketullo95's `--delete` request):**
- [ ] `dbget && confim` for a fresh `dev` copy, then `lando drush cr` (a brand-new service is not in the compiled container until caches rebuild).
- [ ] `lando drush ys-layouts:orphaned-blocks` — expect 4 orphans (10, 13, 23, 32) and 1 revision-only (21) on a current `dev` copy.
- [ ] Note the row counts first: `SELECT COUNT(*) FROM block_content;` / `... WHERE reusable = 0` / `SELECT COUNT(*) FROM inline_block_usage;` (305 / 303 / 303 on `dev` today).
- [ ] `lando drush ys-layouts:orphaned-blocks --delete` — expect exactly 4 deleted and every count down by 4.
- [ ] Confirm block 21 still exists with all its revisions: `SELECT COUNT(*) FROM block_content_revision WHERE id = 21;`
- [ ] `lando drush cr`, then load `/events/dinner-with-dad`, `/blocks-for-visreg`, `/blocks-for-visreg/button`, `/blocks-for-visreg/grand-hero` — all 200, all visually unchanged.
- [ ] Re-run the report — expect "No orphaned inline blocks found" with block 21 still withheld.

References yalesites-org/YaleSites-Internal#1503

---

Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/README.md
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/README.md
@@ -24,6 +24,55 @@ This module includes a service called `MetaFieldsManager` that was specifically 
 
 The `ics_url` is auto-calculated if there is an ICS URL provided from Localist, then use it. If not, calculate an ICS URL dynamically from the first date in the series.
 
+## Orphaned inline blocks
+
+When an editor removes a non-reusable ("inline") block from a page's layout, the block
+disappears from the page but its `block_content` record stays in the database. Nothing in the
+admin UI can reach it afterwards: non-reusable blocks have no canonical route, so
+`/block/{id}/edit` returns 404 for every one of them, orphaned or not.
+
+This is documented Drupal core behaviour, not a YaleSites defect, and core cannot clean it up
+for a node that still exists:
+
+- `layout_builder_cron()` collects only `inline_block_usage` rows whose `layout_entity_type`
+  and `layout_entity_id` are **both** NULL, which happens when the parent entity is deleted.
+- `InlineBlockEntityOperations::removeUnusedForEntityOnSave()` returns early for anything
+  implementing `RevisionableInterface`. Nodes always do, so it never runs for them. Core skips
+  it deliberately: an older revision of the node might still reference the block, and core
+  cannot prove otherwise cheaply from inside a single save.
+
+### Sweeping them up
+
+A deliberate sweep can afford the proof core cannot, so this module provides one **on demand**.
+It is **not** wired into cron, because automatically deleting editor content on a schedule is a
+much riskier default than leaving inert rows in place.
+
+```bash
+# Report only. Changes nothing.
+lando drush ys-layouts:orphaned-blocks
+
+# Delete the reported orphans.
+lando drush ys-layouts:orphaned-blocks --delete
+```
+
+Safety guarantees, in order of importance:
+
+1. **Report by default.** Deletion happens only with the explicit `--delete` flag.
+2. **A block counts as referenced if *any* revision of *any* layout-bearing entity points at
+   *any* of its revisions.** This is stricter than core's own intent and is what makes deletion
+   safe.
+3. **Layout Builder default layouts stored in config are checked too.** They belong to no node,
+   so a node-only sweep would report their blocks as orphans and delete live content.
+4. **Blocks referenced only by a non-default revision are never deleted.** They are listed
+   separately in the command output, because rolling a node back to such a revision would make
+   the block live again. Choosing to delete that category would be a separate, explicit
+   decision.
+5. **Deletion derives its own list** and accepts no IDs from the caller, so removing a block
+   that is still on a page is structurally impossible rather than merely guarded against.
+
+Reusable blocks from the Custom Block Library are never candidates; only `reusable = 0` blocks
+are considered.
+
 ## Running tests
 
 This module has PHPUnit tests under `tests/src/` (`Unit/` and `Kernel/`). Run them from the project root on the local Lando environment, passing the module's `tests` path so PHPUnit only discovers this module's tests (not Drupal core/contrib):
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/drush.services.yml
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/drush.services.yml
@@ -1,0 +1,6 @@
+services:
+  ys_layouts.commands:
+    class: \Drupal\ys_layouts\Commands\YaleSitesLayoutsCommands
+    arguments: ['@entity_type.manager', '@ys_layouts.orphaned_inline_block_cleaner']
+    tags:
+      - { name: drush.command }
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Commands/YaleSitesLayoutsCommands.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Commands/YaleSitesLayoutsCommands.php
@@ -1,0 +1,105 @@
+<?php
+
+namespace Drupal\ys_layouts\Commands;
+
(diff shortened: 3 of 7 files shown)


miketullo95 (changes requested):
@dblanken-yale This is strong investigative work — catching that the issue's own "known orphans" (341/342/344/346/347) aren't actually orphans, and that its own Verification criteria contradict each other, is exactly the kind of thing that needed catching before anyone built a deletion tool around the wrong test cases. The revision-aware safety model (any revision of any layout-bearing entity, plus config-stored defaults, plus testing revisionability on the entity type rather than the storage class) is thorough.

Decisions, now made:
- **Build it** — yes. The underlying problem is real and this correctly finds the 21 genuine orphans while excluding all 5 of the ticket's wrongly-assumed ones.
- **Shape** — report-only by default, `--delete` opt-in, not cron-wired. Confirmed, no changes needed there.
- **Priority** — downgraded to Low on yalesites-org/YaleSites-Internal#1503; updated the issue directly with the corrected findings so the Verification AC doesn't send a future reader chasing the wrong block IDs.

**Blocking:** run `--delete` against a real dev/staging database copy before this merges, and report the results (count deleted, confirm pages still render for a few spot-checked IDs, confirm a second run reports empty) — the same rigor as the report-mode verification you already did. This permanently deletes content and the kernel tests, while solid, haven't been exercised against real data yet. You noted running `--delete` wasn't yours to do unprompted — consider this the prompt.

Requesting changes pending that verification.

miketullo95 (approved):
The `--delete` verification against a real dev copy checks out: 4 real orphans deleted (10, 13, 23, 32), the revision-only block (21) correctly survived all 24 revisions, second run reports empty, affected pages byte-identical before/after, and the independent cross-check matched after you tracked down the `deleted=1` field-instance discrepancy. No code changed since my last review, just a develop merge.

On the two open points from that verification comment:
- Report/delete deriving separate lists (delete re-scans fresh rather than deleting exactly the reviewed set) — fine with that trade-off. Every block it touches still has to pass the same safety check fresh at delete time either way, so this is about audit predictability, not safety, and that's an acceptable gap for a manually-run maintenance command.
- Expected orphan scale (single-digit on dev) — a README note is enough, no threshold guard needed for this PR.

Approving.

@dblanken-yale

PR #1451, opened 2026-08-12
1511: RC: Delete collection on content page turns sub pages to new collections

dblanken-yale:
### Description of work
- Deleting a content collection's top-level page from Manage Content promoted every sub-page into the top page of a brand-new collection. The pages were never lost, but editors were left with a pile of junk collections filling the Collection dropdown on every page edit form, and no bulk way to undo them.
- Root cause: contrib `BookManager::deleteFromBook()` re-points a book root's children at themselves as new book IDs. `book` and `ys_book` are both weight 0, so `module_implements()` ordered them alphabetically and `book_node_predelete()` ran first; our own predelete only invalidated caches and ran too late to matter.
- Fixed entirely inside `ys_book` — no patch to contrib `book`, per the ticket. `ys_book_module_implements_alter()` now moves `ys_book` to the front for `node_predelete`, and `ys_book_node_predelete()` takes the collection apart through the existing `_ys_book_dismantle_collection()` helper from yalesites-org/YaleSites-Internal#1350 rather than a second implementation of it. Dismantling deepest-first means no page is ever a root with children left to promote, so sub-pages come out as ordinary standalone content — the same outcome "Delete collection" on Manage Content Collections already produces.
- Deleting a page in the **middle** of a collection is deliberately unchanged: contrib moves its children up under its parent and the collection survives. There is a regression test pinning that.
- Contrib `book_node_predelete()` only checks the in-memory copy of the book data, so a page whose outline row we just removed has that copy cleared, otherwise contrib reads `bid`/`pid` off the empty array `loadBookLink()` returns. This also covers bulk delete, where selecting a collection's top page together with one of its sub-pages runs predelete for every selected node before any row is written.
- The Manage Content delete confirmation was the generic content-delete form, whose only description is "This action cannot be undone" — a large part of why losing a collection here was a surprise. It now leads with what happens to the sub-pages, matching the Manage Content Collections wording. Suppressed when the collection is only its top-level page, since then there are no sub-pages to describe.
- Second commit (`test(1505)`) repairs two `ys_book` kernel fixtures that were calling `ys_book_update_10001()`, a function yalesites-org/YaleSites-Internal#1505's fix renumbered away. All 7 `BookCollectionDeleteTest` tests and all 13 `BookLinkRepeatSaveTest` tests were erroring in `setUp()` on `develop` before reaching an assertion, so the fixture had to be repaired before the new tests could be added.

**UX decision taken:** the ticket left open whether sub-pages should become standalone pages or whether this delete path should be blocked and the editor pointed at Manage Content Collections. This implements the first option — the reporter's stated expectation, and what the rest of the ticket's acceptance criteria are written against. It also makes both delete paths behave identically, which the ticket says removes the need for a KB change.

### Please note before reviewing

- **Verified by kernel tests only — the new confirmation copy has never been rendered in a browser.** This checkout had no working local site (the local database held no `node` or `book` tables, and `lando pull` could not authenticate to Pantheon — terminus has no saved machine token in the appserver container). The end-to-end `/admin/content` click-through in the testing steps below was therefore not performed by me. The notice's presence, text, weight and conditional suppression are asserted against the real built form array, and the element is a plain `<p>` so it cannot disturb heading order, but it has not been looked at.
- **The notice does not appear on the bulk-delete confirmation.** `DeleteMultipleForm` extends `ConfirmFormBase`, not `ContentEntityDeleteForm`. The behaviour fix covers the bulk path correctly; only the explanation is single-path.
- **The fix is not retroactive.** Stray collections already created by this bug on existing sites still need manual cleanup. Worth deciding whether that cleanup needs its own ticket.

### Test evidence

- `ys_book` kernel suite: **30 tests / 237 assertions / 20 errors** on `develop` → **35 tests / 450 assertions / 0 errors** on this branch.
- `lando composer code-sniff` exits 0; `lando composer lint:php` exits 0.
- `npm run lint:styles` fails, but only on `web/themes/contrib/atomic/_yale-packages/tokens/build/scss/tokens.scss` — a generated build artifact in the local gyst'd atomic clone. `web/themes/contrib/` is gitignored, this PR contains no SCSS, and that path does not exist in CI.

### Functional testing steps:
- [ ] Create a content collection with a top-level page and at least three sub-pages, including one nested two levels deep.
- [ ] Go to Manage Content (`/admin/content`), open the edit dropdown for the collection's top-level page, and choose Delete.
- [ ] Confirm the delete message clearly states what will happen to the sub-pages before you confirm.
- [ ] Confirm the delete.
- [ ] Go to Manage Content Collections and confirm no new collections were created.
- [ ] Edit one of the former sub-pages and confirm the "Collection" dropdown shows no collection selected, and that the dropdown itself contains no junk entries.
- [ ] Confirm every former sub-page is still published and reachable at its own URL.
- [ ] Repeat with a page in the middle of a collection: delete it and confirm its children move up under its parent and the collection is otherwise intact.
- [ ] Confirm deleting a whole collection from Manage Content Collections still works as it did before.

References yalesites-org/YaleSites-Internal#1511
References yalesites-org/YaleSites-Internal#1505

---
Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_book/tests/src/Kernel/BookCollectionDeleteTest.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_book/tests/src/Kernel/BookCollectionDeleteTest.php
@@ -54,13 +54,6 @@ class BookCollectionDeleteTest extends KernelTestBase {
     _ys_book_add_book_title_column();
     $this->installConfig(['node', 'book', 'field']);
 
-    // ys_book keeps custom menu link titles in a column it adds to contrib
-    // book's table from ys_book_update_10001(). installSchema() only builds
-    // contrib book's own hook_schema(), so run the update hook to match a
-    // real site -- without it every book node save fails on a missing column.
-    $this->container->get('module_handler')->loadInclude('ys_book', 'install');
-    ys_book_update_10001();
-
     // The delete route is gated on 'administer book outlines'; build the router
     // so the confirm form's cancel link can resolve the collections overview.
     $this->container->get('router.builder')->rebuild();
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_book/tests/src/Kernel/BookCollectionDeleteTest.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_book/tests/src/Kernel/BookCollectionDeleteTest.php
@@ -6,6 +6,7 @@ use Drupal\KernelTests\KernelTestBase;
 use Drupal\Tests\user\Traits\UserCreationTrait;
 use Drupal\node\Entity\Node;
 use Drupal\node\Entity\NodeType;
+use Drupal\node\NodeInterface;
 use Drupal\ys_book\Controller\YsBookController;
 use Drupal\ys_book\Form\BookCollectionDeleteForm;
 use Drupal\Core\Form\FormState;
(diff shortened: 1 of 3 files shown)


miketullo95 (approved):
Tested deleting a content collection myself (including a grandchild-level sub-page) and it looked good — sub-pages survived as standalone content, no stray collections created.

Checked against #1511's AC: the reorder-and-dismantle fix stays inside ys_book with no contrib patch, reuses _ys_book_dismantle_collection() from #1350 rather than duplicating it, mid-collection deletes are untouched and still tested, and the Manage Content delete confirmation now explains the sub-page consequence (suppressed correctly for single-page collections).

Two open items from the PR description, decided:
- Retroactive cleanup of stray collections already created by this bug on live sites — not opening a follow-up ticket for that right now.
- The new confirmation copy not appearing on the bulk-delete path (different form class) — acceptable as-is, since the underlying behavior fix still applies there regardless of the missing copy.

Approving.

@dblanken-yale

PR #1452, opened 2026-08-12
1518: RC: Layout Section color picker: add a 6th color option (slot-three) for parity with block pickers

dblanken-yale:
### Description of work

Every block-level color picker on the platform offers six background colors; the Layout Section "Component theme" picker offered five. The missing one was slot-three, dropped by #897 and left out again when #1153 added slot-nine on the `five` key. yalesites-org/YaleSites-Internal#1506 chose to document the gap (Option A, merged as #1449) and split the real fix out as this ticket (Option B).

- `ColorTokenResolver::getColorStylesForEntity()` gains `'six' => 'three'` in the `layout_section` mapping, so the admin swatch resolves the new option under all seven global themes.
- `YSLayoutOptions::buildConfigurationForm()` exposes the matching `#options` entry.
- The label is the ordinal **"Six"**, not a color name. Section labels are deliberately ordinals because the underlying color differs per global theme: slot-three is light blue on Old Blues but red on It's Your Yale, so any single color name would be wrong on most themes. **Easy to change if a color name is preferred.**
- `processColorPicker()` previously declared the section option order twice, as `$desired_order` and again as `$palette_order` a few lines later. Since this change has to touch both copies, they are collapsed into one `$palette_order` declared before the branch. Behavior is unchanged; the non-section path still yields `NULL`.
- The two tests that pinned the five-option contract are updated: `ColorTokenResolverTest::testGetColorStylesForEntitySectionExposesFiveOptions` becomes `...ExposesSixOptions` and now asserts that option `six` resolves to slot-three under **each** of the seven global themes, and `YSLayoutOptionsTest::testBuildConfigurationFormExposesDividerAndTheme` asserts seven `#options` keys.

Global theme four (Onha) swaps slot-two and slot-five only, so slot-three passes through untouched there and the admin swatch matches the page.

- Other work completed in: yalesites-org/component-library-twig#679

**This PR does nothing on its own.** `ColorTokenResolver` only drives the admin color-picker swatches; the rendered page color comes from `_yds-layout.scss` in the companion PR. Merging this alone would add a picker option that paints no background.

### Functional testing steps:

- [ ] Edit a page, add or open a **50/50** or **33/33/33** section, and open its settings. Confirm the Component theme picker now shows **six** color swatches plus "Default - no color", matching a block-level picker such as Callout.
- [ ] Select the sixth option, save, and view the published page. Confirm the section background is the global theme's slot-three color and the text is dark and legible.
- [ ] Confirm the swatch shown in the editor matches the color that actually renders on the page.
- [ ] Repeat under several global themes, **including Onha (theme four)** -- that is the theme with the slot-two/slot-five swap, so it is the one most likely to disagree between swatch and page.
- [ ] Confirm the existing five options and "Default - no color" are unchanged.
- [ ] Confirm yalesites-org/YaleSites-Internal#1153's release-testing Step 1 (which already says sections should show 6 slots) now needs no correction.

### Verification performed

- `lando composer code-sniff` -- exit 0.
- `composer lint:php` -- no syntax errors.
- `ys_themes` and `ys_layouts` unit suites run before and after. Both new assertions pass. Pre-existing failures are unchanged in count **and name** (see below), so there are no regressions from this change.
- Compiled-CSS verification of the companion branch: built the component library and confirmed `dist/css/style.css` emits `.yds-layout[data-component-theme=six]` with `background-color: var(--color-layout-theme)` resolving to `--color-slot-three`, and that `--color-slot-three` resolves per global theme -- including theme four, where the swap re-declares only slot-two and slot-five, leaving slot-three untouched.

### Not verified -- needs a human

**There is no browser verification in this PR.** The local Drupal database had no content (no `node` table) and could not be restored, so the Layout Builder UI and the published page were never actually loaded. The functional testing steps above are all still owed. The claim most worth checking first is that the editor swatch and the rendered page agree under Onha.

### Pre-existing issues found, not fixed here

- Four unit tests in `ColorTokenResolverTest` **already fail on `develop`** and still fail: `testBuildColorInfoResolvesComponentThemeVariable`, `testGetColorStylesForEntityCalloutBundleMapping`, `testGetColorStylesForEntityFactsBundleUsesFourOptionOverride`, `testGetColorStylesForEntityQuoteCalloutBundleMapping`. They expect `getColorStylesForEntity()` to emit a `var(--component-themes-*-background)` direct override for the callout/facts/quote-callout `five` option, but `buildColorStyles()` only ever emits `var(--global-themes-*-colors-slot-*)` -- no code path produces that value. These assert behavior the code does not have. They are untouched by this PR (different bundles), but they are genuinely broken and worth their own ticket. CI runs no PHPUnit (`composer unit-test` is an echo stub), which is why this went unnoticed.
- `Drupal\Tests\ys_layouts\Unit\BlockContextualLinksTest::testRemoveStaysLastAndLabelsDropTheWordBlock` also already fails on `develop` (it does not expect `ys_layouts_block_detach`). Unrelated to this change.
- `theme` is missing from `YSLayoutOptions::defaultConfiguration()` while `divider` is present. Both the form and the Twig templates compensate with their own fallbacks, so nothing is broken today, but any third consumer would need its own fallback. Left alone as out of scope.

References yalesites-org/YaleSites-Internal#1518

---

Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Plugin/Layout/YSLayoutOptions.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Plugin/Layout/YSLayoutOptions.php
@@ -80,9 +80,11 @@ class YSLayoutOptions extends LayoutDefault implements ContainerFactoryPluginInt
 
     // Use the saved theme value directly from configuration.
     $saved_theme = $this->configuration['theme'] ?? 'default';
-    // Sections offer five color options by design, where block components
-    // offer six. See ColorTokenResolver::getColorStylesForEntity() for which
-    // palette slot each option resolves to and why slot-three is excluded.
+    // Sections offer six color options, matching the block component pickers
+    // (#1518). See ColorTokenResolver::getColorStylesForEntity() for which
+    // palette slot each option resolves to. Labels are deliberately ordinals
+    // rather than color names: the underlying color differs per global theme,
+    // so 'six' is light blue on Old Blues but red on It's Your Yale.
     $form['theme'] = [
       '#type' => 'select',
       '#title' => $this->t('Component theme'),
@@ -94,6 +96,7 @@ class YSLayoutOptions extends LayoutDefault implements ContainerFactoryPluginInt
         'three' => $this->t('Three'),
         'four' => $this->t('Four'),
         'five' => $this->t('Five'),
+        'six' => $this->t('Six'),
       ],
       '#weight' => 10,
       '#after_build' => [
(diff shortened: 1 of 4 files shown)


dblanken-yale:
Closing for now until we get the section work in.

PR #1454, opened 2026-08-12
1517: RC: og:image can still comma-split if an uploaded image's filename contains a comma

dblanken-yale:
### Description of work

**The reported bug does not reproduce, so no production code changed here.** The
ticket's first acceptance criterion was to confirm it before fixing, and confirming it
showed there is nothing to fix on our side. What ships is the ticket's third
acceptance criterion turned into a regression test, so the invariant is enforced
rather than remembered.

Why a comma in an uploaded filename cannot split `og:image`:

- The media image `token` view display renders `field_media_image` with the
  `image_url` formatter and no image style, so `ImageUrlFormatter` calls
  `FileUrlGenerator::generateString()`.
- That reaches `PublicStream::getExternalUrl()`, which passes the path through
  `UrlHelper::encodePath()`, i.e. `str_replace('%2F', '/', rawurlencode($path))`.
  `rawurlencode` percent-encodes a comma, so `sunset, beach.jpg` arrives as
  `sunset%2C%20beach.jpg`.
- `MetaNameBase::output()` then splits on `metatag.settings:separator` (a comma) and
  finds nothing to split.

The encoding is structural, not incidental: an image style would route through
`ImageStyle::buildUrl()`, which also ends at
`FileUrlGenerator::generateAbsoluteString()` and the same `encodePath`. Editors cannot
bypass it either, since `metatag.settings:entity_type_groups` exposes only the `basic`
and `ys_beacon` groups per node bundle, not `open_graph`, so `og:image` is never typed
by hand.

**No metatag patch was added.** The upstream defect the ticket cites is real: the
embedded-image check in `parseImageUrl()` is `strpos($value, '`. That is the mechanism that garbled #1495. But it is
unreachable here now that yalesites-org/yalesites-project#1445 routes the token
through `image_url`, and a `patches/contrib/metatag/` entry would be a permanent cost
on every metatag bump for no reachable code path. A bug report is drafted for filing
upstream separately.

Two cases are added to the existing #1495 test class, since they share its media,
file and token-display fixtures:

- A file named `sunset, beach.jpg` resolves through the real shipped view display and
  metatag's own `og_image` plugin to exactly one absolute tag ending in the fully
  encoded name. Asserting through the plugin rather than on the token string matters,
  because `output()` also runs `parseImageUrl()`, `tidy()`'s whitespace collapse,
  absolute-URL handling and `trimValue()`, any of which could mangle an encoded URL.
- A control case feeding `og:image` a raw, unencoded comma, which metatag does split
  into several tags. Without it the one-tag assertion could pass for the wrong reason
  and nobody could tell.

Also pins `metatag.settings:separator` to the platform's exported value so the tests
read production's separator instead of contrib's blank default, and folds three
repeated reads of it into a `separator()` helper.

### Testing limitation, please read before reviewing

**No browser or visual verification was performed on this branch.** The local Drupal
database was empty (no `node` table) and Terminus had no machine token, so no site
could be brought up to render a real page. All evidence here is from kernel tests
against their own isolated test database, plus reading core and contrib source. That
is a good fit for this particular change, since the whole question is PHP value
handling with no theme or browser involvement, but it does mean nobody has looked at a
rendered `<meta property="og:image">` tag in this run.

Recommendation once the multidev is up: confirm the behaviour once in a browser using
the steps below, then **verify and close**
yalesites-org/YaleSites-Internal#1517 rather than treating it as a code fix.

### Functional testing steps:

- [ ] `lando composer code-sniff` exits 0 (it does locally, along with `lint:php`).
- [ ] Run the test file and see 4 passing tests:
      `lando ssh -c "env SIMPLETEST_DB=mysql://pantheon:pantheon@database/pantheon?module=mysql SIMPLETEST_BASE_URL=http://appserver vendor/bin/phpunit -c /app/phpunit.xml web/profiles/custom/yalesites_profile/modules/custom/ys_core/tests/src/Kernel/MediaImageTokenTest.php --testdox"`
- [ ] On the multidev, upload an image whose filename contains a comma, e.g.
      `sunset, beach.jpg`, into the media library.
- [ ] Set it as the teaser image on a Post and view the published Post.
- [ ] View source and confirm there is exactly **one** `<meta property="og:image">`
      tag, that its URL is absolute, and that the filename appears encoded as
      `sunset%2C%20beach.jpg`.
- [ ] Optionally paste the Post URL into Slack or the Facebook sharing debugger and
      confirm the preview image resolves.

References yalesites-org/YaleSites-Internal#1517

---

Created with AI

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/tests/src/Kernel/MediaImageTokenTest.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/tests/src/Kernel/MediaImageTokenTest.php
@@ -23,6 +23,10 @@ use Drupal\Tests\user\Traits\UserCreationTrait;
  * core's default image formatter and the token resolves to a whole <img> tag
  * with the alt text inside it.
  *
+ * The separator can also arrive from the uploaded file's own name, which
+ * nothing sanitizes, so this also carries the chain through metatag's og:image
+ * plugin and asserts on the tags it actually emits.
+ *
  * @group ys_core
  * @group yalesites
  */
@@ -42,6 +46,14 @@ class MediaImageTokenTest extends KernelTestBase {
   private const FILE_NAME = 'sculpture-in-snow.jpg';
 
   /**
+   * A file name containing metatag's separator, which nothing sanitizes away.
+   *
+   * Every option under file.settings:filename_sanitization is off, so an editor
+   * can upload this name verbatim.
+   */
+  private const SEPARATOR_FILE_NAME = 'sunset, beach.jpg';
+
+  /**
    * {@inheritdoc}
    */
   protected static $modules = [
@@ -52,6 +64,8 @@ class MediaImageTokenTest extends KernelTestBase {
     'image',
     'media',
     'token',
+    'metatag',
+    'metatag_open_graph',
   ];
 
   /**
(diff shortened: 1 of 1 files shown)


miketullo95 (approved):
Approving on the kernel-test evidence — the trace (image_url formatter → FileUrlGenerator → encodePath percent-encoding the comma before metatag ever sees it) is convincing, and the added regression tests (including the control case that proves an unencoded comma really would split) lock the invariant in properly.

One ask: please file the drafted upstream bug report for metatag's broken  detection on drupal.org and link it back on #1517 for tracking. It's not reachable in our code today, but it's a real defect and worth having on record in case the image formatter ever changes.

@dblanken-yale

PR #1462, opened 2026-08-13
1526: Feature Request: CSV Import for Resources

dblanken-yale:
### Description of work

- Adds a CSV bulk importer for the Resource content type at `/admin/content/manage-resources/csv-import`, gated on `create resource content`, with a **CSV Import** action link on Manage Resources beside the existing **Export to CSV**.
- Mirrors the existing Profiles importer rather than starting a second import framework. `TaxonomyResolverService` is reused as-is, and `CsvValidatorService` grows a shared `parseCsv()` that both importers call instead of being copy-pasted. **The Profiles path is unchanged.**
- Preview-only mode is on by default and saves nothing; the import summary reports created / skipped / errored and **lists which resources still need Resource Media attached**.
- Columns: Title (required), Description, Abstract, Citation, Journal Publication Name, Journal Publication Issue, Resource Category, Audience, Custom Vocab, Resource Publication Date, Date Format, Tags, Teaser Title, Teaser Text, External Source, CAS Login Required, Pin to beginning of list.

### Three places the ticket and the data model disagreed

Worth a look, since two change what the ticket asked for:

1. **Resource Media is not a required field.** `field.field.node.resource.field_media.yml` is `required: false`, and a programmatic save runs no form-level validation anyway. The AC asking for "field validation adjusted for the import path so nodes can save without it" needed **no code** — imported nodes pass `$node->validate()` with **zero violations**. The AC's intent is met instead: nodes import without media, and the summary flags them.
2. **Affiliation does not exist on Resource.** Only `node.profile` has `field_affiliation`. The nearest Resource field is `field_authors` ("Affiliated Authors"), an entity reference to *profile nodes* — a different thing needing a node-title resolver. **Dropped from the column set.**
3. **Teaser Media can't come from a CSV cell either.** It is a `media:image` reference — the exact limitation the ticket already accepted for Resource Media, but it was still listed in the mapping AC. **Dropped, for the same stated reason.**

### Decisions made

- **Duplicates key on Title** (the AC asked for this decision) — the only required, human-meaningful field. "Skip duplicates" defaults on; unchecking it *is* the allow-duplicates option, so a second toggle would be redundant. Titles repeated **within one file** count too, so the preview can't promise more than the import delivers.
- **Dates** accept `YYYY-MM-DD` and `MM/DD/YYYY`, padded or not (Excel writes `3/4/2026`), and are round-trip checked so `2026-02-31` is rejected rather than rolled to March. `strtotime()` was deliberately avoided — a silently wrong publication date is worse than a rejected row.
- **A bad cell fails only its own row**, reported by real CSV line number, instead of rejecting the whole file for one typo.
- **Values are read from their owning config**, not hardcoded: allowed Date Format options, the Teaser Text length cap, and the text format per long-text field.
- **Resources import as drafts** — the editorial workflow's own `default_moderation_state`. Setting `status` instead is silently overridden.
- **`Custom Vocabulary` and `CAS Protected` are accepted header aliases** so a `ys_content_export` CSV can be edited and imported back. Unrecognised headers are **warned about** rather than dropped in silence.

### Two pre-existing bugs fixed in shared code

Both are in services the new importer routes through, so the Profiles importer benefits too:

- `parseCommaSeparatedValues()` filtered before trimming, so a whitespace-only segment survived and **created a nameless taxonomy term**.
- `findOrCreateTerm()` re-queried every occurrence of a repeated term name. It now resolves each distinct name once per request — a 500-row import was issuing thousands of uncached term queries.

Two existing tests *characterised* these as intended behaviour; both were rewritten to assert the corrected behaviour.

### Known follow-ups (not in scope here)

- `ProfileImportService::createProfileNode()` sets `'status' => 1` on a bundle under the editorial workflow, so it is silently overridden and imported profiles land as drafts anyway — same class of bug as the one fixed here for resources, but in shipped, tested, out-of-scope code.
- No Batch API: a very large import runs in one request, matching the existing Profiles importer.
- `mockQuery()` is now duplicated in three test classes, and `submitForm()` is identical between the two import forms — both are extraction candidates once a third importer appears.
- The importer creates terms in `field_audience` / `field_custom_vocab`, whose node-form widgets do not allow term creation. Consistent with the Profiles importer, so flagging it as a product call rather than changing it silently.
- The user-guide documentation AC is not covered here — that content lives on yalesites.yale.edu, outside this repo.

### Verification

- **Unit tests: 125 tests / 408 assertions, 0 failures.** The 88 / 247 figure above predated the batch commits; the pre-change baseline for the most recent round was **123 / 353**, and the four added columns brought two new drift-guard tests with them. Existing Profiles and TaxonomyResolver tests stayed green throughout, which matters because this edits shared services.
- **Runtime on Lando**, not just green tests: real imports created nodes whose fields were read back and checked (both date shapes normalised, terms in the right vocabularies, link stored as a URI, teaser fields omitted when blank, `field_media` empty); `$node->validate()` returned 0 violations; the browser flow was driven end to end for preview, import, and duplicate re-import.
- **Text-format check, red/green as a real editor** (not user 1, who bypasses every permission check): both long-text fields render **editable**; forcing the previously-chosen `basic_html` back onto the same node reproduces a **disabled** widget and a "sufficient permissions" notice. Both fields restrict `allowed_formats` to a single format and no role holds `administer filters`, so the format now comes from each field's own config.
- **Accessibility (light gate): PASS** — contrast, images/media, form labels, interactive names, scoped to our own markup.
- **Security review: no high-confidence findings.** Checked route permission, `sticky`/`field_login_required` (not escalation — both are ordinary widgets on the node form for these roles), XSS via `@` placeholders and the rendered preview table (both escape), the External Source scheme allow-list (rejects `javascript:`, `data:`, protocol-relative), entity-query injection, mass assignment, and the private-filesystem upload.
- **`lando composer code-sniff` exit 0**, `lando composer lint:php` exit 0.

### Functional testing steps:

- [ ] As an editor, go to **Content → Manage Resources** and confirm a **CSV Import** action link appears next to **Export to CSV**.
- [ ] Open the form and confirm the **Recognised columns** table lists all 17 columns, and that **Preview only** and **Skip duplicates** are both checked by default.
- [ ] Upload a CSV with a header row and a few resources. With **Preview only** checked, submit and confirm a preview table appears, the **Needs media** column reads Yes for rows without an External Source, and **no resources were actually created**.
- [ ] Uncheck **Preview only** and submit. Confirm the resources are created, and that the warning lists exactly the ones with no External Source as still needing Resource Media.
- [ ] Confirm the imported resources are **Drafts**, and that opening one shows Description and Teaser Text as **editable** (not greyed out with a permissions notice).
- [ ] Include Abstract, Citation, Journal Publication Name and Journal Publication Issue values in a row, import it, and confirm all four land on the node -- and that **Abstract and Citation are editable** on the node form (not greyed out with a permissions notice), the same check as Description and Teaser Text above.
- [ ] Submit the same file again with **Skip duplicates** checked and confirm it reports them as skipped rather than creating a second copy.
- [ ] Include one row with a bad date (e.g. `not a date`) and one with a bad link (e.g. `example.com`) and confirm each is reported against **its own row number** while the other rows still import.
- [ ] Rename a column to something unrecognised (e.g. `Tag` instead of `Tags`) and confirm the form warns that it will be ignored.
- [ ] Confirm the Profiles CSV import at **Manage Profiles → CSV Import** still works unchanged.

References yalesites-org/YaleSites-Internal#1526

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_migrate/src/Batch/CsvImportBatch.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_migrate/src/Batch/CsvImportBatch.php
@@ -26,34 +26,30 @@ class CsvImportBatch {
   const CHUNK_SIZE = 50;
 
   /**
-   * Builds a batch definition that imports $rows through $import_service_id.
+   * Builds a batch definition that imports $rows through $options.
    *
-   * @param string $import_service_id
-   *   Service id of an import service exposing
-   *   processImport(array $rows, bool $skip_duplicates): array, returning
-   *   'created', 'skipped', 'errors' and (optionally) 'needs_media' keys.
+   * @param array $options
+   *   Batch options: 'import_service_id' (service id of an import service
+   *   exposing processImport(array $rows, bool $skip_duplicates): array,
+   *   returning 'created', 'skipped', 'errors' and, optionally,
+   *   'needs_media' keys), 'skip_duplicates' (whether to skip duplicates),
+   *   and 'entity_label' (singular noun for the imported entity, e.g.
+   *   'profile' or 'resource').
    * @param array $rows
    *   The full set of already-parsed CSV rows.
-   * @param bool $skip_duplicates
-   *   Whether to skip duplicates.
    * @param int $file_id
    *   The uploaded file entity id, deleted once the batch finishes.
-   * @param string $entity_label
-   *   Singular noun for the imported entity, e.g. 'profile' or 'resource'.
    * @param string $title
    *   The batch progress title.
    *
    * @return array
    *   A batch definition suitable for batch_set().
    */
-  public static function build($import_service_id, array $rows, $skip_duplicates, $file_id, $entity_label, $title) {
+  public static function build(array $options, array $rows, $file_id, $title) {
     $operations = [];
 
     foreach (array_chunk($rows, self::CHUNK_SIZE) as $chunk) {
-      $operations[] = [
-        [static::class, 'processChunk'],
-        [$import_service_id, $chunk, $skip_duplicates, $entity_label],
-      ];
+      $operations[] = [[static::class, 'processChunk'], [$options, $chunk]];
     }
 
     // Runs last regardless of how many chunks preceded it, so the upload
@@ -71,21 +67,23 @@ class CsvImportBatch {
   /**
    * Batch operation: imports one chunk and folds its results into $context.
    *
-   * @param string $import_service_id
-   *   Service id of the import service to invoke.
+   * Each chunk's 'errors' and 'needs_media' are appended as their own
+   * sub-array rather than merged into a single running list, so this stays
+   * O(1) per chunk regardless of how many rows have failed or need media so
+   * far; finished() flattens the sub-arrays once, at the end.
+   *
+   * @param array $options
+   *   The 'import_service_id', 'skip_duplicates' and 'entity_label' options
+   *   described in build().
    * @param array $chunk
    *   The rows in this chunk.
-   * @param bool $skip_duplicates
-   *   Whether to skip duplicates.
-   * @param string $entity_label
-   *   Singular noun for the imported entity, carried through to finished().
    * @param array $context
    *   The batch context, passed by reference.
    */
-  public static function processChunk($import_service_id, array $chunk, $skip_duplicates, $entity_label, array &$context) {
+  public static function processChunk(array $options, array $chunk, array &$context) {
     if (!isset($context['results']['created'])) {
       $context['results'] = [
-        'entity_label' => $entity_label,
+        'entity_label' => $options['entity_label'],
         'created' => 0,
         'skipped' => 0,
         'errors' => [],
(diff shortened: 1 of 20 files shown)


miketullo95 (changes requested):
Great level of documentation on the tradeoffs here, the reasoning on dropping Affiliation and Teaser Media (and why Resource Media needed no code change) is genuinely useful for future reference. I checked those decisions against the actual Resource content type config and they check out.

One thing before this can merge:

**Blocking**

Four Resource fields are missing from the CSV mapping: Abstract (`field_abstract`), Citation (`field_citation`), Journal Publication Name (`field_journal_publication_name`), and Journal Publication Issue (`field_journal_publication_issue`). These aren't in the ticket's AC either, so this is new scope rather than something the PR missed against spec, but all four are plain fields on Resource (two `text_long`, two `string`), the same shape as Description/Teaser Text/Teaser Title, which the importer already handles. Please add them following the existing pattern:

- `CsvValidatorService::EXPECTED_RESOURCE_COLUMNS` (`CsvValidatorService.php:44`)
- `ResourceCsvImportForm::columnNotes()` (`ResourceCsvImportForm.php:275`)
- `ResourceImportService::prepareResourceData()` and `createResourceNode()` (`ResourceImportService.php:174`, `:233`) — Abstract and Citation are `text_long`, so they'll need the same `textFormat()` handling as `field_content_description`
- `ResourceCsvSampleController::exampleRow()` (`ResourceCsvSampleController.php:66`)
- Matching coverage in `ResourceImportServiceTest`, `CsvValidatorServiceTest`, `ResourceCsvImportFormTest`, and `ResourceCsvSampleControllerTest`

**Confirmed, no action needed**

- Affiliation dropped from the column set: confirmed Resource has no `field_affiliation`, and the nearest field (`field_authors`) needs a different node-title resolver than the taxonomy resolver used here. Scoping it out is fine.
- User guide documentation AC: I'm opening a separate follow-up ticket for that, no action needed in this PR.

@dblanken-yale

miketullo95 (approved):
The four missing fields from my last review are all in now — Abstract, Citation, Journal Publication Name, and Journal Publication Issue are wired through consistently: `CsvValidatorService::EXPECTED_RESOURCE_COLUMNS`, `ResourceCsvImportForm::columnNotes()`, `ResourceImportService::prepareResourceData()`/`createResourceNode()` (with the right `text_long` format handling for Abstract/Citation), and the sample CSV controller. Test coverage is solid, including a drift-guard test (`testEveryExpectedColumnIsReadIntoPreparedData`) that fails automatically if a future column gets declared but never read.

I also see Batch API support landed since my last pass — that was a listed "known follow-up," not something I asked for, but it's a good addition given a large CSV could otherwise time out a single request.

I ran a fresh CSV with all 17 columns populated (including the four new fields) through the multidev — import worked as expected.

Approving.

@dblanken-yale

PR #1469, opened 2026-08-14
1550: Bug: Beacon system instructions treat the recommended 4,000-character length as a hard maximum

dblanken-yale:
### Description of work
- Removed the hard validation error in `SystemInstructionsForm::validateForm()` that blocked saving Beacon system instructions over 4,000 characters — the field's own label and character counter already called this a "recommendation," but the form rejected the save outright. It's now a soft cap: the existing counter still warns editors when they're over (turns to a bold warning/error color), but saving is no longer blocked.
- Renamed `getMaxInstructionsLength()` to `getRecommendedInstructionsLength()` and reworded the config schema label and a stale JS comment that both still described the old hard-cap behavior.
- Added PHPUnit coverage (`tests/src/Unit/SystemInstructionsFormTest.php`) confirming instructions over the configured length validate cleanly, plus a regression check that whitespace/markup-only input is still rejected as empty.

### Functional testing steps:
- [ ] Go to **Configuration → YaleSites Beacon settings → System instructions**
- [ ] Paste or type instructions longer than 4,000 characters
- [ ] Confirm the counter below the editor turns to a bold warning/error color once you're near/over the limit, but nothing blocks you from typing further
- [ ] Click **Save Instructions** — confirm it saves successfully ("System instructions saved as version N.") instead of showing "Instructions are N characters, which exceeds the maximum of 4,000 characters."

References yalesites-org/YaleSites-Internal#1550

Screenshot 2026-08-14 at 9 00 37 AM

Screenshot 2026-08-14 at 9 00 34 AM

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/config/schema/ys_beacon.schema.yml
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/config/schema/ys_beacon.schema.yml
@@ -67,7 +67,7 @@ ys_beacon.settings:
       label: 'Context window in tokens of the model Portkey routes to'
     instructions_max_length:
       type: integer
-      label: 'Maximum system instructions length in characters'
+      label: 'Recommended system instructions length in characters'
     instructions_warning_threshold:
       type: integer
       label: 'Character count at which the instructions editor shows a warning'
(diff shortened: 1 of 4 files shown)


miketullo95 (approved):
Clean fix, matches #1550 exactly. The hard validation block in `SystemInstructionsForm::validateForm()` is gone, saving is no longer blocked over 4,000 characters, and the warning styling in `system-instructions.js` still fires to nudge editors without stopping them. Confirmed the field label already said "recommended" pre-fix, so no other stale hard-cap language is left behind.

Test coverage looks right — `testValidateFormAllowsInstructionsOverRecommendedLength` and the whitespace/markup-only regression check.

Approving.

@dblanken-yale

PR #1470, opened 2026-08-14
1549: Layout Section: add Component theme picker support to One column layout

dblanken-yale:
### Builds on

This PR stacks on **unreviewed** yalesites-org/yalesites-project#1465 (base branch
`1541-color-slot-9-section-picker`) — should not merge before #1465 does. It reuses
that PR's `YSLayoutOptions`/`YsLayoutsDefinitionsTest` pattern for a layout core owns
(`layout_onecol`), rather than one `ys_layouts.layouts.yml` declares.

### Retargeted from Banner

This is the retargeted follow-up to now-closed yalesites-org/yalesites-project#1468,
which implemented this on the **Banner** layout instead. That approach was abandoned:
Banner is locked out of "Configure Section" for content editors
(`layout_builder_lock`), and even where reachable, its blocks (hero image/video)
typically fill the section, so a background color wouldn't be visible — the PR's own
"Known follow-up" note found this independently. See
yalesites-org/YaleSites-Internal#1549 for the full investigation. `layout_onecol`
("One column") is the actual gap: the generic, fully-editable main-content section
every page gets, and — once #1465 merges — the only user-addable Section layout still
missing the picker (50/50 ✅, 33/33/33 ✅, 70/30 via #1465, One column ❌ until this PR).

### Description of work

- New `hook_layout_alter()` in `ys_layouts.module` attaches a new
  `YSLayoutOneColumn` class (extends `YSLayoutOptions`) to core's `layout_onecol`
  plugin. It's a core (`layout_discovery`) plugin, not one `ys_layouts.layouts.yml`
  declares, so its `class` can't be set there the way #1465 sets it for
  `ys_layout_two_column` — it has to be attached at runtime instead.
  `LayoutDefinition::setLibrary()` replaces rather than merges, so the same alter
  attaches a new library (`ys_layouts/ys_layout_onecol`) that carries forward core's
  own `.layout--onecol .layout__region { width: 100%; }` rule alongside the new
  themed-section CSS, so it isn't silently dropped.
- `YSLayoutOneColumn` hides (not `unset()`s) the inherited Divider checkbox — a single
  region has nothing to divide — via `#access = FALSE`, and overrides
  `defaultConfiguration()` to normalize `divider` to an `int` default. `unset()` broke
  config round-tripping (`YSLayoutOptions::buildConfigurationForm()` reads
  `$this->configuration['divider']` before the subclass's removal would run); the
  int-vs-string mismatch matters because a real checkbox submission always stores an
  int, and the new config schema below now enforces the type.
- New `config/schema/ys_layouts.schema.yml` fixes a real, previously-latent
  `SchemaIncompleteException` for `layout_onecol` sections (`divider`/`theme`/the
  section-padding third-party key were all undeclared) — surfaced by exercising a real
  `core.entity_view_display.*` save under strict schema checking, which this repo's own
  exported config already does (`core.entity_view_display.node.resource.default.yml`
  has `padding_options` on `layout_onecol` sections today). **Scoped only to
  `layout_onecol`** — the same gap is real for `ys_layout_two_column` (#1465),
  `ys_layout_two_column_50_50`, and `ys_layout_three_column_33_33_33`, but fixing those
  is a deliberate follow-up left to reviewer judgment rather than expanded silently (see
  Known follow-up below).
- Atomic's `layout--onecol.html.twig` (companion PR yalesites-org/atomic#504) adds the
  `yds-layout` class and `data-component-theme`/`data-component-padding` attributes —
  only when a theme other than "Default - no color" is picked, so an unthemed section's
  markup is byte-identical to before this change. Deliberately does **not** embed
  `@organisms/layout/layout` the way Two Column/Three Column do: this layout's own
  child components already carry their own site-width/gutter containment, so embedding
  the organism's `__inner` wrapper would double it up, and its unconditional flex
  layout would shrink a single region to its content width instead of filling the
  section.
- New scoped `ys_layouts/layouts/onecol/onecol.css` supplies the themed-only vertical
  padding and the `data-component-padding` (`no-top`/`no-bottom`/`no-padding`) handling
  `_yds-layout.scss`'s `.yds-layout__inner` normally provides — including a
  `--main-content-bottom-margin` override so a themed One column section that happens
  to be the last section on a page doesn't trap the theme's own 10rem bottom margin
  inside this rule's padding (verified against the theme's compiled
  `layout-builder.css`, matching the two-column reference: 128px top/bottom either way).
- `docs/color-theme.md` documents the `layout_onecol` exception and that Banner is
  intentionally not themed.

### Other work completed in

yalesites-org/atomic#504

### Known follow-up / open question (resolved in review)

1. ~~Is the padding-block breathing room right, or should it go through a design
   pass?~~ **Answered:** no design pass needed — consistency with 50/50 and
   33/33/33 is what we want, so `var(--size-spacing-10)` stays.
2. Same open question closed PR #1468 raised and left alone: on a themed section whose
   block fills the section, the color isn't visible — inherent to the ask, not a defect
   here.
3. ~~The config-schema gap this PR fixes for `layout_onecol` is real for the other
   three layouts too.~~ **Done in this PR** at the reviewer's request:
   `ys_layout_two_column`, `ys_layout_two_column_50_50`, and
   `ys_layout_three_column_33_33_33` are now mapped as well. Doing it surfaced
   that the mapping alone was not enough — `YSLayoutOptions` defaulted `divider`
   to `''` while the schema declares an integer, which would have traded a
   missing-schema failure for a wrong-type one, so the shared default is now
   int `0` and `YSLayoutOneColumn`'s single-layout workaround is removed.
   `ys_layout_banner` and `ys_layout_page_meta` take only the padding key and
   remain unmapped, flagged for the reviewer.

### Functional testing steps

- [ ] On any page, add or edit a "One column" Section (`layout_onecol`), choose
      **Configure Section**. Confirm a **Component theme** picker is present with the
      color swatches, and that there is **no** Divider checkbox.
- [ ] Pick a color, save the section, then **Save layout**. Confirm the section renders
      with that background color and readable text, with breathing room above and
      below the content matching Two Column/Three Column sections.
- [ ] Re-open the section, set the theme back to **Default - no color**, and save.
      Confirm the section returns to exactly its previous appearance.
- [ ] Try the **Section padding options** (No top/No bottom/No padding) on a themed
      One column section and confirm the corresponding edge's spacing is removed.
- [ ] If a themed One column section is the last section on a page, confirm there
      isn't an oversized gap of color below the content.
- [ ] Repeat with a few different global themes and confirm the section color changes
      with the global theme and text stays readable.
- [ ] Confirm Two Column (50/50), Three Column (33/33/33), and Two Column (70/30)
      (#1465) still behave exactly as before.

Added after review:

- [ ] Add a **One column** Section, pick a colour, and **save the layout without
      putting any blocks in it**. Confirm the colour band now appears (it did not
      before), roughly the same height as a themed Two Column section.
- [ ] Set that same empty section back to **Default - no color** and save.
      Confirm it renders nothing at all, i.e. the page looks exactly as it did
      before this PR.
- [ ] On a **Post**, an **Event**, and a **Page** that have never had their
      layout customised, confirm the gap between the last piece of content and
      the site footer is unchanged. (Those default displays ship a trailing
      empty One column section; this is the regression the themed-only guard
      exists to avoid.)
- [ ] Confirm the first element on a page still has its normal top gap below the
      breadcrumb.

References yalesites-org/YaleSites-Internal#1549


miketullo95 (approved):
@dblanken-yale Approving this. I tested on the multidev and the core ask is met: the Component theme picker shows up on One column sections, the colors render, and the Divider checkbox is correctly gone. That was the whole point of the ticket and I can confirm I see it.

A few notes below. None of them block the approval.

---

### Merge order

This is based on `1541-color-slot-9-section-picker` (#1465), which is still unreviewed. I'm marking this `ready to merge` for the label's sake, but please hold it until #1465 lands. Flagging so nobody reads the label as "merge now."

---

### Non-blocking: empty themed section renders nothing on save

Found this while testing. If I pick a color on a One column section but don't put any blocks in it, I don't see the color band after I hit Save layout. Drop a block in and the section shows up fine, so it's not breaking anything real. But I'm fairly sure 50/50 and 33/33/33 don't behave that way, which is why it caught my eye.

Possible starting point: atomic's `templates/layouts/layout--onecol.html.twig` guards the whole wrapper on `{% if content.content %}` (that one specific region), while `ys_layouts/layouts/two_column_50_50/layout--two-column--50-50.html.twig:13` guards on `{% if content %}` (the whole array). Different guard, so plausibly different empty-state behavior. Worth a quick look to confirm that's actually the cause.

I don't think this needs to hold up the PR. An empty colored section isn't a real editor workflow. But if the fix is small, I'd rather One column match the other layouts than have editors hit an inconsistency later.

---

### Non-blocking, but I'd like it in this PR: extend the config schema

You flagged this one for me in the Known follow-up section, so here's my answer: I'd rather the schema fix land here than as a follow-up ticket.

`ys_layouts/config/schema/ys_layouts.schema.yml` currently only maps `layout_onecol` to `ys_layouts.layout_settings.theme_and_divider`. Same `SchemaIncompleteException` gap is real for `ys_layout_two_column` (#1465), `ys_layout_two_column_50_50`, and `ys_layout_three_column_33_33_33`, and you've already noted it's one line each with the reusable type you built. Splitting a three-line fix into its own ticket costs more than it saves, and a half-fixed schema is the kind of thing that gets forgotten.

Not gating my approval. If you'd rather not expand the diff, say so and I'll open a ticket instead.

---

### Padding question: matching the others is right

On your other open question, the `var(--size-spacing-10)` top and bottom padding: no design pass needed. Consistency with 50/50 and 33/33/33 is exactly what we want here. An editor switching between section types shouldn't see the spacing change.

---

### Docs

I'm opening a `Docs:` ticket on my end for the yalesites.yale.edu user guide. This is a visible change for editors on the section every page gets, so it needs a write-up. Nothing needed from you on that.

PR #579, opened 2024-02-22
YSP-353: Release Bug: Footer link style -- DO NOT MERGE

dblanken-yale:
### Description of work
- Fix footer shadow in Component Library

### Functional testing steps:
- [ ] Log in and attempt to change footer styles
- [ ] Verify that footer links look ok (no awful shadow taking over the link)

PR #1475, opened 2026-08-15
579: YS Core cleanup — Phase 1: extract ys_media

dblanken-yale:
### Description of work

Phase 1 of the ys_core cleanup: extracts the media/image cluster into a new `ys_media` module. **Behavior-preserving — no functionality changes.** `ys_core.module` drops from 1,705 to 1,665 lines.

Moved, unchanged apart from namespace and asset paths:

- `src/YaleSitesMediaManager.php` → `ys_media`, as `ys_media.media_manager`, with `ys_core.media_manager` kept as a **real alias** (hard constraint 3). It keeps reading `ys_core.site` (hard constraint 1).
- `ys_core_imagemagick_arguments_alter()` (PNG transparency) → `ys_media.module`.
- The favicon half of `ys_core_page_attachments()` → `ys_media_page_attachments()`. The Siteimprove half stays in ys_core, so both modules now implement the hook — that split is intentional.
- `images/favicons/` (4 fallback assets).

`ys_core_preprocess_image_widget()` stays put, as the ticket specifies — it serves the Site/Header Settings previews, which also stay.

**Three things worth a reviewer's attention:**

**1. The favicon URL path changes, and it broke something the ticket does not mention.** The fallback hrefs are hardcoded strings, so moving the assets changes the public URL to `.../modules/custom/ys_media/images/favicons/...`. A consumer audit turned up `web/themes/custom/ys_admin_theme/templates/content-edit/file-managed-file--favicon.html.twig` — a **theme** hardcoding a **module's** asset path in two places, to render the fallback preview on the Site Settings form. Nothing tests that template, so without the fix it would have become two silently broken images with a fully green suite. Fixed here, and verified in the real UI rather than by inspection. The `ys_media` README now records that the template is unguarded so the next person to move these files knows to check it by hand.

**2. A required service reference here would have broken `drush deploy` on every existing site.** `drush deploy` runs `updatedb` **before** `config:import`, and new modules are enabled by config import (hard constraint 8) — so there is exactly one bootstrap where the new `ys_core` code is on disk but `ys_media` is not yet enabled. I spiked this rather than assuming: a required `@ys_media.media_manager` argument makes the container fail to compile there —

```
In CheckExceptionOnInvalidReferenceBehaviorPass.php line 119:
  The service "ys_core.core_twig_extension" has a dependency on a
  non-existent service "..."
```

— which takes down the whole deploy, not just the feature. An `hook_update_N` **cannot** rescue it either: the hook runs after bootstrap, and bootstrap is what fails. So the argument is optional (`@?ys_media.media_manager`) and `CoreTwigExtension::__construct` takes `?YaleSitesMediaManager`, with a guard at the one call site. This is not a new invention — it is the same pattern `ys_ai` already uses for its optional `@?ys_beacon.authorization` reference, and the comment style follows it. For the same reason there is deliberately **no `ys_core_update_10021()`**: config import enables the module, and unlike `ys_core_update_10017` (Beacon) there is no config for config_ignore to strip, so an install hook would be cargo-cult.

The two settings forms deliberately keep a **hard** `$container->get('ys_media.media_manager')`. That asymmetry is intentional: a required argument in a *service definition* fails container compilation, but a hard `get()` inside a lazily-instantiated form only throws if someone loads that admin route — which cannot happen mid-deploy. Making those optional would push a NULL into the form and turn a loud, bounded failure into a silent one.

**3. Two path literals inside a Phase 0 test assertion were changed.** A strict reading of the acceptance criterion ("moving a test file between modules is allowed; changing assertions is not") forbids this, so flagging it rather than hoping it slides. It is defensible because that test file's own docblock anticipated it verbatim:

> The other fallback tests assert these hrefs as hardcoded strings, so they stay green even if the files move or are deleted -- which is what Phase 1 of #579 does when it relocates this class and the images/favicons/ directory into a new module.

The sibling guard `testGetFaviconsFallbackFilesExistOnDisk` **fails** unless the literals are repointed — that is exactly what it was written to do. No assertion was weakened or removed.

**Also of note:** `ys_media.info.yml` declares `imagemagick:imagemagick`, the first hard declaration of it anywhere in the profile. Not scope creep — the `favicon_16x16_ico` image style converts to `.ico`, which the GD toolkit cannot produce, so the favicon pipeline genuinely requires it. It was already enabled platform-wide (`core.extension.yml`) and pinned in the profile's `composer.json`; it was simply never declared by an owner.

**Tests.** The 16 characterization tests move with the code. Added: one Kernel test pinning the BC alias is a *real* alias and not a duplicate definition; one Unit test guarding the `core.extension.yml` registration (drop that line and the module never enables, the optional reference stays NULL forever, and favicons vanish with no error); and one Unit test in `CoreTwigExtensionTest` covering the NULL-media-manager path — the only genuinely new behavioural branch in the PR, and precisely what has to hold during the deploy window above.

**Not fixed here, needs its own phase.** `ys_media` reads `ys_core.site` while `ys_core` depends on `ys_media` — the module dependency points one way and the config dependency the other. That is required by hard constraint 1 and is documented in the module README, but it should get an explicit later phase (move `custom_favicon` to a `ys_media.settings` object with an update hook) rather than quietly becoming permanent.

This is Phase 1 of 6. **The issue stays open.**

### Functional testing steps:

- [ ] `lando composer code-sniff` exits 0, and `lando composer lint:php` exits 0
- [ ] Run the suites green (swap the module path in each):

  ```
  lando ssh -c "env SIMPLETEST_DB=mysql://pantheon:pantheon@database/pantheon?module=mysql \
    SIMPLETEST_BASE_URL=http://appserver \
    php /app/vendor/bin/phpunit -c /app/phpunit.xml \
    /app/web/profiles/custom/yalesites_profile/modules/custom/ys_media/tests --testdox"
  ```

  What I measured, each as its own run: `ys_media/tests` = 18 tests / 56 assertions. `ys_core/tests/src/Unit/` = 222 tests / 242 assertions / 1 pre-existing skip (baseline before this PR was 237; 16 moved to ys_media, 1 added here). `ys_core/tests/src/Kernel/` = 37 tests / 504 assertions, identical to baseline. The Kernel suite takes ~27 minutes.
- [ ] **The deploy gate — this is the important one.** Against a real pulled DB that predates the change: `npm run db:get && npm run confim`. Confirm it exits 0 and that the `updatedb` phase shows **no** `ServiceNotFoundException` / "non-existent service" error. A `site:install --existing-config` does **not** exercise this path and will not catch a regression here.
- [ ] After that deploy, confirm `ys_media` is enabled: `lando drush pm:list --status=enabled | grep ys_media`
- [ ] Front page still emits four favicon `<link>` tags in the head, and each resolves (not 404)
- [ ] **Fallback branch:** on a site with no custom favicon set, the four tags point at `/profiles/custom/yalesites_profile/modules/custom/ys_media/images/favicons/...` and all four load
- [ ] **The theme fix:** open `/admin/yalesites/settings` on a site with no custom favicon and confirm the Custom Favicon preview renders both images (the Yale "Y" and the simulated browser tab) rather than broken-image icons
- [ ] Upload a custom favicon in Site Settings, save, and confirm the head tags switch to the generated image-style derivatives; then remove it and confirm the fallbacks come back
- [ ] Upload an SVG site name image under Header Settings and confirm it still renders with the site name as its `<title>` (this is the `CoreTwigExtension` → media manager path)
- [ ] Confirm the BC alias survives: `lando drush php:eval 'var_dump(get_class(\Drupal::service("ys_core.media_manager")));'` returns `Drupal\ys_media\YaleSitesMediaManager`
- [ ] Upload a transparent PNG as an image and confirm transparency survives image-style processing (the relocated ImageMagick alter)

References yalesites-org/YaleSites-Internal#579


miketullo95 (approved):
@dblanken-yale Approving. This is a clean Phase 1 and two of the catches in it are the kind that would have hurt us quietly.

**The deploy window is the one I care about most.** Working out that `drush deploy` runs `updatedb` before `config:import`, and that this leaves exactly one bootstrap where the new ys_core code is on disk but ys_media is not enabled, is not something that turns up by reading the diff. Spiking it to get the actual container-compile error rather than reasoning about what might happen is the right instinct, and so is following the existing `ys_ai` to `ys_beacon` pattern instead of inventing something. The reasoning on why an update hook can't rescue it, since the hook runs after the bootstrap that's already failing, is the part I'd have gotten wrong.

The asymmetry between the optional service reference and the hard `$container->get()` in the two settings forms also makes sense to me as you've explained it. A loud failure on an admin route nobody can reach mid-deploy is better than a NULL quietly propagating into a form.

**The theme template catch is the other one.** A theme hardcoding a module's asset path in two places, with nothing testing that template, is exactly how a behavior-preserving refactor ships two broken images on the Site Settings form with a green suite. Good that you found it with a consumer audit and verified it in the real UI, and better that the README now records the template as unguarded so the next person to move these files knows to check it by hand.

## The AC exception: granted

Changing the two path literals in the Phase 0 assertion is fine and you should not have had to worry about it. The rule exists so nobody weakens a characterization test to make a refactor pass, and that is not what happened here: the test's own docblock predicted this exact move, `testGetFaviconsFallbackFilesExistOnDisk` fails unless the literals are repointed, and no assertion got weakened or removed. Flagging it rather than hoping it slid through is the behaviour I want, so thank you for that.

## One ask before merge

I can't verify the deploy gate myself. It needs a real pulled database and `npm run confim`, and you're right that `site:install --existing-config` doesn't exercise that path, so there's nothing I can click on a multidev that would catch a regression there. Can you point me at this multidev's build log showing a clean `updatedb` phase with no `ServiceNotFoundException`? That plus your local run is enough evidence for me.

I'll be testing the rest on the multidev: favicon tags resolving on the front page, the fallback branch pointing at the new `ys_media` path, the Site Settings favicon preview rendering both images rather than broken-image icons, a custom favicon upload and removal round trip, and the SVG site name image.

## Follow-ups, both mine to file

**The backwards config dependency is getting a real phase, not a README note.** You're right that `ys_media` reading `ys_core.site` while `ys_core` depends on `ys_media` shouldn't quietly become permanent, and I noticed Phases 2 through 6 as scoped don't cover it. I'm adding it to #579 as its own phase so it stays in the sequence instead of turning into orphaned cleanup.

**The config schema blocker still had no ticket.** The 2026-08-14 re-groom notes recommended spinning it off and nothing was ever filed. I checked. Filing it now, since it's the one open Phase 0 blocker, it's what stops the custom-favicon test branch from being written, and those config objects are unvalidated in production today regardless of this refactor.

Neither one gates this PR.

**Minor, no action needed:** worth being aware that `ys_core` now transitively hard-depends on `imagemagick` through `ys_media`. Your justification holds since the `favicon_16x16_ico` style genuinely can't run on GD and it was already enabled platform-wide, so this is just declaring something that was true and unowned. Just noting the coupling is new.

miketullo95:
Both follow-ups from my review are filed, neither gating this PR:

- **Config schema**: yalesites-org/YaleSites-Internal#1563. Scoped from the 2026-07-25 audit, assigned to you, `claude` labelled since the scope is fully enumerated.
- **The backwards config dependency**: added to #579 as **Phase 1b**, covering moving `custom_favicon` into a `ys_media.settings` object with an update hook. Sequenced after #1563 so the schema pattern is established first.

Phase 1b is the one place in this refactor where hard constraint 1 gets knowingly traded away rather than honoured, so I noted on the ticket that it needs an explicit update hook and a release-notes callout, not a quiet rename.

Still would like the multidev build log showing a clean `updatedb` phase before this merges.

dblanken-yale:
## Deploy gate verified against a real pulled DB

Ran the deploy gate ("the important one" in the testing steps) against a freshly pulled `yalesites-platform` DB, capturing config values **before** config import so the before/after could be diffed byte-for-byte. Summary: **the deploy passes and no settings are lost, in either direction.**

Worth noting up front that `npm run confim` is literally `lando drush deploy` (`scripts/local/confim.sh`), and `settings.php` sets `config_sync_directory` to the profile's `config/sync` — so the checklist gate and the production deploy path are the same command. The PR head was 0 commits behind `develop`, so this isolates the PR diff.

### Pre-state (before any import)

The pulled DB was a good test subject: `ys_core` enabled, **`ys_media` absent**, `imagemagick` already enabled, and a real custom favicon set — so this exercised the custom branch, not only the fallbacks.

```
custom_favicon: ['50']   ->  fid 50 = public://favicons/android-chrome-512x512.png
```

### Deploy result

```
[success] Finished performing updates.
[success] Config import start.
[success] The configuration was imported successfully.
[success] Cache rebuild start.
[success] Cache rebuild complete.
[success] Finished performing deploy hooks.
```

No `ServiceNotFoundException`, no "non-existent service", no class-not-found anywhere in the run. `ys_media` came out enabled, `ys_core` still enabled.

### Config and data preservation

Every value diffed against the pre-import snapshot:

| Check | Result |
| --- | --- |
| `ys_core.site` (holds `custom_favicon`) | identical |
| `ys_core.header_settings` | identical |
| `ys_core.social_links` | identical |
| File entity fid 50 | intact |
| 4 rendered favicon `<link>` tags | identical |
| BC alias | `ys_core.media_manager === ys_media.media_manager` returns `true` (real alias, same instance, not a duplicate definition) |

This is consistent with the design: `ys_media` ships no `config/install`, no `config/schema` and no `.install` file, so enabling it cannot create, overwrite or delete a config object. The only `config/sync` change is the one added line in `core.extension.yml`. `ys_core.site` and `ys_core.header_settings` keep their names, owner and schema, and the forms still read and write `ys_core.site`.

### Functional checks (all verified live, not by inspection)

- **Fallback branch** — with `custom_favicon` cleared, all 4 tags point at `.../ys_media/images/favicons/...` and every one returns 200. The old `ys_core` path 404s, which is precisely why the theme fix was needed.
- **The theme fix** — `/admin/yalesites/settings` renders both preview images from the `ys_media` path, with zero remaining `ys_core/images/favicons` references in the rendered page.
- **ImageMagick** — all four favicon image styles generate, including `favicon_16x16_ico` producing a real 1150-byte `.ico`. GD cannot do that, so the new `imagemagick:imagemagick` declaration is genuinely required rather than scope creep.
- **`CoreTwigExtension` -> media manager** — an uploaded SVG had its `<title>` correctly replaced with the live site name.
- **Hook relocation** — `ys_core_imagemagick_arguments_alter` gone, `ys_media_imagemagick_arguments_alter` present and registered; both modules implement `page_attachments`, as intended by the split.

All temporary test artifacts were removed and config re-verified identical afterwards.

### The optional service reference is load-bearing

Rather than take this on trust, I reproduced the deploy window directly (new `ys_core` code on disk, `ys_media` removed from active `core.extension`):

- With `@?ys_media.media_manager` (as in this PR): container compiles, site serves 200, favicons simply do not emit for that window. Bounded and cosmetic; no data touched.
- With the reference changed back to a required `@ys_media.media_manager`: the error in the PR description reproduces verbatim —

  ```
  In CheckExceptionOnInvalidReferenceBehaviorPass.php line 119:
    The service "ys_core.core_twig_extension" has a dependency on a
    non-existent service "ys_media.media_manager".
  ```

  and both `drush updatedb` and the full `drush deploy` then fail. Since bootstrap itself is what breaks, an update hook could not rescue it.

So the optional reference is necessary, and the reasoning for omitting `ys_core_update_10021()` holds.

### Rollback also verified

Testing the direction the checklist does not cover, since it is the one that could silently lose data. Deploying `develop` over a DB that already has `ys_media` enabled: config import cleanly removes the module, and `ys_core.site`, `ys_core.header_settings`, `ys_core.social_links`, fid 50 and all 4 favicon tags return **identical to the original pre-deploy state**, with the service reverting to `Drupal\ys_core\YaleSitesMediaManager`. No cascade into `ys_core`, because `develop`'s `ys_core.info.yml` carries no `ys_media` dependency. Reverting this PR is safe.

### Notes, none blocking

1. Each deploy direction logs one `[error]` line in the `updatedb` phase — "module marked as installed in core.extension but missing" (`ys_media` on rollback, `ys_servicenow` on the first run). The deploy still completes and config import reconciles it. The `ys_servicenow` one is pre-existing and unrelated to this PR: not on disk in this checkout, not in `develop`'s sync `core.extension.yml`, not touched by the diff.
2. Immediately after the code swap, an ad-hoc `drush php:eval` threw `Class "Drupal\ys_core\YaleSitesMediaManager" not found` from the stale cached container. `drush deploy` is unaffected because its update-mode bootstrap rebuilds the container, and web traffic continues serving from the cached valid container until the rebuild. This is inherent to relocating any class in any Drupal deploy, not something this PR introduces. Flagging only so nobody is surprised by it during release.

### Scope of this verification

Covered: the deploy gate, `ys_media` enablement, favicon head tags in both custom and fallback branches, the theme preview fix, the BC alias, the ImageMagick-dependent `.ico` style, the site name image path, and the rollback direction.

Not covered here: the PHPUnit suites and `composer code-sniff` / `lint:php`, and the transparent-PNG upload check. Those remain open on the checklist.

PR #1477, opened 2026-08-18
1542: Feature Request: Office Hours block (wraps Drupal Office Hours contrib module)

dblanken-yale:
### Description of work

Adds an **Office Hours block** that editors can place in every content section layout, wrapping the contrib [Office Hours](https://www.drupal.org/project/office_hours) module.

**No custom PHP was needed.** The contrib module supplies the field type, widget and table formatter, so this is config plus one preview icon. The issue proposed following the `ys_alert` pattern — that turned out to be the wrong shape (see "Why not `ys_alert`" below).

**The module**
- `drupal/office_hours` **1.29.0** added to the *profile* `composer.json`, pinned exactly, and enabled. Core-only dependencies, no new contrib pulled in.

**The block**
- `block_content` type `office_hours` with a multi-value `field_office_hours`.
- Field storage enables **exceptions** (holidays) and **all-day**, and uses a **12-hour** clock. Seasons are off for v1.
- Form display uses the *week with exceptions* widget: the Sun–Sat grid, per-day "Add time slot" / "Clear" / "Copy previous day", and an **Add exception** control for holidays.
- View display uses the table formatter with:
  - `show_closed: open` — **only the days the editor filled in are rendered**, no blank days (an explicit AC, met by config rather than code).
  - current status shown above the table ("Currently open!" / "Currently closed").
  - exception horizon raised from the contrib default of **7 days** to **180**, so a Thanksgiving closure is visible in advance rather than only during that week.

**Placement — all sections and regions**
- Layout Builder Browser entry under **Text and Content**, with a new preview icon.
- Added to the `layout_builder_restrictions_by_region` allow-lists for **70/30 (content *and* sidebar), 50/50, and 33/33/33** across the five view displays that restrict those categories.
- **Full width (`layout_onecol`) needed no entry.** `Custom block types` and `Inline blocks` are not in `restricted_categories` there, and per the plugin's logic (`EntityViewModeRestrictionByRegion.php` lines 132–140) a category with no allow-list key and no restriction permits everything — so adding a key there would have *newly restricted* the category. Worth knowing for the next block.
- Banner and page-meta layouts are deliberately excluded.

**Verified on the running site** — the block appears in all **eight** content regions across the four layouts, is correctly **absent** from the banner region, and renders in a 33% column:

| Region | Result |
|---|---|
| Full width (`layout_onecol`) | pass |
| 70% content | pass |
| 30% sidebar | pass |
| 50/50 primary + secondary | pass |
| 33/33/33 primary + secondary + tertiary | pass |
| Banner (should be excluded) | correctly excluded |

<details>
<summary><b>Evaluation notes and known follow-ups</b></summary>

#### Why not the `ys_alert` pattern
The issue description proposed following `ys_alert`. That is a singleton block placed sitewide in the atomic `header` region via Block Layout, and its category sits in `restricted_categories` for *every* layout and region — it is deliberately excluded from Layout Builder. The right precedent is the component-block pattern (accordion / text / quick_links): a `block_content` type + field + browser entry + per-region allow-lists, which is what this PR does.

#### What contrib gives us for free
Weekly Sun–Sat schedule, up to 2 time slots per day, closed days, holiday/exception dates (rendered as a separate "Exception hours" table with full dates), seasons, open/closed status, and a schema.org `openingHours` formatter.

#### Still to do (not in this PR)
1. **Theming — the main gap.** The table currently renders **contrib markup** with contrib's own zebra striping, not Yale design tokens. Needs a CLT/atomic component + Twig override. **Blocked on @atiddei's mockup**, which is the real critical path.
2. **Narrow-column behaviour.** With comments filled in, the 3-column table **overflows at 33%** — the Comment column clips mid-word and time ranges wrap ("9:00 am-5:00 / pm"). With comments empty it fits fine. This is the concrete thing the mockup needs to solve: stacked day/time, comment on its own line, or drop the comment column below a breakpoint.
3. **Accessibility.** Contrib renders through core's `stable9` table template as `<th colspan="3" class="visually-hidden">Day</th>` with **no `scope` attributes**, and the `colspan="3"` does not line up with the 3 body cells. The ticket's "proper table headers/scope attributes; WCAG 2.1 AA" AC needs the Twig override in (1). Flagging rather than silently passing.
4. **Style dials.** No `field_style_*` / padding options yet — deferred until the mockup says which dials apply.

#### Open questions
1. **The reusable-block question is already answered by the platform** — the inline-block form shows the standard "Reusable Block" toggle, so the editor chooses at creation. Suggest closing that AC.
2. Seasons (summer/winter hours) come free from contrib. In or out of v1? Currently **out**.
3. schema.org `openingHours` formatter — worth enabling for SEO? Currently **out**.
4. Is 180 days the right exception horizon?
5. Should this also be allowed on Event and Resource nodes? Currently it follows the existing per-display restrictions: Page, Post (50/50 only), Profile, and Resource (unrestricted, so allowed automatically).

</details>

### Functional testing steps:
- [ ] Pull the branch, run `lando composer install`, then `npm run confim` and `lando drush cr`.
- [ ] Edit any Page → **Edit Layout and Content** → add a section, click **Add block** → confirm **Office Hours** appears under **Text and Content** with its preview icon.
- [ ] Repeat in each layout and region: full width, 70% content, 30% sidebar, both 50/50 regions, all three 33/33/33 regions. It should be offered in all of them.
- [ ] Confirm it is **not** offered in the Banner section.
- [ ] Add the block, set hours for a few days, leave at least one day blank, and add a holiday via **Add exception**. Save.
- [ ] On the rendered page confirm: the filled days show a 12-hour range and the open/closed status shows above the table. *(Superseded by round 2: blank days now render as "Closed" rather than being omitted.)*
- [ ] Toggle **Reusable Block** on a second instance and confirm it is reusable across pages.
- [ ] Known and expected at this stage: the table is unstyled contrib markup pending the design mockup, and it overflows in a 33% column when comments are filled in. Please review those against the mockup rather than as blockers here.

#### Review round 2 - editor form changes (yalesites-org/yalesites-project#1477 review of 2026-08-18)

- [ ] Re-run `npm run confim` and `lando drush cr` after pulling, so the updated field description and formatter settings land.
- [ ] Open **Edit Layout and Content** -> add an **Office Hours** block. Confirm the Hours help text is separated from the "Hours" summary above it and now reads "Tick Closed to show a day as closed, or All day for a day that is open around the clock."
- [ ] Confirm the widget table has a **Closed** column between **All day** and **From**, and that each row's three action links are now behind a single **...** overflow menu.
- [ ] On a day: enter a From/To time and confirm **Closed** unticks. Tick **Closed** and confirm both times clear. Tick **All day** and confirm **Closed** unticks, greys out, and the From/To fields become visibly disabled. Untick **All day** and confirm everything re-enables.
- [ ] Open the **...** menu and use **Clear** (times clear, Closed re-ticks) and **Copy previous day** (times copy, Closed unticks). Confirm the menu closes after each.
- [ ] Tab to a **...** toggle and confirm it takes a visible focus ring and opens with Enter or Space.
- [ ] Repeat the above with Gin dark mode on: the disabled From/To fields should now read clearly darker than the editable ones.
- [ ] Untick **Closed** on a day and confirm the cursor jumps to that day's **From** field.
- [ ] Fill Mon-Fri, leave Sun and Sat closed, add an exception for a date marking it closed, and save. On the rendered page confirm every weekday now shows a row - filled days with hours, untouched days as **Closed** - and that the exception still appears under its own **Exception hours** heading.
- [ ] Reopen the block and confirm the **Closed** ticks match what you saved.
- [ ] Still known and expected: the rendered table is unstyled contrib markup pending April's mockup, and a near-term exception does not yet replace its weekday row (see the summary comment for the three options and why the contrib setting alone is not enough).

References yalesites-org/YaleSites-Internal#1542

### DIFF
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -86,6 +86,7 @@
     "drupal/multiple_fields_remove_button": "2.3",
     "drupal/multivalue_form_element": "1.0.0-beta8",
     "drupal/node_revision_delete": "2.1.1",
+    "drupal/office_hours": "1.29.0",
     "drupal/override_node_options": "2.9.0",
     "drupal/pantheon_advanced_page_cache": "2.3.4",
     "drupal/pantheon_secrets": "1.0.6",
--- a/web/profiles/custom/yalesites_profile/config/sync/block_content.type.office_hours.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/block_content.type.office_hours.yml
@@ -1,0 +1,33 @@
+uuid: 44ee820d-c13a-4b80-b935-fcd8fa518f7a
+langcode: en
+status: true
+dependencies:
+  module:
+    - entity_redirect
+third_party_settings:
+  entity_redirect:
+    redirect:
+      add:
+        active: 0
+        destination: default
+        url: ''
+        external: ''
+      edit:
+        active: 0
+        destination: default
+        url: ''
+        external: ''
+      delete:
+        active: 0
+        destination: default
+        url: ''
+        external: ''
+      anonymous:
+        active: 0
+        destination: default
+        url: ''
+        external: ''
+id: office_hours
+label: 'Office Hours'
+revision: true
+description: 'Display opening hours for a location, with a weekly schedule and holiday or exception dates.'
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.block_content.office_hours.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_form_display.block_content.office_hours.default.yml
@@ -1,0 +1,38 @@
+uuid: 9e0a721a-5a4c-4592-958e-7327738a2bb5
+langcode: en
+status: true
+dependencies:
+  config:
+    - block_content.type.office_hours
+    - field.field.block_content.office_hours.field_office_hours
+  module:
+    - hide_revision_field
+    - office_hours
+id: block_content.office_hours.default
+targetEntityType: block_content
+bundle: office_hours
+mode: default
+content:
+  field_office_hours:
+    type: office_hours_exceptions
+    weight: 1
+    region: content
+    settings:
+      collapsed: false
+      collapsed_exceptions: true
+    third_party_settings: {  }
+  revision_log:
+    type: hide_revision_field_log_widget
+    weight: 80
+    region: content
+    settings:
+      rows: 5
+      placeholder: ''
+      show: true
+      default: ''
+      permission_based: false
+      allow_user_settings: true
+      hide_revision: false
+    third_party_settings: {  }
+hidden:
+  info: true
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.block_content.office_hours.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.block_content.office_hours.default.yml
@@ -1,0 +1,46 @@
+uuid: 2e7a9621-5bb8-4195-9ecd-14f376654d18
+langcode: en
+status: true
+dependencies:
+  config:
+    - block_content.type.office_hours
+    - field.field.block_content.office_hours.field_office_hours
+  module:
+    - office_hours
+id: block_content.office_hours.default
+targetEntityType: block_content
+bundle: office_hours
+mode: default
+content:
+  field_office_hours:
+    type: office_hours_table
+    label: hidden
+    settings:
+      day_format: long
+      time_format: g
+      compress: false
+      grouped: false
+      show_empty: false
+      show_closed: open
+      closed_format: Closed
+      all_day_format: 'All day open'
+      separator:
+        days: '<br />'
+        grouped_days: ' - '
+        day_hours: ': '
+        hours_hours: '-'
+        more_hours: ', '
+      current_status:
+        position: before
+      exceptions:
+        restrict_exceptions_to_num_days: 180
+      timezone_field: ''
+      office_hours_first_day: ''
+      schema:
+        enabled: false
+    third_party_settings: {  }
+    weight: 1
+    region: content
+hidden:
+  info: true
+  search_api_excerpt: true
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.block_content.office_hours.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.block_content.office_hours.default.yml
@@ -21,7 +21,7 @@ content:
       compress: false
       grouped: false
       show_empty: false
-      show_closed: open
+      show_closed: all
       closed_format: Closed
       all_day_format: 'All day open'
       separator:
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.page.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.page.default.yml
@@ -136,6 +136,7 @@ third_party_settings:
               - pull_quote
               - text
               - video
+              - office_hours
             'Inline blocks':
               - 'inline_block:button_link'
               - 'inline_block:divider'
@@ -145,6 +146,7 @@ third_party_settings:
               - 'inline_block:pull_quote'
               - 'inline_block:text'
               - 'inline_block:video'
+              - 'inline_block:office_hours'
         ys_layout_two_column:
           content:
             'Custom block types':
@@ -165,6 +167,7 @@ third_party_settings:
               - text
               - video
               - wrapped_text_callout
+              - office_hours
             'Inline blocks':
               - 'inline_block:accordion'
               - 'inline_block:button_link'
@@ -183,6 +186,7 @@ third_party_settings:
               - 'inline_block:text'
               - 'inline_block:video'
               - 'inline_block:wrapped_text_callout'
+              - 'inline_block:office_hours'
             'YaleSites Layouts': {  }
           sidebar:
             'Custom block types':
@@ -194,6 +198,7 @@ third_party_settings:
               - text
               - video
               - webform
+              - office_hours
             'Inline blocks':
               - 'inline_block:button_link'
               - 'inline_block:divider'
@@ -203,6 +208,7 @@ third_party_settings:
               - 'inline_block:pull_quote'
               - 'inline_block:text'
               - 'inline_block:video'
+              - 'inline_block:office_hours'
         ys_layout_two_column_50_50:
           all_regions:
             'Custom block types':
@@ -217,6 +223,7 @@ third_party_settings:
               - text
               - video
               - webform
+              - office_hours
             'Inline blocks':
               - 'inline_block:accordion'
               - 'inline_block:button_link'
@@ -229,6 +236,7 @@ third_party_settings:
               - 'inline_block:tabs'
               - 'inline_block:text'
               - 'inline_block:video'
+              - 'inline_block:office_hours'
       denylisted_blocks: {  }
       restricted_categories:
         ys_layout_two_column:
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.page.single.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.page.single.yml
@@ -74,6 +74,7 @@ third_party_settings:
               - pull_quote
               - text
               - video
+              - office_hours
             'Inline blocks':
               - 'inline_block:button_link'
               - 'inline_block:divider'
@@ -83,6 +84,7 @@ third_party_settings:
               - 'inline_block:pull_quote'
               - 'inline_block:text'
               - 'inline_block:video'
+              - 'inline_block:office_hours'
         ys_layout_two_column:
           content:
             'Custom block types':
@@ -103,6 +105,7 @@ third_party_settings:
               - text
               - video
               - wrapped_text_callout
+              - office_hours
             'Inline blocks':
               - 'inline_block:accordion'
               - 'inline_block:button_link'
@@ -121,6 +124,7 @@ third_party_settings:
               - 'inline_block:text'
               - 'inline_block:video'
               - 'inline_block:wrapped_text_callout'
+              - 'inline_block:office_hours'
             'YaleSites Layouts': {  }
           sidebar:
             'Custom block types':
@@ -132,6 +136,7 @@ third_party_settings:
               - text
               - video
               - webform
+              - office_hours
             'Inline blocks':
               - 'inline_block:button_link'
               - 'inline_block:divider'
@@ -141,6 +146,7 @@ third_party_settings:
               - 'inline_block:pull_quote'
               - 'inline_block:text'
               - 'inline_block:video'
+              - 'inline_block:office_hours'
         ys_layout_two_column_50_50:
           all_regions:
             'Custom block types':
@@ -155,6 +161,7 @@ third_party_settings:
               - text
               - video
               - webform
+              - office_hours
             'Inline blocks':
               - 'inline_block:accordion'
               - 'inline_block:button_link'
@@ -167,6 +174,7 @@ third_party_settings:
               - 'inline_block:tabs'
               - 'inline_block:text'
               - 'inline_block:video'
+              - 'inline_block:office_hours'
       denylisted_blocks: {  }
       restricted_categories:
         ys_layout_two_column:
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.post.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.post.default.yml
@@ -124,6 +124,7 @@ third_party_settings:
               - text
               - video
               - webform
+              - office_hours
             'Inline blocks':
               - 'inline_block:accordion'
               - 'inline_block:button_link'
@@ -136,6 +137,7 @@ third_party_settings:
               - 'inline_block:tabs'
               - 'inline_block:text'
               - 'inline_block:video'
+              - 'inline_block:office_hours'
       restricted_categories:
         ys_layout_two_column_50_50:
           all_regions:
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.profile.default.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.profile.default.yml
@@ -173,6 +173,7 @@ third_party_settings:
               - text
               - video
               - wrapped_text_callout
+              - office_hours
             'Inline blocks':
               - 'inline_block:accordion'
               - 'inline_block:button_link'
@@ -190,6 +191,7 @@ third_party_settings:
               - 'inline_block:text'
               - 'inline_block:video'
               - 'inline_block:wrapped_text_callout'
+              - 'inline_block:office_hours'
             'YaleSites Layouts':
               - profile_contact_block
           sidebar:
@@ -204,6 +206,7 @@ third_party_settings:
               - text
               - video
               - webform
+              - office_hours
             'Inline blocks':
               - 'inline_block:accordion'
               - 'inline_block:button_link'
@@ -215,6 +218,7 @@ third_party_settings:
               - 'inline_block:tabs'
               - 'inline_block:text'
               - 'inline_block:video'
+              - 'inline_block:office_hours'
             'YaleSites Layouts':
               - profile_contact_block
       denylisted_blocks:
--- a/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.profile.single.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.entity_view_display.node.profile.single.yml
@@ -97,6 +97,7 @@ third_party_settings:
               - tabs
               - text
               - video
+              - office_hours
             'Inline blocks':
               - 'inline_block:accordion'
               - 'inline_block:button_link'
@@ -107,13 +108,16 @@ third_party_settings:
               - 'inline_block:tabs'
               - 'inline_block:text'
               - 'inline_block:video'
+              - 'inline_block:office_hours'
             'YaleSites Layouts':
               - profile_contact_block
           sidebar:
             'Custom block types':
               - text
+              - office_hours
             'Inline blocks':
               - 'inline_block:text'
+              - 'inline_block:office_hours'
             'YaleSites Layouts':
               - profile_contact_block
       denylisted_blocks: {  }
--- a/web/profiles/custom/yalesites_profile/config/sync/core.extension.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/core.extension.yml
@@ -114,6 +114,7 @@ module:
   mysql: 0
   node: 0
   node_revision_delete: 0
+  office_hours: 0
   options: 0
   override_node_options: 0
   page_cache: 0
--- a/web/profiles/custom/yalesites_profile/config/sync/field.field.block_content.office_hours.field_office_hours.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/field.field.block_content.office_hours.field_office_hours.yml
@@ -1,0 +1,21 @@
+uuid: 04f25f2e-8f93-41d6-89ee-8529655df59a
+langcode: en
+status: true
+dependencies:
+  config:
+    - block_content.type.office_hours
+    - field.storage.block_content.field_office_hours
+  module:
+    - office_hours
+id: block_content.office_hours.field_office_hours
+field_name: field_office_hours
+entity_type: block_content
+bundle: office_hours
+label: Hours
+description: 'Set the opening and closing time for each day. Leave a day blank to show it as closed. Use "Add exception" for holidays and one-off closures.'
+required: false
+translatable: true
+default_value: {  }
+default_value_callback: ''
+settings: {  }
+field_type: office_hours
--- a/web/profiles/custom/yalesites_profile/config/sync/field.field.block_content.office_hours.field_office_hours.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/field.field.block_content.office_hours.field_office_hours.yml
@@ -12,7 +12,7 @@ field_name: field_office_hours
 entity_type: block_content
 bundle: office_hours
 label: Hours
-description: 'Set the opening and closing time for each day. Leave a day blank to show it as closed. Use "Add exception" for holidays and one-off closures.'
+description: 'Set the opening and closing time for each day. Tick Closed to show a day as closed, or All day for a day that is open around the clock. Use "Add exception" for holidays and one-off closures.'
 required: false
 translatable: true
 default_value: {  }
--- a/web/profiles/custom/yalesites_profile/config/sync/field.storage.block_content.field_office_hours.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/field.storage.block_content.field_office_hours.yml
@@ -1,0 +1,32 @@
+uuid: 6a519de7-817a-4197-81ee-72a37a855f07
+langcode: en
+status: true
+dependencies:
+  module:
+    - block_content
+    - office_hours
+id: block_content.field_office_hours
+field_name: field_office_hours
+entity_type: block_content
+type: office_hours
+settings:
+  time_format: g
+  element_type: office_hours_datetime
+  increment: 30
+  valhrs: false
+  required_start: false
+  limit_start: ''
+  required_end: false
+  limit_end: ''
+  all_day: true
+  exceptions: true
+  seasons: false
+  comment: 1
+  cardinality_per_day: 2
+module: office_hours
+locked: false
+cardinality: -1
+translatable: true
+indexes: {  }
+persist_with_no_fields: false
+custom_storage: false
--- a/web/profiles/custom/yalesites_profile/config/sync/layout_builder_browser.layout_builder_browser_block.office_hours.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/layout_builder_browser.layout_builder_browser_block.office_hours.yml
@@ -1,0 +1,11 @@
+uuid: 7755b16e-6721-4378-b3f4-1e6307a70283
+langcode: en
+status: true
+dependencies: {  }
+id: office_hours
+block_id: 'inline_block:office_hours'
+category: text_and_content
+label: 'Office Hours'
+image_path: /profiles/custom/yalesites_profile/modules/custom/ys_core/images/preview-icons/office-hours.svg
+image_alt: ''
+weight: -10
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/images/preview-icons/office-hours.svg
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/images/preview-icons/office-hours.svg
@@ -1,0 +1,19 @@
+<svg width="501" height="261" viewBox="0 0 501 261" fill="none" xmlns="http://www.w3.org/2000/svg">
+<rect width="501" height="261" fill="white"/>
+<circle cx="70" cy="60" r="18" stroke="#222222" stroke-width="3"/>
+<path d="M70 50V60L77 65" stroke="#222222" stroke-width="3" stroke-linecap="round" stroke-linejoin="round"/>
+<rect x="102" y="50" width="120" height="8" rx="4" fill="#222222"/>
+<rect x="102" y="66" width="76" height="6" rx="3" fill="#8C8C8C"/>
+<line x1="36" y1="100" x2="465" y2="100" stroke="#222222" stroke-width="2"/>
+<rect x="36" y="114" width="86" height="7" rx="3.5" fill="#222222"/>
+<rect x="250" y="114" width="104" height="7" rx="3.5" fill="#8C8C8C"/>
+<line x1="36" y1="136" x2="465" y2="136" stroke="#D8D8D8" stroke-width="2"/>
+<rect x="36" y="150" width="96" height="7" rx="3.5" fill="#222222"/>
+<rect x="250" y="150" width="104" height="7" rx="3.5" fill="#8C8C8C"/>
+<line x1="36" y1="172" x2="465" y2="172" stroke="#D8D8D8" stroke-width="2"/>
+<rect x="36" y="186" width="78" height="7" rx="3.5" fill="#222222"/>
+<rect x="250" y="186" width="72" height="7" rx="3.5" fill="#8C8C8C"/>
+<line x1="36" y1="208" x2="465" y2="208" stroke="#D8D8D8" stroke-width="2"/>
+<rect x="36" y="222" width="110" height="7" rx="3.5" fill="#222222"/>
+<rect x="250" y="222" width="88" height="7" rx="3.5" fill="#8C8C8C"/>
+</svg>
--- a/web/themes/custom/ys_admin_theme/css/admin-ui.css
+++ b/web/themes/custom/ys_admin_theme/css/admin-ui.css
@@ -1,3 +1,4 @@
+@charset "UTF-8";
 :root {
   --drupal-displace-offset-top: 35rem;
   --drupal-displace-offset-bottom: 35rem;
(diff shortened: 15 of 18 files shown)


dblanken-yale:
Setting to "don't merge" until April can take a look and make recommendations.

miketullo95 (changes requested):
@dblanken-yale Really solid evaluation write-up, and the call to go config-only instead of building `ys_officehours` makes sense to me. The reasoning on why `ys_alert` was the wrong precedent is convincing, and the `layout_onecol` finding about allow-list keys newly restricting a category is worth keeping somewhere durable for the next block we add.

Requesting changes on the editor form experience. None of this is about the front-end rendering, which I agree waits on April's mockup. These are all things in the block edit form that I'd like sorted before this goes further, since they're independent of how the table ends up looking.

## Blocking

**1. "Clear the day to represent closed" doesn't actually show the day as closed.**

The field help text in `config/sync/field.field.block_content.office_hours.field_office_hours.yml:15` says "Leave a day blank to show it as closed." But `show_closed: open` in `config/sync/core.entity_view_display.block_content.office_hours.default.yml:24` drops blank days from the output entirely, so clearing a day makes it vanish rather than render as closed. Those two contradict each other today.

What I'd like instead of just fixing the wording: give each day an explicit **Closed** control next to All day, rather than making "empty" carry that meaning. A day can't be both all-day and closed, so a toggle between the two states reads better than two independent checkboxes. Being explicit also means an editor can tell the difference between "closed on Sunday" and "I haven't filled in Sunday yet," which right now look identical.

If contrib can't express that without a widget override, say so and we can talk through what's reasonable. But the current state where the help text promises one behavior and the config does the opposite needs to change either way.

**2. Disabled all-day fields don't read as disabled.**

When All day is checked in the widget (`office_hours_exceptions`, set in `config/sync/core.entity_form_display.block_content.office_hours.default.yml:17`), the from/to fields lose their clock icon but otherwise look the same as an editable field. Losing an icon isn't a disabled affordance. Needs the standard Claro disabled treatment, greyed background and all, so it's obvious the fields are inert and not just empty.

**3. Collapse the per-row operations into a menu.**

Each day row carries three separate action controls. That's a lot of visual weight across seven rows, and it competes with the fields that actually matter. I'd like these behind a single overflow menu, along the lines of the three-dot menu Drupal uses on paragraph rows. Same actions, less noise.

**4. Add top padding to the field help text.**

The description renders in `<div class="claro-details__description">` with no separation from the element above it. Small thing, but it currently reads as crowded against the widget.

**5. Exceptions don't override the regular weekly hours, and I expected them to.**

I added an exception for this coming Thursday marking it closed, and the regular Thursday hours still showed in the weekly table. Your PR body describes exceptions rendering as a separate "Exception hours" table, which matches what I saw, so I don't think this is a bug so much as contrib's default shape.

My expectation as an editor is that an exception inside the display window replaces the normal day it falls on, since that's the whole point of setting one. Showing "Thursday 9-5" and "Thursday Nov 27: Closed" in two different tables puts the work of reconciling them on the visitor.

Can you confirm what contrib actually supports here, whether that's a formatter setting or needs an override? Once we know the constraint, April and I can settle what the right display is. Flagging it now so it gets into the mockup conversation rather than surfacing after.

## Not blocking, just answering your open questions

- **Reusable block:** agreed, the platform already answers it with the standard toggle on the inline-block form. Let's close that AC.
- **Seasons:** keep out for v1. Nobody has asked for summer/winter hours yet and it's another dimension of complexity in a widget that already has plenty.
- **schema.org `openingHours`:** I'd like this eventually, but not in this pass. Worth a follow-up ticket once the markup settles, since the formatter output and the theming will interact.
- **180 day exception horizon:** fine by me. Far enough out to cover a fall semester's holidays without the table filling up with things nobody needs yet.
- **Event and Resource nodes:** leave it as-is following the existing per-display restrictions. No reason to special-case it.

## Housekeeping

The issue AC still says to build a custom `ys_officehours` module following the `ys_alert` pattern. Your config-only approach is the better call, so I'll update the AC on #1542 to match rather than leave it looking unmet.

cc @atiddei on items 1, 2 and 5, since those touch what the editor form and the rendered table need to communicate.

dblanken-yale:
@miketullo95 All five blocking items are addressed. Two of them landed somewhere other than where you'd expect, so those are spelled out first.

Everything here is still config plus admin-theme assets — no new module, no custom PHP, no contrib fork. The editor-form work is independent of the front-end theming, so this does not touch what's waiting on @atiddei's mockup.

---

## 1. "Clear the day to represent closed" — fixed, and here's the constraint behind it

**The contradiction is gone.** The help text now reads "Tick Closed to show a day as closed, or All day for a day that is open around the clock", and `show_closed` moved from `open` to `all` so a day with no hours renders as **Closed** instead of vanishing.

**There is now an explicit Closed control**, a checkbox in its own column next to All day, mutually exclusive with it. Ticking Closed clears that day's times; entering a time unticks it; ticking All day unticks and greys out Closed, and vice versa.

**But it cannot be a genuine third state, and you should know that before signing off.** The contrib field stores four columns per slot — `day | starthours | endhours | comment` (`OfficeHoursItemBase::schema()`). "All day" isn't a column either; it's encoded as `starthours = endhours = 0`. So a weekday has exactly two representable states: *has hours* or *empty*. Recording "closed" separately from "not filled in yet" would mean forking the contrib field type and owning that data model — including migrations every time contrib touches it. That's the "say so and we can talk through what's reasonable" case you flagged.

What I built instead is honest about that: the Closed checkbox **mirrors** the stored state rather than inventing one. It ticks itself for any day with no hours, and what it shows is exactly what the front end will render.

**The one thing I need you to confirm:** this makes the older AC "Front-end table only renders rows for fields the editor filled in, no blank days" impossible to keep at the same time. With contrib, an untouched day and a deliberately-closed day *are the same stored value*, so it's one setting for all seven days:

- **`show_closed: all`** (what's in the PR now) — every weekday renders; untouched days show **Closed**. The help text and the Closed control are true. Blank days are no longer omitted.
- **`show_closed: open`** (what was there before) — only filled days render. No blank days, but a day you deliberately mark Closed disappears, so the Closed control would lie.

I went with `all` because the Closed control is the newer AC and the one you argued for. It's a one-line revert if you'd rather keep "no blank days". Worth knowing: a *completely* empty Office Hours block still renders nothing at all — you only get the seven Closed rows once at least one day or exception is filled in.

## 2. Disabled all-day fields — partly your call, partly measured otherwise

I couldn't reproduce this in light mode, and I'd rather show you the numbers than quietly "fix" something that isn't broken. With All day ticked, the From/To fields already get the full Gin/Claro disabled treatment: background `#eaeaea` vs `#ffffff`, text `#767676` vs `#222330`, border `#c2c2c2`, and `cursor: not-allowed`. The clock icon disappearing is the browser, but it isn't the only cue.

**Where you're right is dark mode**, and there it's a real failure: Gin's `--gin-color-disabled-bg` is `#47474c` against an input background of `#3b3b3f` — about a 4% difference, which genuinely reads as editable. Fixed: within this widget, disabled inputs now sit on `--gin-bg-layer` (`#2a2a2d`), so they're clearly recessed. Scoped to the office hours field rather than applied platform-wide, since this is the first widget we ship that disables inputs dynamically — happy to lift it globally in a follow-up if you want.

If you were seeing this in light mode, tell me what you were looking at and I'll dig further.

## 3. Per-row operations collapsed into a menu — done

The three links are now behind a single **⋯** toggle per row. I used a native `<details>` disclosure rather than Drupal's dropbutton: a dropbutton always surfaces its first action next to the arrow, and contrib's own JS shows and hides the "Add time slot" link dynamically, which would have left a blank primary slot. The disclosure keeps all three actions hidden until you ask for them, which is what you described, and it's keyboard-operable with a real accessible name ("Actions for Sunday"). The menu closes itself after you pick an action.

## 4. Field help text spacing — done

The description now has top spacing separating it from the Hours summary above it.

## 5. Exceptions vs weekly hours — confirmed, and the answer is "needs an override"

You were right that this is contrib's default shape rather than a bug. Contrib does have a setting for what you expected — `exceptions.replace_exceptions` — and I turned it on, measured it, and then turned it back off. Here is why, because the measurements are the useful part.

**What `replace_exceptions: true` actually does**, tested against the real view display with hours Mon-Fri 9-5 and a closed exception on the coming Thursday:

| | `false` (what's shipped) | `true` |
|---|---|---|
| weekly row for the exception's day | `Thursday: 9:00 am-5:00 pm` | `Thursday: Closed` |
| the exception itself | own table under an **Exception hours** heading | an unlabelled dated row appended to the weekday table |
| a second exception 45 days out | listed under **Exception hours** | weekday row **unchanged**, still listed as a dated row |

So it fixes the contradiction you hit — but it costs two things:

1. **The "Exception hours" heading disappears.** With the flag on, contrib never emits the caption row, so it never starts a second table. Every exception inside the 180-day horizon becomes an unlabelled dated row trailing the weekday list, and a near-term one renders twice (once as `Thursday: Closed`, once as `Nov 27: Closed`). Contrib's own default horizon with this flag is **7 days**, not 180 — the flag and the 180 days we settled on don't really go together.
2. **The replacement only reaches about two weeks out.** Contrib resolves it against a rolling this-week/next-week calendar (`OfficeHoursItemListFormatter::addOfficeHours()`). Verified above: a 45-day-out exception doesn't replace anything.

**So the honest answer to your question is: a setting gets you part of it, and the display you described needs an override.** I left the setting off rather than trade a labelled exception list for a long unlabelled tail — that's a display call, and you said you and @atiddei wanted to settle it once we knew the constraint. Three options for that conversation:

- **as shipped** — separate, labelled Exception hours table; the weekly row can still contradict a near-term closure.
- **`replace_exceptions: true`** — weekly row correct inside ~2 weeks, but the heading goes, near-term exceptions double up, and the 180-day horizon wants dropping to ~7. One-line change.
- **a Twig / formatter override** — the only way to get "one row per day, the exception wins, across the whole 180-day horizon". That belongs with the theming work rather than here.

---

### Also in this push

- Found and fixed an accessibility gap in my own new markup: the **⋯** toggle had no visible focus indicator. It now uses Gin's own focus ring on `:focus-visible`.
- Contrib's `colspan`/`scope` problems on the *rendered* table are untouched and still depend on the Twig override in the mockup work — already captured on yalesites-org/YaleSites-Internal#1542.

### Where the code lives

- `web/profiles/custom/yalesites_profile/config/sync/` — field description, `show_closed`, `replace_exceptions`.
- `web/themes/custom/ys_admin_theme/js/office-hours-widget.js` — the Closed column and the operations menu, as a Drupal behavior. The checkbox is deliberately unnamed so it is never submitted; it only mirrors and controls the real inputs.
- `web/themes/custom/ys_admin_theme/scss/admin-ui.scss` (+ built CSS) — spacing, the new controls, the focus ring, the dark-mode disabled contrast.

`lando composer code-sniff` passes, stylelint and eslint are clean on the changed files, and the whole flow was driven end to end on a local site — add a block, set hours, clear a day, copy a day, add an exception, save, reopen, confirm the Closed ticks match what was stored. Testing steps are appended to the PR description.

Back to you as `needs review`. Two things are waiting on you rather than on me, and both are one-line changes once decided: the `show_closed` trade-off in item 1, and which of the three exception displays in item 5 you and @atiddei want.

@dblanken-yale note: your "don't merge until April can take a look" hold still stands — none of this touches the rendered table, which is what's waiting on the mockup.

miketullo95:
@dblanken-yale I confirmed the round 2 changes are on the branch (b34cf2c7) and deployed to the multidev. The new `office-hours-widget.js` and the `office-hours-*` CSS classes are live on pr-1477, and the reworded Hours help text is showing.

The issue is the editor form still renders the old way there: no **Closed** column, and each day's three action links are not collapsed into the `...` menu (screenshot below). Since both of those are injected client-side by the new Drupal behavior, it looks like the multidev is serving a stale aggregated JS bundle.

Can you run `drush cr` on the pr-1477 multidev (and clear the Pantheon CDN cache) so the form picks up the new JS? Once the Closed column and the `...` menu show up I will finish the round 2 review.

dblanken-yale:
## Fix: the Closed column and operations menu never loaded in the Layout Builder modal

The round-2 editor work was landing on full-page admin forms but **not** in the
"Configure block" modal, which is the path editors actually use. Reported against
the pr-1477 multidev; reproduced and fixed.

### Root cause

The widget JS was attached through the `ys_admin_theme/global` library. A theme's
`libraries:` entries are attached by `system_page_attachments()`
(`core/modules/system/system.module:599`), which is a `hook_page_attachments()`
implementation - it runs **only for a full HTML page render**.

Layout Builder renders in the front-end theme and loads the block form over AJAX,
so no page attachment ever happens there:

| Page | Active theme | Our behavior registered | Widget JS loaded |
|---|---|---|---|
| `/node/N/layout` (Layout Builder) | `atomic` | no | no |
| `/block/add/office_hours` | `ys_admin_theme` | yes | yes |

That is why it looked fine on the standalone block form and on a direct
`/layout_builder/add/block/...` URL - both render a full page in the admin theme -
while the modal, which is what editors see, got nothing.

### The fix

Move the widget JS and its styles out of `global` into a dedicated
`ys_admin_theme/office_hours_widget` library, and attach it with `libraries-extend`
against `office_hours/office_hours_widget`, so our assets load wherever the contrib
widget itself renders - full page or AJAX dialog alike.

The office-hours SCSS moved out of `admin-ui.scss` into its own
`scss/office-hours-widget.scss` so it travels with the JS.

### Verification

A/B on the modal's AJAX response (`?_wrapper_format=drupal_modal`):

| | `office-hours-widget.js` | `office-hours-widget.css` |
|---|---|---|
| with `libraries-extend` | yes | yes |
| without | no | no |

Contrib's own `office_hours.js` is present in both, confirming the widget library
is attached and only ours was missing. The full-page block form still renders the
Closed column and the overflow menus (no regression). `composer lint:php`,
`composer code-sniff` and `npm run lint:styles` all clean for the touched files.

### Separate pre-existing bug, not addressed here

`web/themes/contrib/gin/dist/js/description_toggle.js:9` calls
`elem.closest('.help-icon__description-container:not(.form-type--textarea)')
.querySelectorAll(...)` with no null guard. On any admin form with a textarea
description - the **Revision log message** field, whose container carries
`form-type--textarea` and is therefore excluded by the `:not()` - `closest()`
returns `null` and the behavior throws. Core catches it so nothing else breaks,
but Gin's description toggles silently stop attaching partway down such forms.
Worth its own ticket and a `composer-patches` entry rather than a drive-by here.

### Testing steps for this change

- [ ] Pull the branch, `lando composer install`, `npm run confim`, `lando drush cr`.
- [ ] Edit a Page -> **Edit Layout and Content** -> add an **Office Hours** block so
      the **Configure block** modal opens.
- [ ] In the modal, confirm the widget has a **Closed** column between **All day**
      and **From**, and that each row's actions sit behind a single **...** menu.
- [ ] Confirm the same still holds on the standalone block form
      (`/admin/content/block` -> add an Office Hours block).

dblanken-yale:
1477-proof-edit-sidebar
1477-proof-add-modal

dblanken-yale:
## Correction: we can have both behaviours - the premise blocking it is wrong

The open question here was framed as a trade-off: either `show_closed: all` (every weekday renders, unfilled days show "Closed", so the Closed control is truthful) or `show_closed: open` (only filled days render, but a deliberately-closed day vanishes). **That trade-off does not exist.**

**The data model distinguishes the two cases.** `OfficeHoursItem::isValueEmpty()` treats a weekday with no hours but a non-empty `comment` as *not* empty, so it persists as a real row - while an untouched day is never stored at all. The module's own kernel test asserts exactly this pair:

```php
// Weekday without hours, with comment.
$this->assertEquals('3/0////An empty weekday with comment', $test_value);
// Weekday without hours, without comment. This is not stored.
$this->assertEquals(NULL, $test_value);
```

And `keepOpenDays()` filters on `!empty($info['items'])` - whether a row exists, not whether it has hours. So under `show_closed: open`, a comment-carrying closed day renders and an unfilled day does not.

**Where the wrong premise entered.** `ys_admin_theme/js/office-hours-widget.js` states it as fact in its docblock:

> "there is no third column to record 'closed' separately from 'not filled in yet'"

There is - `comment` - and this field has `comment: 1` enabled in `field.storage.block_content.field_office_hours.yml`. `addClosedControl()` then makes the tick deliberately non-submitting ("The control is not submitted: a closed day is stored as a day with no hours"), which is precisely what collapses the two states. My earlier summary comment on this PR repeated the same premise; it was wrong.

**What this changes.** The Closed control needs to store a value instead of submitting nothing, and `show_closed` moves to `open`. Both acceptance criteria on yalesites-org/YaleSites-Internal#1542 then hold simultaneously, so neither has to be struck - that ticket has been updated to say so.

Two routes, not yet chosen:

1. **Comment column** - JS writes `Closed` into that day's comment on tick (guarding any editor-authored note), and one Twig condition renders `comments` as the hours text when `formatted_slots` is empty. No custom PHP, and the CLT table component is already a planned follow-up, so the Twig line lands there anyway.
2. **`OfficeHoursEvents::POST_FORMAT` subscriber in `ys_core`** (~25 lines) - keep `show_closed: all`, drop never-filled rows, and move `Closed` from `comments` into `formatted_slots`. Keeps the closed state out of the editor-facing comment field, at the cost of code to maintain.

Setting `needs work`: regardless of route, the docblock claim is false and the Closed control is non-functional.

Not yet verified live - the module is not installed on `develop`, so the above is read from pinned `8.x-1.29` source plus its kernel test. Worth one `drush sqlq` against `block_content__field_office_hours` on this branch to confirm the round trip when the work is picked up.

dblanken-yale:
## Fixed: the Closed control now stores the day it marks closed

Addressed the `needs work` finding ([comment](https://github.com/yalesites-org/yalesites-project/pull/1477#issuecomment-5637961590)). The premise correction in it was right, and I verified every claim against the pinned `8.x-1.29` source before acting on it rather than taking it on faith:

- `OfficeHoursItemBase::schema()` — four columns per slot, and the fourth is `comment`.
- `OfficeHoursItem::isValueEmpty()` — start **and** end empty **and** `empty($value['comment'])` is what makes a slot empty. So a hours-less day with a comment persists; an untouched day is dropped by `filterEmptyItems()`.
- `keepOpenDays()` — filters on `!empty($info['items'])`, i.e. row existence, not on whether the row has hours.

So under `show_closed: open` a deliberately-closed day renders and an untouched one does not. **Both acceptance criteria on yalesites-org/YaleSites-Internal#1542 hold at once — neither has to be struck.**

### Two questions for you before this is signed off

**1. Route confirmation.** You listed two routes and left it unchosen. I went with **route 1 (comment column, no custom PHP)**, because route 2 is not actually an alternative to it — `comment` is the only free column, so the `POST_FORMAT` subscriber would *also* need the comment as its stored marker. Route 2 is route 1 plus ~25 lines that relocate the text. Your own note ("the CLT table component is already a planned follow-up, so the Twig line lands there anyway") pointed the same way. Nothing here forecloses route 2 later.

**2. Known interim limitation — the closed text lands in the Comment column, not the Hours column.** A closed day currently renders:

| | | |
|---|---|---|
| Sunday: | | Closed |
| Monday: | 9:00 am-5:00 pm | |

The Hours cell is empty and "Closed" sits in the Comment cell. Route 1's other half — "one Twig condition renders `comments` as the hours text when `formatted_slots` is empty" — **cannot land in this repo today**: the contrib table formatter builds its rows as a `#type => table` render array, and `templates/office-hours-table.html.twig` is only `<div class="office-hours">{{ table }}</div>`. There is no per-cell Twig seam. That condition belongs in the CLT/atomic table component, which #1542 has blocked on @atiddei's mockup. **Confirm that is an acceptable interim, or say the word and I'll do route 2's subscriber instead so the Hours cell reads "Closed" now.**

### What changed

Three files, no new module, no custom PHP, no contrib fork.

**`js/office-hours-widget.js`** — the docblock's false claim is gone and replaced with the actual four-column model and the two contrib methods that make it work. The control writes the day's comment instead of submitting nothing.

Keeping the tick *honest about storage* needed three guards beyond the basic fix — a checkbox that misreports storage is what got this round a `needs work`, so these are the substance, not polish:

- **Reads span every slot row, writes target the first.** `cardinality_per_day: 2` and contrib gives every slot its own comment box, so a note typed into a continuation row stores that day too and has to read as closed.
- **Entering hours retracts our label.** Without this, an editor who ticks Closed and then just types into the enabled From/To boxes — without unticking — leaves `comment = "Closed"` behind, and the front end renders `| Monday | 9:00am-5:00pm | Closed |`. Reproduced before fixing.
- **A day closed by the editor's own note reports closed but disables the tick.** That day genuinely is closed, but retracting it would mean deleting their note, so it shows the state rather than silently re-ticking itself and fighting them.

**`core.entity_view_display...office_hours.default.yml`** — `show_closed: all` -> `open`.

**`field.field...field_office_hours.yml`** — help text gains "A day you leave blank is not shown at all", which `show_closed: open` makes true and an editor has to know, since that is precisely why they must tick Closed.

### Verification

Measured, not assumed. There is no JS test harness in this repo, so this is a before/after reproduction on Lando plus the `drush sqlq` round-trip you asked for.

**Before** — Closed checkbox had **no `name` attribute** (never submitted), and on a fresh form **every untouched day was already ticked**. Ticking Sunday Closed + Monday 9-5 and saving stored **one row**: `day=1, 900, 1700`. Sunday stored nothing.

**After** — same scenario:

| delta | day | start | end | comment |
|---|---|---|---|---|
| 0 | 0 (Sun) | NULL | NULL | `Closed` |
| 1 | 1 (Mon) | 900 | 1700 | |
| 2 | 3 (Wed) | NULL | NULL | `Closed for renovation` |

Tuesday, deliberately untouched, is **absent**. Re-opening the saved block re-ticks exactly Sunday and Wednesday from storage.

Rendered on a real page: Sunday and Wednesday present as closed, Monday with its hours, Tue/Thu/Fri/Sat omitted.

**Behavioural checks, all passing:** untouched day not pre-ticked; tick writes the label; untick retracts only ours; an editor's own note survives untick; All day retracts the label and disables Closed; entering hours retracts the label and unticks; hours in a *continuation* row also count as open; a foreign note ticks-and-disables and releases once cleared.

**Regression on the Layout Builder path you flagged in round 2** — still fixed. The off-canvas AJAX response carries both `office-hours-widget.js` and `office-hours-widget.css` next to contrib's `office_hours.js`, and I drove it rather than just reading the asset list: the Closed column and the `...` overflow menus render there, ticking writes `Closed`, and saving from that path stores the identical three rows.

Scoping confirmed: the exceptions table gets no Closed column, with and without an exception row present.

**Gate:** `composer code-sniff` exit 0, `composer lint:php` exit 0, `eslint` + `prettier --check` clean on the changed JS. `npm run confex` surfaced two unrelated local drifts (`inline_message` form-display ordering, a YAML quoting change in `ys_core.dashboard_settings.yml`) — both reverted, not this work-unit's.

Leaving `don't merge` in place, since the front-end theming still waits on the mockup.

### Not addressed here

`web/themes/contrib/gin/dist/js/description_toggle.js:9` still has the unguarded `closest()` noted earlier. Still worth its own ticket and a `composer-patches` entry rather than a drive-by.

dblanken-yale:
We're going to wait for design before addressing this now.

dblanken-yale:
## Acknowledged: parked for design, route 2 not implemented

Answering [the decision above](https://github.com/yalesites-org/yalesites-project/pull/1477#issuecomment-5639785171) — "We're going to wait for design before addressing this now."

**No code changed in this pass, and nothing was pushed.** The branch is exactly where the last round left it (`54a3c94`), and the PR is mergeable against `develop`.

### What that decision settles

The open question was the second half of route 1 — the one Twig condition that renders `comments` as the hours text when `formatted_slots` is empty, so a deliberately-closed day reads `| Sunday | Closed | |` instead of putting "Closed" in the Comment cell. That condition has no seam in this repo: contrib's table formatter builds its rows as a `#type => table` render array and `templates/office-hours-table.html.twig` is only `<div class="office-hours">{{ table }}</div>`. It belongs to the CLT/atomic table component.

The alternative on the table was route 2 — a `OfficeHoursEvents::POST_FORMAT` subscriber in `ys_core` (~25 lines) that would move `Closed` out of `comments` and into `formatted_slots` now, at the cost of custom PHP to maintain against a contrib event.

**Route 2 is not being built.** The interim stands: the day is stored correctly and renders, with "Closed" landing in the Comment cell until the table component catches up. That keeps the cell placement a theming decision made once, in the component, alongside @atiddei's mockup — rather than code we would write now and then unwind.

### What still holds from the last round

Unchanged and already verified on Lando — the storage half of route 1 is done and is not waiting on design:

- The Closed tick writes the day's `comment`, so a closed day persists as a real row while an untouched day stores nothing. Round-tripped via `drush sqlq`.
- `show_closed: open`, so only days the editor filled in render — both acceptance criteria on yalesites-org/YaleSites-Internal#1542 hold at once.
- Untick retracts only our own label; an editor's own note survives; entering hours retracts the label; continuation rows count as open.
- The Layout Builder off-canvas path carries the widget JS/CSS and stores identically.
- Gate: `composer code-sniff` exit 0, `composer lint:php` exit 0, eslint + prettier clean.

### State of the PR

- `don't merge` **stays** — the front-end theming still waits on the mockup.
- `question` **removed** — the question it was tracking has been answered.
- @atiddei is still the requested reviewer that this is blocked on. The mockup is what unblocks the remaining work; nothing further is needed from @miketullo95 or @dblanken-yale to proceed.

Blocking review items 1-4 from the round-1 review remain addressed. Item 5 (exceptions overriding the weekly row) is the part folded into the mockup conversation and is deliberately still open.

dblanken-yale:
## Code-review remediation — asset ownership and a data-loss bug in the Closed control

**TL;DR:** Four findings from a code review of this PR. Two are corrections to how the editor assets are wired (they were owned by the admin theme but attached by a module, which fails silently); one fixes a real bug where ticking hours on a day **deleted an editor's own note** if they had typed the word "Closed"; one makes the `core/once` dependency explicit in code. **Nothing here touches the front-end rendering that's parked for design** — all four are editor-form concerns that change no rendered output. Pushed as `99be46d0` and `cc36e8de`.

One thing needs your decision, at the bottom: half of the "Closed" finding is blocked by the same display-layer work you parked on 2026-09-11, so I deliberately did not do it.

---

### 1. The editor assets were owned by the theme but attached by a module

Your 2026-08-27 fix moved the JS off `ys_admin_theme/global` and onto a dedicated library so it would reach the Layout Builder modal. That was right, and this continues the same trajectory one step: the library was still *declared* in `ys_admin_theme` while being *attached* from `ys_core`.

That combination fails quietly. A module cannot declare a dependency on a theme, so nothing tied the string `'ys_admin_theme/office_hours_widget'` to the theme's `libraries.yml` — and core resolves a missing library by deleting it, not by complaining:

```php
// core/lib/Drupal/Core/Asset/LibraryDiscovery.php:57
public function getLibraryByName($extension, $name) {
  $libraries = $this->getLibrariesByExtension($extension);
  if (!isset($libraries[$name])) {
    return FALSE;      // no exception, no log
  }

// core/lib/Drupal/Core/Asset/AssetResolver.php:156-164
$definition = $this->libraryDiscovery->getLibraryByName($extension, $name);
if (empty($definition[$asset_type])) {
  unset($libraries[$key]);   // silently dropped
}
```

So a rename in the theme would have removed the Closed column and the operations menu from the widget with nothing to indicate it had happened — the same class of invisible failure as the modal bug, just with a different trigger.

The CSS and JS now live in `ys_core` alongside the hook that attaches them. Verified on Lando that the asset is served from the module path:

```
https://yalesites-claude.lndo.site/profiles/custom/yalesites_profile/modules/custom/ys_core/js/office-hours-widget.js
```

and that the widget still builds the column: 1 `th.th__closed`, 7 `.office-hours-closed__input` (one per weekday), 14 `details.ys-office-hours-actions`.

**Shipped as plain CSS, SCSS source dropped.** No custom module in the profile carries SCSS or a sass build step — `ys_core`'s other ten stylesheets are hand-written CSS. The SCSS was only nesting and BEM concatenation, no variables or mixins, so nothing was lost except the comments, which I restored into the CSS (the compiler had stripped them, so the shipped file previously had none). The Gin rules are unchanged and still no-op elsewhere: each is either scoped to `.gin--dark-mode` or reads a `var()` with a fallback.

### 2. A stale comment in `libraries.yml`

It claimed `libraries-extend` did the attaching. Nothing on the branch used `libraries-extend` — your `ys_core` hook does, and its docblock already explained why. A maintainer following the comment would have gone looking in `ys_admin_theme.info.yml` for an entry that was never there. Removed with the library move.

### 3. The Closed control deleted an editor's own note

This is the one with real user impact.

The control records a closed weekday by writing the translated word "Closed" into contrib's `comment` column, then worked out later whether that value was *ours* by comparing the string:

```js
const retractLabel = () => {
  if (comment.value.trim().toLowerCase() === closedLabel.toLowerCase()) {
    comment.value = "";
  }
};
```

That guess cannot distinguish our marker from an editor who simply typed "Closed" — and the function **deletes what it matches**. So an editor who typed `Closed` as their own note and then entered hours for that day had it silently erased. Reproduced against the old logic:

```
editor typed "Closed"                -> old code leaves: ""
editor typed "Closed for renovation" -> old code leaves: "Closed for renovation"
```

It also stopped matching whenever the interface language or the translation changed, orphaning markers we had written ourselves.

**Fix:** provenance now lives out of band, on a `data-ys-closed-marker` attribute set at the moment we write the label and released the moment the editor edits that field. The rule is *never delete a string we did not write*. `hasOwnNote()` reads the same attribute, so a day closed by the editor's own prose still reports as closed and still disables the tick.

Verified in the widget on Lando, all four cases:

| Case | Result |
|---|---|
| Editor types "Closed", then enters hours | note **survives** |
| Editor ticks Closed (we write the marker) | marker written, tick checked |
| …then enters hours | our marker **is** retracted |
| Editor edits our marker to "Closed all week", then enters hours | now theirs, **survives** |

### 4. `once` was reached for as a global

`((Drupal) => { … })(Drupal)` used `once()` from global scope. It worked, because `core/once` is a declared dependency that exposes a global, but it diverged from the convention of naming dependencies in the wrapper — which is also what makes the dependency legible at the top of the file. Now `((Drupal, once) => { … })(Drupal, once)`.

Dropped two of the file's three eslint errors. The remaining two are the `Drupal.behaviors` `no-param-reassign` and a call-site `no-undef`, both of which are mis-tuned-config artefacts rather than code defects, tracked on yalesites-org/YaleSites-Internal#1684.

---

## Needs your decision: the other half of finding 3

The review recommended going further — replace the written word "Closed" with a non-prose sentinel like `YS_CLOSED` and translate it only for display. **I did not do that, because it would have shipped a visible regression**, and I think it runs into the exact thing you parked on 2026-09-11.

The stored comment is simultaneously the persistence mechanism *and* the visible label:

- `OfficeHoursItem::isValueEmpty()` drops a weekday with no hours **and** no comment, so a deliberately-closed day must carry a non-empty comment to persist at all.
- `OfficeHoursItemListFormatter` applies its own `closed_format` label **only** when `empty($info['comments'])` — a non-empty comment suppresses it.
- `formatComments()` then renders the stored string verbatim.

So writing `YS_CLOSED` would put "YS_CLOSED" on the public page for every closed day until the display layer is ours. Translating it for display needs that Twig condition — which is the "second half of route 1" you described in [your 2026-09-12 comment](https://github.com/yalesites-org/yalesites-project/pull/1477#issuecomment-5643065274) and parked pending design.

**What that leaves:** two cases go un-retracted rather than risking an editor's note — a marker written before a save, and one written before contrib's "Add time slot" AJAX replaces the table (the attribute does not survive the rebuild; the stored value does). Both leave a visible stale "Closed" beside hours that the editor can clear. That is a worse cosmetic outcome than the old code and a much better data outcome, and it seemed the right side to err on. **Say if you'd rather I inverted that trade-off** — the alternative is to keep deleting on a string match and accept the data loss until the display work lands.

<details>
<summary><b>Also worth knowing</b></summary>

**Not done, deliberately:** the review also asked for unit tests on the pure helpers (`isFirstSlotRow`, `dayLabelOf`, and the retraction predicates). There is no JS test harness in `yalesites-project` — no jest, vitest or mocha in `devDependencies`, no `node --test` script, and no existing `*.test.js` anywhere in the custom tree. Adding one is repo infrastructure well beyond this PR, and `lint:js` is not in CI either, so a new suite would not be enforced. Flagging rather than bolting a framework on unilaterally.

**Not done, deliberately:** rebuilding the Closed control through the Form API (a real `#type => checkbox` with a value callback rather than a JS-built control). It is the sound long-term shape — it would give `#default_value`, validation, a non-JS fallback, and testability without a browser — but it is a larger change than this remediation pass, and it overlaps the parked display work.

**Pre-existing, not mine:** the form logs a console error from Gin's own `description_toggle.js`. That is yalesites-org/YaleSites-Internal#1676, unrelated to this PR.

**Verification caveat:** I exercised the widget directly rather than completing a full save-and-render round trip — `/block/add/office_hours` renders no block-description field on this branch, so the form would not submit. The diff touches no storage or rendering code (the only PHP change is one library string and a docblock), so the storage contract is unchanged by construction, but the round trip is worth doing during functional review.

</details>

### Functional testing steps for this pass

- [ ] Open an Office Hours block's edit form (both the standalone form and the Layout Builder "Configure block" modal). The **Closed** column and the `⋯` operations menu still appear — this confirms the library still resolves after moving from the theme to `ys_core`.
- [ ] View source and confirm `office-hours-widget.js` is served from `…/modules/custom/ys_core/js/`, not from `…/themes/custom/ys_admin_theme/js/`.
- [ ] On a day with no hours, type `Closed` into the Comment field yourself, then enter hours for that day. **Your note must still be there** — this is the bug that was fixed.
- [ ] On another day, tick **Closed**, confirm the comment fills in, then enter hours for that day. The label we wrote **should** disappear.
- [ ] Tick **Closed** on a day, then edit the comment text to something of your own, then enter hours. Your edited text must survive.
- [ ] Switch the admin theme to Claro and confirm the widget still works and nothing looks broken — the Gin-specific rules should simply not apply.

PR #1478, opened 2026-08-18
1531: Release Feedback: Export Content Events Add Dates

dblanken-yale:
### Description of work

- Adds a **Dates** column to the Events CSV export and a **Resource Publication Date** column to the Resources CSV export. Both sit immediately after Title, so the file reads in roughly the same order as the Manage table it was exported from.
- The Events **Dates** cell lists *every* occurrence of the event, oldest first, separated by `", "`. Stored Smart Date delta order is not chronological, so the occurrences are sorted before rendering — a recurring series exports all of its dates, where the on-screen Date column shows only the first and last (it is labelled "Date (First and last only)").
- Occurrences render in the site's configured timezone using the platform's existing `event_date_only` and `event_time_only` date formats rather than a new pattern — the same pair `ys_localist`'s `MetaFieldsManager::orderEventDates()` already uses. An all-day occurrence reads `(All day)` instead of a 12:00 am to 11:59 pm range, and an occurrence that crosses midnight repeats the end date.
- **Resource Publication Date** is emitted verbatim from storage. The field is date-only and already stored as `Y-m-d`, which is exactly what the Manage Resources column renders via `html_date` — so there is no `strtotime()` round-trip whose only possible effect would be shifting the date by a day across a timezone boundary.
- An event with no dates, or a resource with no publication date, exports an empty cell. No placeholder text, no 1970 fallback.
- Both new columns go through `ContentExportBuilder::sanitizeCell()` like every other cell, so a date value can never be read as a spreadsheet formula.
- `ContentExportBuilder` stays free of injected services so its column map remains directly unit testable: the `DateFormatterInterface` is a `getRow()` parameter, and `ContentExportController` injects `date.formatter` and passes it down. There is no `\Drupal::` call in the builder.
- 15 new unit tests in `ContentExportBuilderTest` (16 -> 31 tests, 41 -> 65 assertions). The event fixtures use real timestamps pulled from `node__field_event_date`, including a series whose stored deltas run 2036, 2029, 2024, 2025, 2027 — so the chronological sort is actually exercised.
- Updates the column list in `ys_content_export/README.md`. The user-guide page is tracked separately in yalesites-org/YaleSites-Internal#1472.

Known trade-off, accepted during grooming: a weekly recurring series produces a long cell, and a cell holding many dates cannot be sorted as a date in a spreadsheet. If editors later ask to sort or filter by date, a separate first-instance start column can be added as a follow-up.

No config changes, no new module dependencies, and no update hook.

### Functional testing steps:

- [ ] Go to **Manage Events** (`/admin/content/manage-events`) and press **Export to CSV**. Confirm the file has a **Dates** column immediately after Title.
- [ ] Open the file in Excel *and* Google Sheets. Confirm the dates are intact, accented characters still render (e.g. the en-dash in "K–12 Environmental Science Curriculum Guide" on Resources), and a multi-date cell stays in a single cell rather than spilling across columns.
- [ ] Find a recurring event with several occurrences and confirm the CSV lists **all** of them, oldest first — not just the first and last that the on-screen column shows.
- [ ] Confirm an all-day event reads as `(All day)` rather than showing a 12:00 am to 11:59 pm range, and that a multi-day all-day event shows both its start and end date.
- [ ] Confirm times show in Eastern time (`EST`/`EDT` as appropriate for that date) and not as raw UTC.
- [ ] Create or find an event with no date and confirm its Dates cell is empty — no placeholder, no 1970 date.
- [ ] Apply an exposed filter or search on Manage Events, then export, and confirm the export still matches what is on screen and still has the Dates column.
- [ ] Repeat on **Manage Resources**: confirm a **Resource Publication Date** column immediately after Title, formatted `YYYY-MM-DD`, matching the value shown in the on-screen column for the same row.
- [ ] Clear the publication date on a resource and confirm its cell exports empty.
- [ ] Spot-check that the other exports (Pages, Posts, Profiles) are unchanged — they should have no date column.

References yalesites-org/YaleSites-Internal#1531

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_content_export/README.md
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_content_export/README.md
@@ -12,12 +12,26 @@ includes one row per item with these columns:
 
 - **Title**, **URL** (path alias), **Published** (Yes/No), and
   **CAS Protected** (Yes/No — whether the item requires CAS login)
+- The type's date column, immediately after Title — **Dates** (Events),
+  **Resource Publication Date** (Resources). Other types have no date column.
 - **Tags**, **Audience**, **Custom Vocab** (on every type)
 - The type's category column — **Category** (Pages, Posts), **Event Category**,
   **Resource Category** — or **Affiliation** (Profiles)
 
 Taxonomy cells list every applied term, separated by ", " (matching the on-screen
-columns). The export reflects the same items you see in the Manage view — including
+columns).
+
+The Events **Dates** cell lists *every* occurrence of the event, oldest first,
+separated by ", " — so a recurring series shows all of its dates, where the
+on-screen Date column shows only the first and last. Dates render in the site's
+timezone, and an all-day occurrence reads "(All day)" rather than a 12:00 am to
+11:59 pm range. **Resource Publication Date** is the same `YYYY-MM-DD` shown on
+the Manage Resources screen. An item with no date exports an empty cell.
+
+Note that a cell holding many dates cannot be sorted as a date in a spreadsheet;
+that trade-off was accepted in favour of showing every occurrence.
+
+The export reflects the same items you see in the Manage view — including
 any filters or search you have applied — and both published and unpublished items,
 subject to your access.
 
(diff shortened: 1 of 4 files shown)


miketullo95 (approved):
@dblanken-yale Looks good to me. Approving now, and I'm testing it in the multidev.

I read the diff against #1531 and it lines up with the acceptance criteria point for point:

- **Dates** (Events) and **Resource Publication Date** (Resources) both sit immediately after Title, via the new `BUNDLE_DATE` map in `ContentExportBuilder.php`. Headers match what editors already see on the Manage screens.
- Occurrences get sorted before rendering, so a recurring series with out-of-order stored deltas still exports oldest first. Good catch that the stored delta order isn't chronological, and nice that the test fixtures use real timestamps that actually exercise it.
- All-day reads `(All day)` instead of a 12:00 am to 11:59 pm range, dates render in Eastern, and an item with no date exports an empty cell with no 1970 fallback.
- Both new columns go through `sanitizeCell()`, and there's a test asserting a leading `-` gets quoted.
- Resource publication date emitted verbatim rather than round-tripped through `strtotime()` is the right call, since that could only ever shift the date by a day.
- Builder stays service-free with the formatter passed into `getRow()`, which keeps the column map directly unit testable.
- README column list is updated.

Nothing here goes beyond what the ticket asked for, and I don't see anything in the AC left unaddressed. The long-cell tradeoff on weekly recurring series is the one we already accepted in grooming, and it's documented in the README, so no concern from me there.

One follow-up on my side, not on you: the user guide column list in #1472 needs both new columns added. I'll handle that.

PR #1479, opened 2026-08-18
1558: RC: Dashboard: indicate which announcement is new in the Announcements section

dblanken-yale:
### Description of work

The Dashboard menu item already picks up an unread badge when a new announcement is published, but the dashboard itself rendered every announcement identically — editors knew *something* was new, not *which* post. Worse, `DashboardController::content()` called `markAllRead()` on every page view, so an editor glancing at the dashboard on their way to edit a page silently burned their unread markers without reading anything.

This adds a per-item marker and moves clearing off the page-view path onto an explicit control.

- **Per-item "New" marker.** `DashboardAnnouncements::getAnnouncementsForUser()` decorates each item with an `is_new` flag derived from the same `announcements_last_seen` comparison that already drives the toolbar badge count. `getUnreadCount()` is now re-expressed on top of it, so the badge number and the on-page markers are the same fact and cannot drift — no second source of truth, and no change to the read-state data model.
- **Explicit clearing.** A new `MarkAnnouncementsReadForm` calls the existing `markAllRead()`. The automatic stamp on dashboard view is gone: viewing the dashboard no longer clears anything. The control is only built when something is actually unread, so editors are never offered a button that would do nothing.
- **Cache correctness.** The announcements render now varies per user, so the page carries the `user` cache context (which subsumes the `user.permissions` context it had) and the existing `DashboardAnnouncements::unreadCacheTag($uid)` tag. `markAllRead()` already invalidates that tag, so the markers and the toolbar badge clear in the same interaction with no manual cache rebuild — verified locally.
- **Reuse.** The marker reuses the existing `.ys-dashboard-badge` pill that the toolbar count already uses (`ys_core/dashboard` now depends on `ys_core/dashboard_badge`) rather than restating the pill's shape and colour.

Feed order is untouched — marking an item new does not reorder anything.

**Accessibility (WCAG 2.1 AA).** The marker is not colour-alone: the pill contains the visible word "New" plus a visually hidden " announcement", and sits inside the announcement's own `h3` so it is announced together with the title rather than orphaned from it. The control is a real submit button with the self-describing name "Mark all announcements as read", and pressing it emits a status message ("All announcements have been marked as read.") so the result reaches assistive tech instead of markers merely vanishing. A light accessibility gate over the rendered section passed with no Critical or Serious findings; measured contrast is 4.93:1 for the pill text and 6.38:1 for the button label, both clear of the 4.5:1 AA threshold at the 12px size used.

**Known tradeoff, per the ticket and intended:** if an editor never presses the button, the badge stays up indefinitely. There is deliberately no time-based auto-clear.

**Two things for the reviewer to weigh in on:**

1. **Design/UX sign-off is still outstanding.** The ticket asks Design and UX to pick the affordance rather than shipping developer choice. This PR picks a **"New" text pill** (over a red count or a whole-row highlight) because the text carries the meaning on its own — which the accessibility criteria require anyway — and because a per-item marker scales when several items are new at once. The "Mark all announcements as read" control sits in the Announcements section header, right-aligned beside the heading, so its relationship to the list it acts on is unambiguous. Both are cheap to change: the pill is one Twig block plus scoped CSS.
2. Contrast was measured in Gin **light** mode. The pill's red/white pair is fixed rather than themed so it should not vary, but a dark-mode spot check is worth an extra pair of eyes.

**On test coverage of the POST path, stated plainly:** the Kernel tests drive the form through `formBuilder()->submitForm()`, which core marks as programmed — that path skips CSRF validation and returns before redirect handling. So the automated tests pin `markAllRead()`, the cache-tag invalidation, the status message, and the rendered markup, but *not* the real request cycle. That cycle was verified **by hand** instead, twice: a `curl` POST carrying a genuine `form_build_id`/`form_token` scraped from the page, and a real button click driven through a browser. Both redirected back to the dashboard, emitted the status message, and left zero markers and no button. It is not covered by a `BrowserTestBase` test because Functional tests do not run in this local environment (and CI runs no PHPUnit at all), so an added one would have been unverifiable rather than reassuring.

Not done here: the ticket's documentation item (whether the dashboard user guide page on yalesites.yale.edu needs a line about the marker and the control) is a content change outside this repo.

### Functional testing steps

Dashboard announcements are per-site config (`ys_core*` is in `config_ignore`), so a local test needs the feed pointed somewhere first.

1. Set up a local announcements feed. Easiest is to make the site both source and consumer:
   - `lando drush config:set ys_core.dashboard_settings announcements_enabled 1 -y`
   - `lando drush config:set ys_core.dashboard_settings announcements_source_enabled 1 -y`
   - `lando drush config:set ys_core.dashboard_settings announcements_feed_url "https://<your-lando-host>/api/dashboard-announcements" -y`
2. Create a `tags` taxonomy term named **Dashboard Announcement**, then create 3 **Post** nodes tagged with it and **published** (set the moderation state to Published — `setPublished()` alone is not enough). Give them different `changed` times.
3. Make some of them unread for your user: `lando drush ev '\Drupal::service("user.data")->set("ys_core", <your-uid>, "announcements_last_seen", time() - 86400*3);'` then `lando drush cr`.
4. Visit `/admin/yalesites/dashboard`. **Expect:** a red "New" pill on each announcement newer than that timestamp and none on the older ones; the list order unchanged; a "Mark all announcements as read" button beside the Announcements heading; the Dashboard toolbar link showing a count equal to the number of pills.
5. **Reload the page a few times without pressing the button.** Expect the pills and the toolbar count to still be there — this is the regression the ticket is about, and it used to clear on the first view.
6. Press **Mark all announcements as read**. Expect: a green status message "All announcements have been marked as read."; every pill gone; the toolbar count gone in the same interaction (no cache rebuild); and the button itself gone, since there is nothing left to clear.
7. Keyboard/screen reader pass: Tab to the button and press Enter — it should behave identically. With a screen reader, the marker should read as "New announcement" next to its title, and the status message should be reachable.
8. Publish a new announcement post. Expect the pill and the toolbar count to come back for that item only.

#### Additional steps for the toolbar badge fix (review round 2)

The stale-badge report needs **two accounts and a real render cache** to reproduce, so it is worth
doing these on a multidev rather than locally. A default local checkout cannot show the bug at all:
`web/sites/default/settings.local.php` points the `render`, `page` and `dynamic_page_cache` bins at
`cache.backend.null`, and the bug lives in a cache entry that is never written there. Comment those
three lines out and `lando drush cr` if you do want to see it locally.

9. Make two accounts with the **same role** (e.g. two platform admins), and give them **different**
   unread counts — set `announcements_last_seen` per uid as in step 3.
10. `drush cr`, then load `/admin/content` as the **first** user and note the toolbar count.
11. Load `/admin/content` as the **second** user. **Expect:** their own count. Before this fix they
    saw the first user's count, because the toolbar cache entry is shared between users with the
    same permissions and had the first user's uid baked into it.
12. As the second user, open the dashboard, press **Mark all announcements as read**, then visit
    `/admin/content`, `/admin/config` and `/node/add`. **Expect:** no badge anywhere. This is the
    reported bug — the badge used to keep showing the other user's count on those pages.
13. Reload as the **first** user. **Expect:** their count is untouched — one editor clearing their
    own announcements must not clear anyone else's.
14. Log out and load the front page as an anonymous visitor. **Expect:** a normal page, no badge.

### Testing checklist

- [x] `lando composer lint:php` — exit 0
- [x] `lando composer code-sniff` (Drupal + DrupalPractice, whole custom tree) — exit 0
- [x] PHPUnit: `ys_core` suite — baseline 274 tests / 796 assertions / 0 failures / 1 skipped; after **290 tests / 868 assertions / 0 failures / 1 skipped** (same pre-existing skip). 16 new tests, no regressions.
- [x] Visual + interaction verification on local Lando
- [x] **Round 2:** two-user reproduction of the stale badge with the render/page/dynamic-page caches
      re-enabled, then the same sequence re-run against the fix — see steps 9-14
- [x] **Round 2:** `ys_core` Kernel suite 47 -> **52 tests / 580 assertions / 0 failures** (5 new,
      no regressions); Unit unchanged at 243 / 311 / 1 pre-existing skip
- [x] Light accessibility gate on the rendered section — pass

References yalesites-org/YaleSites-Internal#1558

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/css/badge.css
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/css/badge.css
@@ -1,0 +1,27 @@
+/**
+ * The unread-announcements pill.
+ *
+ * Modeled on Drupal core's announcements_feed bell-icon dot, but rendered as a
+ * small red pill so editors can see what is waiting. Two consumers, which is
+ * why this is its own library rather than living in either one's stylesheet:
+ * the Dashboard admin menu link shows the unread count in it, and each unread
+ * announcement on the dashboard itself shows the word "New" in it.
+ *
+ * @see web/profiles/custom/yalesites_profile/modules/custom/ys_core/templates/ys-dashboard.html.twig
+ * @see \Drupal\ys_core\Toolbar\DashboardToolbarLazyBuilders::renderBadge()
+ */
+
+.ys-dashboard-badge {
+  display: inline-block;
+  min-width: 1.25em;
+  margin-inline-start: 0.4em;
+  padding: 0 0.4em;
+  border-radius: 999px;
+  background-color: #d72c0d;
+  color: #fff;
+  font-size: 0.75rem;
+  font-weight: 600;
+  line-height: 1.4;
+  text-align: center;
+  vertical-align: baseline;
+}
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/css/dashboard-badge.css
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/css/dashboard-badge.css
@@ -1,25 +1,11 @@
 /**
- * Unread-announcements badge on the Dashboard admin menu link.
+ * The Dashboard admin menu link that carries the unread-announcements pill.
  *
- * Modeled on Drupal core's announcements_feed bell-icon dot, but rendered as
- * a small red pill containing the count so editors can see how many items
- * are waiting before they click through.
+ * The pill itself is `ys_core/badge`, shared with the dashboard page; every
+ * rule here is specific to rendering it inside the admin toolbar.
+ *
+ * @see \Drupal\ys_core\Toolbar\DashboardToolbarLazyBuilders::renderBadge()
  */
-
-.ys-dashboard-badge {
-  display: inline-block;
-  min-width: 1.25em;
-  margin-inline-start: 0.4em;
-  padding: 0 0.4em;
-  border-radius: 999px;
-  background-color: #d72c0d;
-  color: #fff;
-  font-size: 0.75rem;
-  font-weight: 600;
-  line-height: 1.4;
-  text-align: center;
-  vertical-align: baseline;
-}
 
 /* FontAwesome Free "gauge" icon (v7.3.0, CC BY 4.0). Positioned absolutely so
    the icon sits vertically centered relative to the link's own box (matching
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/css/dashboard.css
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/css/dashboard.css
@@ -62,6 +62,34 @@
   margin-block-end: 0;
 }
 
+/* Header row: the Announcements heading and the "mark all as read" control,
+   wrapping onto separate lines when the column is too narrow for both. */
+.ys-dashboard__announcements-header {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: baseline;
+  justify-content: space-between;
+  gap: 0.5rem 1rem;
+  margin-block-end: 0.75rem;
+}
+
+/* The header row carries the gap below itself now. */
+.ys-dashboard__announcements-header > h2 {
+  margin-block-end: 0;
+}
+
+/* Gin spaces form actions for the bottom of a settings form; inline in a
+   section header that margin pushes the button out of line. */
+.ys-dashboard__announcements-header .form-actions {
+  margin: 0;
+}
+
+/* Placement of the "New" pill against an announcement title. The pill itself is
+   .ys-dashboard-badge (css/badge.css), shared with the toolbar count. */
+.ys-dashboard__announcement-new {
+  vertical-align: middle;
+}
+
 .ys-dashboard__announcement-list {
   list-style: none;
   margin: 0;
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Controller/DashboardController.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Controller/DashboardController.php
@@ -5,6 +5,7 @@ namespace Drupal\ys_core\Controller;
 use Drupal\Core\Controller\ControllerBase;
 use Drupal\Core\Extension\InfoParserInterface;
 use Drupal\ys_core\DashboardAnnouncements;
+use Drupal\ys_core\Form\MarkAnnouncementsReadForm;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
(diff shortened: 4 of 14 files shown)


miketullo95:
@dblanken-yale one quick thing, noticed that after clicking the clear/read all notifications the dashboard toolbar notification is still showing, the badges are gone but the "3" is still there. maybe a cache clearing thing?

image

dblanken-yale:
@miketullo95 Good catch, and it was not a cache-clearing thing — thank you for screenshotting it,
because the screenshot is what made it findable. Reproduced, root-caused, fixed and pushed.

## The finding

> after clicking the clear/read all notifications the dashboard toolbar notification is still
> showing, the badges are gone but the "3" is still there. maybe a cache clearing thing?

**Fixed.** It turned out to be worse than a stale number: the count you were looking at was very
likely **not yours**.

## What was actually wrong

The badge is attached to the Dashboard menu link as a `#lazy_builder` placeholder, and it was
passing the current user's uid as a **lazy-builder argument**. Two things about Drupal make that
unsafe:

- `PlaceholderGenerator::createPlaceholder()` hashes the *whole* placeholder array — arguments
  included — into the placeholder token, and hands back an element carrying only `#markup` and
  `#attached`. The placeholder's own cache metadata never reaches its parent.
- That parent is core's toolbar, which `toolbar_page_top()` render caches as
  `keys: ['toolbar'], contexts: ['user.permissions']` — **one entry shared by every user holding the
  same permissions**.

So the first editor to warm a given cache entry had their uid frozen into it, and every other editor
reading that entry got *that* editor's badge rendered. And because clearing your own read state
invalidates `ys_core:dashboard_badge:<your-uid>` — which is neither the uid in the token nor a tag
on that entry — pressing the button could not clear it.

That is exactly your screenshot: the pills cleared because the dashboard render declares
`contexts: ['user']` and is never shared, while the toolbar kept rendering somebody else's count.

Two defects, one cause: a **cross-user count leak** and the stale badge you reported.

## The fix

Drop the uid from the lazy-builder arguments and resolve the viewer inside the lazy builder. The
placeholder is then identical for everyone, so sharing the parent entry is correct, and replacement
always renders for whoever is actually looking. Bonus: the user entity load goes away with it, so
that is one fewer storage load on every admin page render.

I left a fairly long comment at the attach site explaining why the uid must not come back — it looks
like a harmless parameter, and it is not.

## Testing gotcha worth knowing about

**This bug cannot be reproduced in a default local checkout.** `settings.local.php` points the
`render`, `page` and `dynamic_page_cache` bins at `cache.backend.null`, so the cache entry the bug
lives in is never written. That is why it survived the first round of local verification — my
"verified locally" claim in the PR description was made in an environment structurally incapable of
showing it, which I should have caught. Steps 9-14 in the description now spell out the two-account
reproduction, and note to run it on a multidev or to comment those three lines out locally.

## Verification

Reproduced first, with the caches restored — editor B (1 unread) saw editor A's count of 3, and
after B pressed the button the badge stayed at 3 on other admin pages. After the fix, the same
sequence: B sees 1, clears it, and the badge is gone everywhere, while A's count is untouched.

The toolbar badge had **no test coverage at all** — the existing tests only covered the dashboard
page. Added a Kernel test that pins the placeholder token being identical across users, the badge
rendering for the viewer rather than for whoever warmed the cache, clearing on mark-all-read, being
tagged for the user it was rendered for, and anonymous getting nothing. Four of the five failed
against the old code.

- `ys_core` Kernel, full suite: 47 -> **52 tests / 580 assertions / 0 failures** (no regressions)
- The two Kernel classes covering this feature (`DashboardAnnouncementIndicatorTest` +
  `DashboardToolbarBadgeTest`) re-run green after the final review edits: **15 tests / 76 assertions**
- `ys_core` Unit: **243 / 311 / 0 failures**, 1 pre-existing skip — unchanged
- `lando composer lint:php` and `lando composer code-sniff` — both exit 0

Also ran a security review over the change, since the leak is an information-disclosure shape: no
issues, and it confirmed the fix closes the leak without opening a different one.

Back to you for re-review.

miketullo95 (approved):
@dblanken-yale Approving.

Tested on the multidev and the feature behaves. On the open design question from your PR description: I am signing off on the "New" text pill and on putting the control beside the Announcements heading. That closes the design AC on #1558, so treat it as settled rather than outstanding.

Deriving `is_new` and `getUnreadCount()` from the same `getAnnouncementsForUser()` call is the right shape. The AC asked for no second source of truth and this satisfies it structurally instead of by convention, which is better than what I asked for.

Two notes, neither blocking the merge:

**The two-account pass still needs doing before release.** My multidev test was single-user, steps 1 through 8 only. So the round 2 toolbar fix, where the uid was baked into the placeholder arguments and editors were reading each other's counts, is the part of this diff I have not actually verified. Since it cannot be reproduced locally either, it needs to go on the release testing steps for #1558 rather than being considered covered.

**That toolbar fix is also scope beyond this ticket.** It is a genuine cross-user correctness bug and none of #1558's acceptance criteria cover it. Good catch and I am glad it is fixed here, but I want it visible at release time rather than buried in a feature PR, so I will make sure it gets its own line.

On dark mode: I think it is a smaller risk than your note suggests. The pill hardcodes its own foreground and background, so its text contrast cannot vary with the theme. The only open question is how the red reads against Gin's dark background, and since the word "New" carries the meaning rather than the pill's shape, I do not think that gates anything.

The documentation item on #1558 stays open. You are right that it is a yalesites.yale.edu content change and not yours to make here.

PR #1480, opened 2026-08-18
1560: RC: Audit settings pages for platform-admin-only fields and move them to Platform Admin Settings

dblanken-yale:
### Description of work

**The audit phase is complete and posted on the ticket** — [inventory comment](https://github.com/yalesites-org/YaleSites-Internal/issues/1560#issuecomment-5325341189). It covers 11 gated items across every `ys_*` settings form, each with a move/stay recommendation, the two decisions the ticket asked for, and two latent bugs the sweep turned up. It does not need to become an epic.

The ticket asks for sign-off before anything moves, so **this PR first moved only the four settings the ticket itself names without qualification, and left every judgment call exactly where it was.** The rest went up as recommendations in that comment; review has since signed them off, and the one it disagreed with — the announcements `source` section — has been moved too, bringing the total to five. See "Deliberately not moved" below for what stays.

**Moved to the Platform Admin Settings page**
- New `SiteBrandingPlatformAdminSetting` plugin (weight `-20`) takes `site_name_image`, `site_wide_branding_name` and `site_wide_branding_link` out of Header settings. These were the ticket's headline example: three whole `details` sections that appeared and disappeared depending on who was looking.
- New `EnvironmentIndicatorPlatformAdminSetting` plugin (weight `-10`) takes `environment_indicator_show` out of Site settings. Whether the dev/test/live banner shows is platform infrastructure, not site content.
- New `AnnouncementsSourcePlatformAdminSetting` plugin (weight `20`) takes `announcements_source_enabled` and `announcements_source_term` out of Dashboard settings, landing directly under the existing **Dashboard Announcements Feed** section. Added on review: `announcements_feed_url` was already on this page, so leaving the publish side behind kept one concern split across two pages. Every announcements control a platform admin owns is now in one place.
  - The `#states` selector had to be re-namespaced — the page renders every section with `#tree`, so the checkbox input name is `announcements_source[announcements_source_enabled]`. The un-namespaced selector carried over from Dashboard settings would silently never have matched, and the tag field would always have shown. A test pins it.
  - `ensureAnnouncementTerm()` now runs **ahead of** the no-change short-circuit and on every save while publishing is on. The tag is the one thing here another actor can delete out from under the config, and the old form was self-healing only because it re-ran on every platform-admin save. A cleared tag field normalizes to `Dashboard Announcement` rather than storing an empty string the controller would silently paper over, and an existing-but-**unpublished** tag now warns instead of reading as fine — the existence lookup is deliberately not access-checked, while the feed's query is.
  - It also uses the injected term storage instead of the static `Term::create()`, which routes it through DI rather than the global container and is what makes the tag-creation path unit-testable.
- **Config keys are unchanged** — the values still live in `ys_core.header_settings`, `ys_core.site:environment_indicator.show` and `ys_core.dashboard_settings`, so there is no update hook and no data migration.
- **The save-path guard is deleted, not relocated.** `HeaderSettingsForm::submitForm()` needed a `ys_core_allow_secret_items()` conditional purely to stop a site admin's ordinary save from NULLing the branding values; with the fields gone the form simply never writes those keys. Same for the environment indicator in `SiteSettingsForm`. Both are now covered by tests (see below), because that is the one thing in this change that could silently wipe live data.

**One gating mechanism**
There were four spellings of "is this a platform admin". They are now one: `hasPermission('administer platform admin settings')` — the permission that already gates the Platform Admin Settings page, granted only to `platform_admin`, and satisfied by user 1 through Drupal's normal permission bypass. So it is exactly equivalent to the old role check, just idiomatic and decoupled from the role machine name.
- `ys_core_allow_secret_items()` had zero callers left afterwards and is **deleted** rather than kept as a wrapper.
- `DashboardSettingsForm::isPlatformAdmin()` was consolidated onto it and then, once the announcements source section moved, had **zero callers left and is deleted** — the same fate as `ys_core_allow_secret_items()`.
- `ys_localist` no longer reaches into `ys_core` behind a `function_exists()` guard — that removed an undeclared soft dependency between the modules.
- `SiteSettingsForm`'s `cas_app_name` check stays `id() == 1` and is now commented as **intentionally narrower**: it is user-1-only, so moving it to a page every platform admin can reach would widen access, not tidy it.
- The permission is exposed as `PlatformAdminSettingInterface::PERMISSION` rather than being hand-typed at each call site, so "one mechanism" is a single constant instead of a copy-pasted string. (`ys_core.routing.yml` and `ys_core.permissions.yml` necessarily still spell it out — YAML cannot reference a PHP constant.)

This also unblocked testing. The old procedural function cannot be loaded in a PHPUnit unit test, which is why `HeaderSettingsForm` and `SiteSettingsForm` had no unit test at all and `LocalistSettingsTest` carried a comment explaining it could only ever exercise the disabled branch. A permission check on a mocked account tests both branches.

**Housekeeping**
- Explicit weights on all four plugins, so the page order is a decision rather than the alphabetical-by-label accident it was (both pre-existing plugins were weight `0`).
- Removed genuinely dead members the move orphaned: `HeaderSettingsForm`'s `AccountProxy` and `YaleSitesMediaManager` dependencies, and `SiteSettingsForm`'s `$currentUserSession`, which was a duplicate alias assigned from the same constructor argument as `$currentUser`.

**Test coverage added** (`ys_core` 274 -> 294 passing, `ys_localist` 69 -> 70)
- `SiteBrandingPlatformAdminSettingTest`, `EnvironmentIndicatorPlatformAdminSettingTest` — fields, defaults, fallbacks, and the exact config keys written.
- `HeaderSettingsFormTest`, `SiteSettingsFormTest` — new files; assert the forms still save what they own **and never write the moved keys**. This is the regression guard for the deleted guards.
- `PlatformAdminSettingsFormTest` (kernel) — new; the page had no test of its own at all. Covers plugin discovery, the `#tree` namespacing, and weight ordering.
- `DashboardSettingsFormTest` — new; both sides of the consolidated permission gate.
- `LocalistSettingsTest` — extended to cover the previously untestable platform-admin branch.

**Two things the review caught, worth knowing about**
- **Saving branding has to flush the render cache, and that nearly got lost.** `HeaderSettingsForm::submitForm()` ends with `$this->cacheRender->invalidateAll()`, and it turns out that call is load-bearing rather than housekeeping: `CoreTwigExtension` grabs `ys_core.header_settings` in its constructor and returns raw values with no cacheability bubbling, so the rendered header carries no `config:ys_core.header_settings` cache tag and `Config::save()`'s own tag invalidation reaches nothing. Without carrying the flush along, a platform admin would upload a new Site Name Image, see "The configuration options have been saved", and the front end would keep showing the old lockup until the next `drush cr`. The plugin now flushes — but only when a branding value actually changed, because this page has one Save button for every section and an unconditional flush would penalise the whole site every time someone saved an unrelated section. Two tests cover both halves.
- **The environment indicator got the same treatment** for a smaller reason: it was rewriting `ys_core.site` (and invalidating its cache tag) on every save of the page regardless of whether the checkbox changed.

**Follow-ups I did not do here** (happy to file tickets):
1. `ys_localist` writes `localist_group` unconditionally although the field is only built when sync is enabled, so saving with sync off wipes the stored group.
2. With `#disabled => TRUE`, `EntityAutocomplete`'s `#element_validate` never runs, so a save could write the label string `"Group Name (42)"` into `localist_group` where a term id is expected. Unreachable today because `yalesites manage localist` is platform-admin-only.
3. `CoreTwigExtension::getHeaderSetting()` bubbles no cacheability. Fixing that would let the render-cache flush above be deleted in favour of a targeted cache-tag invalidation. Worth auditing `getCoreSetting()` for the same gap.
4. `'file_validate_extensions' => ['svg']` is deprecated in Drupal 10.2 and removed in 11. I carried it over verbatim to keep the move behaviour-identical; `SiteSettingsForm` uses the same legacy key for the favicon, so both should migrate to `'FileExtension'` together.
5. `site_wide_branding_link` is a plain textfield with no URL validation, so a `javascript:` value is accepted. Pre-existing and only writable by platform admins, so not a regression from this change, but validation would be an easy win.
6. `site_wide_branding_name` and `site_wide_branding_link` have no entry in `ys_core.schema.yml`.
7. Three ys_beacon unit tests still carry their own copy of the `setProtected()` reflection helper this PR extracted into `Drupal\Tests\ys_core\Traits\ProtectedPropertyTrait`; they can move onto it whenever someone is in there.

**Deliberately not moved — now signed off in review:** `cas_app_name` (user-1-only on purpose, so moving it would widen access rather than tidy it) and the three Localist fields (their gate is dead code today — `yalesites manage localist` is granted only to `platform_admin`, so nobody ever sees the disabled state, and there is no observable problem to fix by moving them). The announcements `source` section was the third item on this list; review asked for it to move, and it has.

### Functional testing steps:
- [ ] As a **platform admin**, go to Settings > Platform Admin Settings. Confirm the sections read top to bottom: **Site Branding**, **Environment Indicator**, **Beacon (AI Chat)**, **Dashboard Announcements Feed**, **Announcements Source**.
- [ ] In **Site Branding**, confirm all three fields are present with their original labels and descriptions, and that Site-wide branding name/link show the current values (defaulting to `Yale University` / `https://www.yale.edu`). Change both, save, and confirm the values persist and the header renders them.
- [ ] Upload an SVG to **Site Name Image**, save, and confirm it replaces the site name text in the header. Then remove it and confirm the text name comes back.
- [ ] In **Environment Indicator**, toggle **Show environment indicator** off, save, and confirm the banner disappears from the front end. Toggle it back on.
- [ ] Go to Settings > Header Settings. Confirm the **Site Name Image** and **Sitewide Branding** sections are gone.
- [ ] Go to Settings > Site Settings. Confirm **Show environment indicator** is gone.
- [ ] **The regression that matters:** log in as a **site admin**, open Header Settings, change nothing, and press Save. Then open Site Settings and press Save. Go back to Platform Admin Settings as a platform admin and confirm the branding name, branding link, site name image and environment indicator toggle are all still exactly as you left them.
- [ ] As a **site admin**, confirm Platform Admin Settings is not reachable (the menu item is absent and the path returns 403).
- [ ] Confirm Dashboard Settings still hides the **Announcements source** section from a site admin and shows it to a platform admin, and that a site admin saving that form does not clear the source settings.
- [ ] Confirm the Localist settings form still behaves as before for a platform admin (fields editable, sync works).

**Added in review — the announcements source move:**
- [ ] As a **platform admin**, on Platform Admin Settings confirm the new **Announcements Source** section sits directly below **Dashboard Announcements Feed**, and that the **Announcement tag** field stays hidden until **Publish an announcements feed from this site** is checked.
- [ ] Go to Settings > Dashboard Settings as a platform admin. Confirm the **Announcements source** section is gone and only the **Announcements** section remains — the form is now identical for platform admins and site admins.
- [ ] **The regression that matters, again:** note the current Announcements Source values, log in as a **site admin**, open Dashboard Settings, change nothing, press Save. Back on Platform Admin Settings as a platform admin, confirm the publish toggle and the announcement tag are unchanged.
- [ ] On Platform Admin Settings, tick **Publish an announcements feed from this site**, enter a tag name that does not exist in the Tags vocabulary, and save. Confirm the status message says the tag was created, that it exists under Structure > Taxonomy > Tags, and that `/api/dashboard-announcements` now answers 200 instead of 404.
- [ ] Save Platform Admin Settings again without touching the Announcements Source fields, and confirm no duplicate tag is created.
- [ ] Untick the publish toggle, save, and confirm `/api/dashboard-announcements` returns 404 again.
- [ ] **Self-heal:** with publishing on, delete the announcement tag from Structure > Taxonomy > Tags, then re-save Platform Admin Settings **without changing anything**. Confirm the tag is recreated rather than left missing.
- [ ] Rename the tag while leaving the publish toggle on, save, and confirm the new name persists.
- [ ] Clear the **Announcement tag** field with publishing on and save. Confirm it stores `Dashboard Announcement` rather than an empty value.

**Added in review — the no-change guards now actually fire:**

Review asked for confirmation that the site branding render-cache flush is conditional on a real change, specifically on a file upload. It is; but confirming it showed the guard never fired at all, because it compared raw stored config against submitted form values and the two sides hold the same values in different shapes (config omits these keys or ships `''`; the form submits an fids list plus the fallbacks it displayed). Both sides are now normalized before comparing. Config keys and stored values are unchanged, so there is still no update hook.

- [ ] As a **platform admin** on Platform Admin Settings, change **nothing** and press Save. Confirm the page saves and the site does not visibly slow down on the next few page loads — the whole-site render cache should no longer be flushed for a save that changed nothing. (Before this change, every save flushed it.)
- [ ] **The flush must still happen when it should:** upload a new SVG to **Site Name Image** and save. Confirm the header shows the new lockup immediately, **without** running `drush cr`. Then remove it and confirm the text site name returns, again with no cache rebuild.
- [ ] Change only **Site-wide branding name**, save, and confirm the header reflects it immediately.
- [ ] Toggle **Show environment indicator** off, save, and confirm the banner disappears from the front end. Toggle it back on and confirm it returns. Do this on a site that has never saved this setting before, if you can — that is the path the fix changes.
- [ ] Save the page repeatedly without touching anything and confirm the branding and environment indicator values stay exactly as they were.

**Note for future site-header work:** `SiteBrandingPlatformAdminSetting::DISPLAY_DEFAULTS` duplicates the `|default('Yale University')` / `|default('https://www.yale.edu')` filters in `component-library-twig/components/03-organisms/site-header/_site-header--secondary.twig`. They must stay in sync — Twig cannot read a PHP constant across repos. Skipping the write when a value is unset is only safe because both sides agree.

**Added in review — the announcements source guard and its warnings:**

Two defects in the Announcements Source section this PR added, both stemming from the page having one Save button for every section. The `ys_core.dashboard_settings` config object does not exist on a site that has never saved this section (`config_ignore` keeps it out of `config:import`), so the no-change guard was comparing an absent stored value against the default the form had just displayed and counting every untouched save as a change. And the section's two taxonomy warnings were re-firing on saves that never touched it. Config keys and stored values are unchanged, so there is still no update hook.

- [ ] As a **platform admin**, on a site that has never saved this section, open Platform Admin Settings, change **nothing**, and press Save. Confirm no Tags warning appears, and that the Announcements Source values are unchanged afterwards.
- [ ] Turn on **Publish an announcements feed from this site** with a tag name, save, and confirm the tag is created. Now go to Structure > Taxonomy > Tags and **unpublish** that term. Return to Platform Admin Settings, change **nothing**, and press Save. Confirm the "exists but is unpublished" warning does **not** appear.
- [ ] **The warning must still reach you when it is actionable:** with the term still unpublished, rename the **Announcement tag** to something else, save, then rename it back to the unpublished term and save. Confirm the "exists but is unpublished" warning **does** appear on that save.
- [ ] **The self-heal must be untouched:** with publishing on, delete the announcement term from Structure > Taxonomy > Tags, then re-save Platform Admin Settings **without changing anything**. Confirm the term is recreated and the "Created the ... tag" message still appears — that message is deliberately not suppressed, because it reports something that actually happened.
- [ ] Save the page repeatedly without touching anything and confirm the Announcements Source values never change.
- [ ] **The feed tolerates a malformed stored tag:** with publishing on and at least one published post tagged as an announcement, set the tag name to whitespace out of band (`lando drush cset ys_core.dashboard_settings announcements_source_term '   '`) and load `/api/dashboard-announcements`. Confirm it still returns the tagged posts rather than an empty `items` array. Set it back afterwards.

References yalesites-org/YaleSites-Internal#1560


miketullo95 (changes requested):
@dblanken-yale This is a genuinely good audit. Rendering each form twice as both roles and diffing the actual build arrays, rather than reading the code and inferring, is why it turned up three items my starting inventory missed plus two latent bugs. That's the level I wanted on this one.

The conservative shape is right too. Moving only the four unqualified items and leaving every judgment call in place is exactly what the ticket asked for, and keeping the config keys where they are means no update hook and no migration to worry about.

Requesting changes on one move, plus a few sign-offs and notes below.

## Sign-offs you asked for

**CAS Application Name: agreed, stays.** Your privilege-widening argument is right. Moving a uid-1-only setting to a page every platform admin can reach makes it less restricted, not tidier. Leave it and keep the comment explaining why it's intentionally narrower.

**Localist fields: agreed, stay.** Since `yalesites manage localist` is platform-admin-only, nobody ever sees the disabled state, so there's no observable problem to fix by moving them. Consolidating the mechanism without relocating the fields is the right amount of work. Worth revisiting the day that permission gets granted to `site_admin`.

**Announcements source: I'd like it moved.** This is the one I'm disagreeing with you on. You flagged the thing that decides it for me: `announcements_feed_url` is already on the Platform Admin Settings page. So the announcements concern is already split across two pages, and leaving the source section behind keeps it split rather than fixing it.

I take the point that the publish side reads well next to the consume side. But I'd rather a platform admin find every announcements control they own in one place than have the publish toggle sit next to fields it relates to but that a different role owns. Please move items 6 and 7 to the Platform Admin Settings page alongside the existing `announcements_feed` section.

Same pattern as the four already moved: the `#access` gate and the submit guard both come out rather than relocating, and `DashboardSettingsForm::isPlatformAdmin()` should have no callers left afterwards, so it can go the way of `ys_core_allow_secret_items()`.

**No read-only mirrors.** Thanks for surfacing the `ys_beacon` precedent, I hadn't clocked that we'd already solved this a third way. I considered it and I'd rather go with clean removal. A read-only field that exists only to say "someone else controls this" is still a row on a form that a site admin has to read past every time, and Header Settings is already long. Site admins never had these values as theirs to set, so I don't think we owe them a breadcrumb on the way out. Same call for the announcements section once it moves.

**Decisions 1 and 2 I'm not signing off on.** Consolidating on one gating mechanism and picking hidden over disabled are architecture calls, not product calls, so they shouldn't rest on my approval. Your reasoning reads well to me and the testability argument is the strongest part of it, given it's the difference between four of these findings being untestable and being covered. But please get our lead developer's eyes on both before this lands. Flagging it because the change is already implemented across five files ahead of that sign-off.

## Process note

The ticket asks for the inventory to be reviewed before anything moves, and the PR opened about an hour after the comment went up. No harm here since you moved only the uncontroversial four and argued everything else instead of doing it, and honestly the inventory is more useful with a concrete implementation next to it. Just want to name it so it doesn't become the default on a change where the gate matters more.

## Notes, not blocking

- The two Localist bugs are listed twice in the description, once under "Follow-ups I did not do here" and again under "Two follow-up tickets worth filing."
- Yes to filing the follow-ups. I'll take that so it doesn't sit on you. The `site_wide_branding_link` one is worth doing sooner than the rest: a plain textfield with no URL validation accepting a `javascript:` value is the kind of thing that's fine right up until it isn't, even platform-admin-only.
- Documentation on the ticket is mine, not yours. I'll sweep the user guide and knowledge base for anything pointing at Header Settings for Site Name Image, and update the internal platform admin docs once this lands.
- The accessibility AC just needs a confirmation that heading order still reads correctly with the new sections on the page. Low risk since nothing custom is being built, but the ticket asks us to check rather than assume.

## What I'll be testing hardest

Your "the regression that matters" step is the right one and it's what I'll hit first. A site admin opening Header Settings, changing nothing, saving, and doing the same on Site Settings, then confirming branding and the environment indicator are untouched. That's the one failure mode here that silently destroys live data.

Second is the render cache flush. I want to upload a Site Name Image, save, and see the header change without a `drush cr`. Good catch that the flush is load-bearing rather than housekeeping, and that `CoreTwigExtension` bubbling no cacheability is what makes it so. The conditional flush is the right call given this page has one Save button for everything, but it's worth confirming live that the "changed" comparison actually fires on a file upload.

miketullo95:
Correcting one thing from my review: I said I'd sweep the user guide and knowledge base and update internal platform admin docs. We're dropping that.

Every setting in scope here is platform-admin-only, so nothing editor-facing on yalesites.yale.edu covers them, and the audience for the internal doc is the same small group already testing this. Writing it would tell them what they just tested.

I've descoped the Documentation section on #1560 accordingly. The only thing left there is confirming no Header Settings screenshot in the user guide was taken while logged in as a platform admin, since one of those would show the Site Branding sections ordinary editors never see. That's mine and it doesn't affect this PR.

The announcements source move I asked for is still the one blocking item.

dblanken-yale:
@miketullo95 Thanks — the announcements source move is done, and both sign-offs are recorded below.

**Leaving this at `needs work` rather than flipping it to `needs review`, deliberately.** Decisions 1 and 2 still need @dblanken-yale's eyes as lead developer, and that is not something this pass can sign off on its own behalf. Everything else you asked for is addressed; the label is the only thing waiting.

## Awaiting lead-developer sign-off (the one open item)

You were right to separate these out as architecture calls rather than product calls, so they are unchanged and unmoved pending review:

1. **One gating mechanism** — every "is this a platform admin" check consolidated onto `hasPermission('administer platform admin settings')`, exposed as `PlatformAdminSettingInterface::PERMISSION`. Touches `HeaderSettingsForm`, `SiteSettingsForm`, `DashboardSettingsForm`, `ys_localist`'s `LocalistSettings`, and deletes `ys_core_allow_secret_items()`.
2. **Hidden over disabled** — moved fields are removed from the mixed-audience form rather than rendered read-only, matching your "no read-only mirrors" call below.

To clear this: a comment on this PR signing off (or asking for a different shape) is all it needs — no code change either way. Once that lands the label can go to `needs review`.

## Blocking item — done

**Announcements source moved to Platform Admin Settings.**

New `AnnouncementsSourcePlatformAdminSetting` plugin at weight `20`, so it renders directly beneath the existing **Dashboard Announcements Feed** section (weight `10`) — the two announcements concerns are now adjacent on one page instead of split across two.

Followed the same pattern as the four already moved, as you asked:

- The `#access` gate and the `submitForm()` conditional are **deleted, not relocated**. The route permission gates the whole page, so the plugin writes unconditionally and Dashboard Settings simply never touches those keys.
- `DashboardSettingsForm::isPlatformAdmin()` had zero callers left afterwards and is **deleted**, the same fate as `ys_core_allow_secret_items()`.
- The form's now-orphaned `EntityTypeManagerInterface` dependency came out too — it only existed to serve `ensureAnnouncementTerm()`, which moved with the fields.
- **Config keys unchanged** (`ys_core.dashboard_settings:announcements_source_enabled` / `announcements_source_term`), so no update hook and no migration. `AnnouncementsFeedController` reads exactly what it read before.

Things worth flagging that came out of doing it:

- **The `#states` selector had to change.** The page renders every section with `#tree` set, so the checkbox's input name is now `announcements_source[announcements_source_enabled]`. The un-namespaced selector the field carried on Dashboard Settings would have silently never matched, and the tag field would have shown unconditionally. There is a test pinning the selector so a future rename cannot quietly break it back.
- **Added the same no-change short-circuit the other sections have.** One Save button submits every section, so without it a save of an unrelated section would rewrite these keys, invalidate `config:ys_core.dashboard_settings`, and drop the cached feed — forcing an HTTP refetch on the next toolbar badge render. Same reasoning as the branding flush.
- **The tag check deliberately runs *ahead* of that short-circuit**, and on every save while publishing is on. The tag is the one thing on this section another actor can delete out from under the config, and the old form was self-healing only as a side effect of re-running on every platform-admin save. Putting the check behind the short-circuit would have meant a deleted tag was never restored and the endpoint quietly answered 200 with zero items. Verified on the running site: enabled publishing, deleted the tag from the Tags vocabulary, re-saved the page **without changing anything**, and the tag came back.
- **A cleared tag field now stores `Dashboard Announcement` rather than an empty string.** An empty value left `AnnouncementsFeedController` falling back to a default name that nothing guaranteed existed — same silent-empty-feed failure, reached a different way.
- **An existing but *unpublished* tag now warns.** The existence lookup is deliberately not access-checked (an access-filtered one that could not see an existing term would create a duplicate), but the feed's query is — so an unpublished term read as "already exists" here and was excluded there, with nothing said to the admin.
- **`ensureAnnouncementTerm()` now uses the injected term storage** instead of the static `Term::create()` it used on the form. Same behaviour, but it goes through DI rather than the global container, which is what makes the tag-creation path unit-testable.

**Test coverage:** `AnnouncementsSourcePlatformAdminSettingTest` — build, default tag fallback, the `#states` selector, both keys written, a tag rename with the toggle unchanged, blank-tag normalization, the no-change short-circuit, tag created on enable, no tag created on disable, a deleted tag recreated when nothing changed, and the unpublished-tag warning. Plus a rewritten `DashboardSettingsFormTest` whose regression guard is that a save **never writes the moved keys**, even when they are injected into the submitted values as a crafted POST would — the same shape as `HeaderSettingsFormTest`. The kernel `PlatformAdminSettingsFormTest` asserts the new section renders and pins the four-section order.

`ys_core` unit went 254 -> 266 passing; kernel is green at 41. `composer code-sniff` is clean.

## Sign-offs — recorded, no code change

- **CAS Application Name stays.** Left `id() == 1` with the comment explaining it is intentionally narrower.
- **Localist fields stay.** Mechanism consolidated, fields not relocated.
- **No read-only mirrors.** Clean removal everywhere; nothing left behind on Header Settings, Site Settings, or Dashboard Settings to say "someone else controls this".

## Notes

- **Duplicated follow-up list:** fixed. The two Localist bugs appear once now, under "Follow-ups I did not do here."
- **Accessibility / heading order:** confirmed on the running site, unaffected. All five sections are `<summary>` inside `<details>` (Claro's own pattern), not headings, so the content region's sequence is still just `h1 Platform Admin Settings` followed by the status/error message regions. Adding a section introduces no heading, so it cannot reorder anything.
- **Documentation:** understood, descoped per your follow-up comment. Nothing done here.
- **Process note:** taken. The inventory-then-implement gap was mine; on a change where the gate matters, the inventory goes up and waits.
- **Follow-up tickets:** all yours, thanks. Nothing filed from this end so we don't end up with duplicates.

Two things I did **not** change, for the record. The form's `parent::__construct($config_factory)` omits `$typedConfigManager` (deprecated in 10.2, required in 11) — but no `ys_*` form in the codebase passes it, so fixing this one alone would be inconsistent; it belongs with the other deprecation already listed as a follow-up. And `buildSettings()` reads override-aware while `submitSettings()` compares override-free, so a `settings.php` override would be written into real config on save — that is identical in `EnvironmentIndicator` and `AnnouncementsFeed`, so it is a page-wide pattern rather than something this change introduced.

The PR description has been updated: the source move is folded into "Moved to the Platform Admin Settings page", "Deliberately not moved" now reflects your sign-offs, the duplicate list is gone, and there are new **Functional testing steps** covering the move — including the site-admin-saves-and-nothing-is-clobbered regression check for this section.

dblanken-yale:
@miketullo95 Answering the one verification you asked for, plus a defect that answering it turned up.

**Still at `needs work` deliberately, for the same reason as before.** Decisions 1 and 2 remain unsigned-off by our lead developer, and that is not something this pass can sign off on its own behalf. Nothing else is outstanding; the label is the only thing waiting.

## The thing you asked me to confirm

> it's worth confirming live that the "changed" comparison actually fires on a file upload

**It does.** Confirmed on a running site by driving the real plugin against real config, the real media manager and a real uploaded SVG: the upload wrote `site_name_image` and fired exactly one `invalidateAll()`, the file was marked permanent, and on the next request the header rendered the new lockup with its `<title>` swapped to the site name — no `drush cr` in between.

**But confirming it exposed the converse: the comparison fired on *every* save, so the conditional flush was not actually conditional.**

Every key involved reads `NULL` on a real site — `ys_core.header_settings` does not exist as a config object at all until something writes it. The comparison was `$submitted === $current` against raw config, and the two sides hold the same values in different shapes:

| | stored | submitted |
|---|---|---|
| `site_name_image` | absent, or `''` as shipped in `config/install` | `[]` from the `managed_file` element |
| `site_wide_branding_name` / `_link` | absent | the fallback string the form just displayed |
| `environment_indicator.show` | absent, or integer `1`/`0` from the old `SiteSettingsForm` | a cast bool |

So a platform admin who touched only the Beacon section still wrote three branding keys and flushed the **whole site's** render cache — the exact penalty the guard exists to avoid, and the reason you were right to ask.

Scope, stated honestly: it is **self-healing after one save** (the new code writes normalized values, so the second save onward short-circuits correctly), and it is **not a regression** — the old `HeaderSettingsForm` flushed unconditionally on every save. So this is one spurious sitewide flush per site, not a permanent leak. I fixed it anyway rather than just reporting it: it is the mechanism you asked about, the fix is a few lines, and the sibling `AnnouncementsSource` / `AnnouncementsFeed` plugins already cast both sides — so this also makes the four plugins consistent instead of two-of-four.

**Both sides now normalize before comparing.** Zero behavioural risk, checked before touching anything: an unwritten key renders identically to a stored default, because the site-header component already applies `|default('Yale University')` / `|default('https://www.yale.edu')` and `ys_core.module` reads `environment_indicator.show ?? TRUE`. Config keys and stored values are unchanged, so still no update hook.

While in there, `EnvironmentIndicatorPlatformAdminSetting::DISPLAY_DEFAULT` was documenting itself as the shared source of truth for that default while `ys_core_preprocess_html()` and `ys_core_page_top()` both still hardcoded `?? TRUE` — three copies, one of them claiming to be canonical. Both hooks now read through the constant, so the render path cannot drift from the form.

## Evidence

Local verification, in order: untouched save -> **0 flushes, nothing written**; SVG upload -> **1 flush**, config written, file made permanent; header renders the new lockup in a fresh request; resave unchanged -> **0 flushes**; image removal -> flush fired and the file was deleted. (The site state was restored afterwards.)

One honest caveat: the local render bin is `cache.backend.null`, so a flush is not observable by inspecting the cache here — the counts above come from a spy on the cache backend while everything else in the path was real. The flush itself is also pinned by unit tests on both branches of the condition.

Tests (scoped to what this change touches, with a baseline captured first): `SiteBrandingPlatformAdminSettingTest` 7 -> 9 passing, `EnvironmentIndicatorPlatformAdminSettingTest` 4 -> 6, `SiteSettingsFormTest` 2/2 unchanged, kernel `PlatformAdminSettingsFormTest` 4/4. Each new test was confirmed failing for the right reason before the fix. One of them pins your file-upload case specifically, so it cannot silently regress:
- `testSubmitFlushesRenderCacheWhenAnImageIsUploadedOverUnsetConfig` — the guard must not swallow the first upload.
- `testSubmitSkipsWriteAndCacheFlushWhenUnsetBrandingIsResubmitted` — an untouched save must not flush.
- `testSubmitSkipsWriteWhenUnsetValueIsResubmittedAsShown` / `...WhenStoredIntegerMatchesSubmission` — the env indicator's absent-key and legacy-integer shapes.

`composer code-sniff` exit 0 (both standards). One gap worth naming: my automated code-review pass on this diff did not complete (the reviewer never returned), so this change has had the quality, security and standards passes but not that one. It is five files and about 40 lines of real change, so a quick human read of the guard logic would close that off.

## One thing worth handing to the lead-developer review

Since Decisions 1 and 2 are the open item, here is the evidence a reviewer would want on the part
of them that actually carries risk — whether swapping the `platform_admin` role check for a
permission widened access. It did not; it narrowed it:

- `administer platform admin settings` is declared with `restrict access: true`, and across all
  seven shipped roles (`anonymous`, `authenticated`, `contributor`, `editor`, `file_manager`,
  `platform_admin`, `site_admin`) it is granted in exactly one — `user.role.platform_admin.yml`.
  No role sets `is_admin` for a blanket bypass, and nothing grants it programmatically or in an
  install/update hook. So the grantee set is `{platform_admin}` + user 1, which is precisely what
  the deleted role check matched.
- The forms these fields came *off* only require `yalesites manage settings`, which **contributor,
  editor, site_admin and platform_admin** all hold. So the moved fields used to sit on pages four
  roles could open, kept out of reach by an `#access` flag plus a submit-handler `if`; they now sit
  behind a route permission one role holds. That is the strongest argument for the "hidden over
  disabled / delete the guard" call: the guard was protecting fields on a page the wrong audience
  could reach, and the move removes that condition rather than papering over it.
- Confirmed the guards are genuinely gone rather than merely hidden: grepping `ys_core/src/Form/`
  for `site_name_image`, `site_wide_branding`, `announcements_source` and `environment_indicator`
  returns zero hits, none of the three forms declares `#config_target`, and `ConfigFormBase` does
  no blanket value-to-config copy — so a crafted POST to the original routes carrying those keys
  is inert.

## Your other points

- **The regression that matters** (site admin saves Header/Site settings, branding and the indicator survive): unchanged, still covered by `HeaderSettingsFormTest` / `SiteSettingsFormTest`.
- **Sign-offs** (CAS stays, Localist stays, no read-only mirrors): recorded, no code change.
- **Announcements source move:** done in the previous push; re-verified here — the plugin exists, `announcements_source` appears nowhere in `DashboardSettingsForm`, and `isPlatformAdmin()` / `ys_core_allow_secret_items()` are both genuinely deleted with zero callers. Plugin discovery on a running site returns the five sections in the advertised order (Site Branding, Environment Indicator, Beacon, Dashboard Announcements Feed, Announcements Source).
- **Accessibility:** unaffected by this change — it renders no markup. Heading order answer stands: the sections are `<summary>`/`<details>`, so they add no headings.
- **Documentation:** descoped per your follow-up, nothing done.

## Two follow-ups worth adding to your list

Both came out of reviewing this fix, and both are consistency rather than bug work:

1. **A shared unchanged-guard contract on `PlatformAdminSettingBase`.** Four plugins hand-roll this guard and two of them just had shape bugs in it. Worth noting for whoever picks it up: it must *not* be lifted to the form, because `AnnouncementsSource` deliberately runs `ensureAnnouncementTerm()` **ahead** of its own guard (that is the deleted-tag self-heal) and `AnnouncementsFeed` normalizes the submitted URL before comparing — a generic form-level "section unchanged, skip submit" would silently kill both.
2. **Per-section save buttons instead of one shared Save.** This is the root cause; it would delete all four guards rather than tidy them. Needs your input, since it changes the page's interaction model.

Nothing in this push touches the companion repos — the change is yalesites-project only.

dblanken-yale:
Correcting something in my comment above, and one more commit.

**The code-review pass I said had not completed did complete** — it was slow, not dead, and I called it too early. Retracting that caveat. It also earned its keep, so the honest version is better than what I posted: it ran both phpcs standards with the cache disabled, built a throwaway probe harness to drive both submit handlers through the shape matrix rather than reasoning about it, and independently confirmed the render-identical premise by tracing `site_header__branding_name` to its single consumer in `_site-header--secondary.twig` and verifying `|default()` matches the PHP constants character for character.

It found a real gap, now fixed in `501491125`.

**The guard fix had tests for only one direction on the environment indicator.** Site branding paired its "untouched save must not flush" test with "a first upload must still write and flush" — the must-not-swallow-a-real-change side. The environment indicator got two skip tests and no counterpart, so nothing pinned **stored unset + admin unticks the box must write FALSE**.

That is the case that matters most, because the shipped `ys_core.site` omits `environment_indicator.show` entirely — so on a real site "unset" is the *normal* state, not an edge case, and it is exactly where normalizing the stored side to the display default could plausibly swallow the change. The behaviour is correct today (verified), but a future edit to the guard — simplifying it to `!== TRUE`, or reordering the `??` — would silently remove a platform admin's ability to turn the banner off, with every existing test still green. Now it cannot.

Added `testSubmitWritesWhenUnsetValueIsUnticked` and `testSubmitWritesWhenStoredLegacyIntegerIsTicked` (the other half of the legacy-integer shape: stored `0` plus a tick must write `TRUE`). That file is 6 -> 8 passing; `composer code-sniff` still exit 0.

**One thing I deliberately did not change,** since the reviewer raised it and explicitly would not block on it: `array_filter` preserves the fid's type, so a config value of `['12']` against a submitted `[12]` would compare unequal and the guard would not fire. Both sides originate from the same `File::id()` call so the types always agree, core flattens the element to a list of integers, and the failure direction is a spurious write rather than data loss. Adding an `intval` map would guard a shape that cannot occur, which is the same reason I dropped a redundant `array_values()` from this fix earlier. Happy to add it if you would rather have the belt and braces.

**For whoever reviews a future site-header change:** `SiteBrandingPlatformAdminSetting::DISPLAY_DEFAULTS` and the `|default()` filters in `component-library-twig/components/03-organisms/site-header/_site-header--secondary.twig` must stay in sync. Twig cannot read a PHP constant across repos, so the duplication is structural. Drift on the PHP side is caught by tests here; drift on the CLT side is not, and cannot be from this repo. The constant's docblock names the file so it is greppable from either end.

dblanken-yale:
@miketullo95 No new findings to work — every item from your review is addressed and re-verified. This push does two things: brings the branch up to date with `develop`, and puts the evidence for the one open item in front of whoever reviews it.

**Still at `needs work`, same reason as the last two pushes.** Decisions 1 and 2 need our lead developer's eyes, and that is not something this pass can grant on its own behalf — it would be signing off on itself. That is the only thing outstanding. To clear it: a comment on this PR either signing off or asking for a different shape. No code change either way.

What I did instead was stop asserting those decisions are safe and go **independently verify** them against the code, adversarially, so the sign-off is a one-read decision rather than a fresh investigation.

## Decisions 1 and 2 — independent verification

All seven claims I made earlier hold up. The ones that actually carry risk, with evidence:

- **`administer platform admin settings` is granted in exactly one role.** Declared `restrict access: true` (`ys_core.permissions.yml:6-9`). A repo-wide grep for the literal string returns six hits: the declaration, the route requirement, the interface constant, two test docblocks, and exactly one grant — `config/sync/user.role.platform_admin.yml:149`. All seven shipped roles enumerated; none sets `is_admin: true` (`platform_admin.yml:121` is `is_admin: null`). No programmatic grants anywhere in the custom tree.
- **The escalation path is closed too**, which I had not checked before: only `platform_admin` holds `assign platform_admin role` (`platform_admin.yml:157`), `site_admin` can assign only contributor/editor/site_admin (`site_admin.yml:132-134`), and no role holds `administer permissions`. So a site admin cannot grant themselves in.
- **The permission check is exactly equivalent to the deleted role check.** The deleted body was `id() == 1 || in_array('platform_admin', ...)` (`git show origin/develop:.../ys_core.module`, lines 1179-1187) — `{uid 1} ∪ {platform_admin}`, which is precisely what the permission resolves to given the grant analysis above. `DashboardSettingsForm::isPlatformAdmin()` was character-for-character the same predicate.
- **The strongest argument for Decision 2, quantified.** The three forms these fields came off require only `yalesites manage settings`, held by **contributor** (`contributor.yml:470`), **editor** (`:493`), **site_admin** (`:511`) and **platform_admin** (`:553`). So the deleted in-form gate was the *only* thing standing between a contributor and platform-admin-owned config on those pages. The move replaces an `#access` flag plus a submit-handler `if` with a route permission one role holds.
- **A crafted POST to the original routes is inert — confirmed, not assumed.** Zero hits for any of the six moved keys across the three `src/Form` files, zero `#config_target` anywhere in `ys_core/src/Form/`, and on Drupal 10.6.13 `ConfigFormBase::submitForm()` (`web/core/lib/Drupal/Core/Form/ConfigFormBase.php:305-311`) and `copyFormValuesToConfig()` (`:330-341`) both iterate *only* the `#config_target` map built at `:169`. Empty map means the parent writes nothing. Also checked no `hook_form_alter` in the custom tree re-adds the elements.

**Access verdict: equivalent. No widening, no regression.** No route, form, plugin or permission is reachable by a role that could not reach it before.

**One correction to my own framing, though.** "Four checks became one" is true for settings forms and overstated for the tree. Two raw `platform_admin` role checks remain in `ys_core.module` — `:681` and `:1518` — both UI-hiding for the Add-user menu item and local action rather than access control, so consolidating them is genuinely out of scope. But five platform-admin/uid-1 predicates still exist, not one, and a lead reading that sentence should know it before signing.

## The thing nobody had checked, and it looked fatal

`config/sync/ys_core.dashboard_settings.yml` ships `announcements_source_enabled: false` and `announcements_source_term: 'Dashboard Announcement'`. On the face of it `drush deploy`'s `config:import` would reset the platform site's publish flag on every release.

It does not — but only because `config_ignore` is enabled (`core.extension.yml:37`) in `mode: simple` with `'ys_core*'` in `ignored_config_entities`, and the sole carve-out (`~ys_core.integration_settings`) holds none of these keys. Branding and the environment indicator are doubly safe: `ys_core.site.yml` in sync omits `environment_indicator` entirely and `ys_core.header_settings.yml` is not in sync at all.

Safe, then, but **the protection rests entirely on one `config_ignore` glob**, and nothing in this PR says so. Naming it here so it is on the record rather than implicit — if that glob ever tightens, this section is the thing that breaks, and it would break quietly.

## Five things worth knowing, none blocking

Ordered by how much I would care:

1. **`SiteBranding::normalize()` compares fid arrays with `===`.** `SiteBrandingPlatformAdminSetting.php:231-232` does `array_filter((array) $value)` and relies on strict array comparison, which requires matching value types. It holds today — `ManagedFile` produces ints (`ManagedFile.php:99`), config round-trips ints — but this is exactly the bug class `339fb40ea` just fixed for the sibling keys, and the new tests use `[7]`/`[9]` on both sides so they cannot catch a regression. This is your "belt and braces" point arriving from a different direction, so I want to flag rather than re-decline it: **say the word and I will add `array_map('intval', ...)` inside `normalize()`**, which makes it structurally immune instead of immune-by-argument. Worst case today is a redundant write plus one sitewide flush, self-healing after a save.
2. **The unpublished-tag warning re-fires on unrelated saves.** `AnnouncementsSourcePlatformAdminSetting.php:218-220` — because `ensureAnnouncementTerm()` deliberately runs ahead of the short-circuit (that is the deleted-tag self-heal, and it should stay), a platform admin who changes only the Beacon section sees a Tags warning on every save while the term is unpublished, plus two entity lookups per save regardless. Gating just the *warning* on a real change would fix it without touching the self-heal. This one I would actually like to fix; it is a small live UX wart in code this PR added.
3. **There are five plugins on this page, not four.** `ys_beacon`'s `BeaconPlatformAdminSetting.php:258-295` writes `ys_beacon.settings` unconditionally and, when authorized and chat is on, calls `enableIndex()` — an Azure-facing verify/provision — on *every* save of the page, including a save that only changed the branding name. Pre-existing and documented as intentional re-verification, so not a regression. But my "the guards are now consistent" line is only true if you exclude it.
4. **`PlatformAdminSettingsForm extends FormBase`, not `ConfigFormBase`** (`:19`), so the page gets neither the "overridden in settings.php" warning nor typed-config validation on save. Small in practice — only `ys_core.dashboard_settings` has a schema at all, and the only `environment_indicator` override in `settings.php:112` targets contrib `environment_indicator.indicator`, not `ys_core.site`. Design note, pre-existing.
5. **`ys_localist` has a live instance of the exact bug this PR set out to kill.** `LocalistSettings.php:193` writes `localist_group` unconditionally, but the element only exists when sync is enabled and groups are imported (`:135-147`) — so saving outside that condition NULLs a stored group. Already on the follow-up list, but worth noting it means Decision 2 got applied to `ys_core`'s forms and not to `LocalistSettings`, which still uses disabled-not-removed. (Its other platform-admin fields are safe from tampering, incidentally: Drupal ignores input for `#disabled` elements — `FormBuilder.php:1224`.) Also a leftover cosmetic: the local variable is still named `$allowSecretItems` after the function it was named for was deleted.

The four unchanged-guards I own all check out — `EnvironmentIndicator:77`, `AnnouncementsFeed:116`, `AnnouncementsSource:181-183` and `SiteBranding:199`+`normalize():224-236` each normalize both sides for every shape I could produce, including the absent-key and legacy-integer cases. And the premise the skip-the-write optimisation rests on was verified end to end: `region--header.html.twig:65-66` passes the values through to `_site-header--secondary.twig:5-6`, whose `|default('Yale University')` / `|default('https://www.yale.edu')` match `SiteBrandingPlatformAdminSetting::DISPLAY_DEFAULTS` exactly.

## Housekeeping

- **Branch is current with `develop` again** (`78c10dc15`). Clean merge, zero conflicts — the three incoming commits are all unrelated `ys_beacon` PDF-extraction work with no overlap on `ys_core` or `ys_localist`.
- Board status for yalesites-org/YaleSites-Internal#1560 is already **In review**; left as is.
- Nothing in this push touches the companion repos.

**Honest limit on this pass:** the shared Lando environment was held by another run, so this push has had `php -l` (clean on all changed files) and the static verification above, but **not** `composer code-sniff` or PHPUnit. Both were exit 0 / green as of `501491125`, and this push adds only a conflict-free merge of already-merged `develop` commits, so I would expect no change — but CI on this PR is the authority, not me, and I would rather say that than imply a gate I did not run.

miketullo95:
Heads up on rebase fallout from yalesites-org/yalesites-project#1500 (Site Settings grouping, closes #1590), which I just approved and is ready to merge — it touches the exact same file this PR does, `SiteSettingsForm.php`, and this PR's diff here predates it (branched 8/18, #1500 opened 8/22).

What #1500 changed there: it restructured `buildForm()`/`submitForm()` into vertical tabs, and nested both `cas_app_name` and `environment_indicator_show` inside a new `$form['advanced']` group, gated at the group level by `'#access' => $allow_secret_items` where `$allow_secret_items = ys_core_allow_secret_items($this->currentUserSession)`. It also kept the `$currentUserSession` property/constructor assignment this PR removes as dead.

Concretely, once you rebase on develop after #1500 lands:

1. **`$currentUserSession`** — this PR deletes the property; #1500 adds a new usage of it (`ys_core_allow_secret_items($this->currentUserSession)` gating the Advanced group). Taking this PR's deletion as-is would leave that call broken.
2. **`environment_indicator_show`** — this PR removes it from a flat top-level field; on develop post-#1500 it'll be nested inside `$form['advanced']` instead, so the removal needs to come from there.
3. **`ys_core_allow_secret_items()`** — this PR deletes it citing zero remaining callers, but #1500 adds one (the Advanced group's `#access`). Worth confirming that's fully migrated before the delete goes in.
4. Once `environment_indicator_show` is out, `$form['advanced']` holds only `cas_app_name` — gated narrower (`$is_user_1`) than the group's own `#access` (`$allow_secret_items`). That reproduces the "empty-looking tab for a platform_admin who isn't user 1" case #1590's AC was written to avoid. Worth a look at whether Advanced still makes sense as its own tab at that point, or if `cas_app_name` should land somewhere else on the form — your call on the cleanest shape.
5. `SiteSettingsFormGroupingTest.php` (new in #1500) has assertions tied to `environment_indicator_show` living in the `advanced` group — those will need updating too.

Header Settings and the branding fields aren't affected — #1500 only touched `SiteSettingsForm.php`, so that part of this PR should rebase clean. Just flagging so this doesn't surface as a confusing merge conflict later. Let me know if you'd rather sequence these differently.

dblanken-yale:
@miketullo95 Thanks for the heads-up on #1500 — it saved a confusing merge later. All five items are worked, `develop` (with #1500 in it) is merged into this branch, and the conflict is resolved rather than deferred.

**Still at `needs work`, same single reason as the last four pushes.** Decisions 1 and 2 still need our lead developer's eyes, and that is not something this pass can grant on its own behalf. That remains the only outstanding item. To clear it: a comment on this PR either signing off or asking for a different shape. No code change either way.

## Your five items

| # | Item | Resolution |
|---|---|---|
| 1 | `$currentUserSession` deleted here, #1500 adds a usage | The usage goes away with item 3, so the deletion stands. Verified: zero references to the property in the merged tree. |
| 2 | `environment_indicator_show` is now nested in `$form['advanced']` | Removal taken from the new location, not the old flat one. |
| 3 | `ys_core_allow_secret_items()` — confirm the migration is complete before deleting | Confirmed after resolution: `grep` over the whole custom tree returns zero callers. The only remaining mentions are two test docblocks describing what the code used to do. |
| 4 | Advanced left holding only user-1-only `cas_app_name` | Group gate narrowed from `$allow_secret_items` to `$is_user_1`. Details below — this is the one where you left the shape to me. |
| 5 | `SiteSettingsFormGroupingTest` assertions tied to the moved field | Updated, plus a new test pinning item 4's decision. |

## Item 4 — the shape I picked, and the one I did not

`$form['advanced']['#access']` is now `$is_user_1`, matching its only remaining child.

The reason is the comment #1500 wrote on that very group: *"an Advanced tab a user cannot open anything inside is worse than no tab at all."* Narrowing the group to its last field's gate is the smallest change that keeps that property true once the environment indicator leaves. A platform admin who is not user 1 now sees no Advanced tab at all, rather than an empty one — and loses nothing, because the setting they actually own moved to a page they can reach.

It also resolves items 1 and 3 as a side effect rather than by argument: with `$allow_secret_items` gone, `ys_core_allow_secret_items()` genuinely has zero callers and `$currentUserSession` is genuinely dead — which is what this PR claimed before #1500 landed, now true again for the right reason.

**The alternative I considered and did not take:** relocating `cas_app_name` out of Advanced and dropping the tab. Rejected as scope creep — CAS is user-1-only by a decision already recorded and signed off on this PR, and Advanced is still an honest home for it. Say the word if you would rather have the tab gone; it is a small change and I would rather you pick than assume.

## What the merge actually changed here

`SiteSettingsForm.php` was the only conflicted file. The resolution keeps every bit of #1500's structure — the six vertical tabs, the plain-language labels, the teaser fallback, the error-opens-its-own-group behaviour — and applies this PR's removal on top of it. Net diff against post-#1500 `develop` is: the `$currentUserSession` property and its assignment, the `$allow_secret_items` variable, the `environment_indicator_show` element, the `environment_indicator.show` write in `submitForm()`, and the group's `#access` expression. Nothing else.

`ys_core.module` merged cleanly and kept the `ys_core_allow_secret_items()` deletion.

## Tests

`SiteSettingsFormGroupingTest` (yours, from #1500):

- `EXPECTED_MEMBERSHIP` drops `environment_indicator_show`, so the "set or order of settings changed" guard now describes the post-move form.
- `testAdvancedGroupIsVisibleToUser1` asserts the environment indicator is **absent** rather than present, with a docblock saying why absent and not merely hidden: a hidden element still submits its default, which would overwrite what a platform admin set on the other page.
- **New — `testAdvancedGroupIsHiddenFromPlatformAdminsWhoAreNotUser1`.** This is the case your item 4 named, and nothing pinned it: it builds the form as a real platform admin who is not user 1 and asserts both the group and the CAS field are inaccessible. Before the gate change the group would have been accessible and empty.
- `testAdvancedGroupIsHiddenWhenItWouldBeEmpty` keeps its uid-2 case unchanged.
- The submitted-values fixture in `testSubmittingStillWritesEveryValueToItsOwnConfigKey` drops the moved key. The crafted-POST regression — this form must never write `environment_indicator.show` even when the value is injected — is already pinned by `SiteSettingsFormTest::testSubmitDoesNotWriteMovedEnvironmentIndicatorKey`, so it is not duplicated here.

Also worth noting the merge brought in the `ys_media` extraction (#1475), which moved `YaleSitesMediaManagerTest` out of `ys_core` — so the `ys_core` Unit count is a smaller number than the one I quoted on 08-21 for reasons unrelated to this change. Baseline was recaptured post-merge rather than compared against the old figure.

## Verification

- `SiteSettingsFormGroupingTest` (yours, from #1500): **10 tests, 192 assertions, green.**
- `ys_core` Unit: **258 tests, 333 assertions, 1 skipped, 0 failures** — identical to the baseline captured after the merge. Note that number is lower than the 270 I quoted on 08-21, and not because anything broke: the merge brought in the `ys_media` extraction (#1475), which moved `YaleSitesMediaManagerTest` out of `ys_core`. Different denominator, same green.
- `composer code-sniff`: **exit 0**, both standards over the whole custom tree, re-run after the final edit.
- The new test was confirmed **failing against develop's exact predicate** (`$is_user_1 || in_array('platform_admin', $this->currentUser->getRoles())`) before the change and passing after, so it genuinely discriminates rather than passing vacuously.

Two things worth naming rather than leaving in the run log:

**My first version of the new test was wrong, and the review pass caught it.** It built its "platform admin" with `setUpCurrentUser(['uid' => 3], [PERMISSION])`. Core's `UserCreationTrait` implements a permissions argument by inventing a role with a *random machine name*, and `createUser()` overwrites any `roles` you pass alongside it — so the account never held `platform_admin`, and develop's gate was a role check. The test would have passed against the old gate too, making it an expensive duplicate of the plain-user case above it. It now assigns the real role and asserts `hasPermission()` up front so the fixture cannot quietly decay that way again.

**A security pass over the diff found nothing**, and confirmed the thing worth confirming: this branch changes no role config at all (`git diff` over `config/sync/user.role.*.yml` is empty), `administer platform admin settings` is still granted in exactly one role, and a crafted POST to the stripped forms is inert because there is no `#config_target` anywhere in the custom tree, so `ConfigFormBase` writes nothing. It also independently reached your `site_wide_branding_link` point from the other direction — unvalidated URL rendered into an `href` — and classified it as relocated, not regressed, which matches the follow-up ticket you took.

## Your non-blocking notes, still standing

- The `intval` map on `SiteBranding::normalize()` is still not implemented — fourth time naming it. Unchanged reasoning: both sides originate from the same `File::id()` call, you would not block on it, and the failure direction is a spurious write. Still a one-line change if you want it.
- The duplicated description entry you spotted (the two Localist bugs listed twice) and the follow-up tickets are yours.

References yalesites-org/YaleSites-Internal#1560

dblanken-yale:
@miketullo95 Thanks for the heads-up on #1500 — you called it exactly right, and it was worth flagging rather than letting it surface as a merge conflict. #1500 landed on `develop` at 15:18 today, so this was live rather than hypothetical by the time I picked it up. All five items are addressed and the branch is merged clean with `develop`.

**Still at `needs work`, same single reason as the last four pushes.** Decisions 1 and 2 need our lead developer's eyes, and that is not something this pass can grant on its own behalf. That remains the only outstanding item. To clear it: a comment on this PR either signing off or asking for a different shape. No code change either way.

## Your five items

**1. `$currentUserSession`.** Worth being precise about, because "this PR deletes a property #1500 adds a usage of" reads worse than it is: `develop` declares **both** `$currentUser` and `$currentUserSession` and assigns *the same* `$account_interface` to each. The one this PR deletes is the duplicate, so the fix was to stop calling through it, not to put it back. Git had already resolved that half correctly on its own; only the call sites needed work.

**2. `environment_indicator_show` nested in `$form['advanced']`.** Removed from there rather than from the flat top level. Its `submitForm()` write auto-merged out correctly.

**3. `ys_core_allow_secret_items()`.** Its last two callers were the two above, so the delete is now genuinely caller-free — verified by grep across the whole custom tree, which returns zero code references (two historical mentions survive in test docblocks).

**4. The Advanced tab — the one real judgment call, and I did not take my first answer.** You are right that `$form['advanced']` ends up holding only `cas_app_name`, gated `uid == 1`, under a group gated wider — the empty-tab case #1590's AC exists to prevent. My first fix was to hardcode the group's gate to `$is_user_1` so it matched its one child. **That was the wrong depth and my own review pass talked me out of it.** Two things decided it:

- `FormBuilder::doBuildForm()` propagates `#access` **downward only**, so a hardcoded parent is the *duplicate* of the child's condition, and I had put the source of truth on the container instead of on the setting that actually has an audience.
- **This module already states the rule one level up, and I was contradicting it.** `ys_core.routing.yml` gives the Platform Admin section no permission of its own — "the menu-block check grants access when at least one child is reachable, so the section can never hide a screen the user may see, and it needs no edit when children change." The Advanced tab is the same container-vs-children problem one level down.

So the group's gate is now **derived**: `$form['advanced']['#access'] = (bool) Element::getVisibleChildren($form['advanced']);`. `cas_app_name` keeps its own `uid == 1` gate as the single source of truth. That is a core idiom, not an invention — core uses `(bool) Element::getVisibleChildren()` as an `#access` value in `shortcut`'s `SetCustomize::actions()`.

Two things I checked so the next person does not have to. Core's `'#optional'` on a details element is *the* idiom for this and **cannot** be used here: it counts `#groups` members, which only elements that *tag* `#group` join, and this form deliberately nests instead — it would find zero children and hide the tab from user 1 too. There is a comment saying so. And my first test locked in the wrong invariant: it asserted `#access` FALSE for a platform admin *unconditionally*, so adding a platform-admin-level setting here tomorrow would have left it green while certifying the defect. Both Advanced tests now assert the *reason* alongside the outcome, which means they start failing the moment a child its audience can see is added.

**Net effect, unchanged from what you asked for:** a platform admin who is not user 1 gets no Advanced tab; user 1 gets it with the CAS field in it; the environment indicator is one page away on Platform Admin Settings.

**5. `SiteSettingsFormGroupingTest.php`.** Updated: `environment_indicator_show` out of `EXPECTED_MEMBERSHIP` and out of the submit values, and `assertArrayHasKey` -> `assertArrayNotHasKey` in the user-1 test. New test `testAdvancedGroupIsHiddenFromPlatformAdminsWhoAreNotUser1` — a plain user fails both gates, so a platform admin who is not user 1 is the only account that can tell a user-1 gate from a platform-admin one. It creates the real `platform_admin` role and asserts `hasPermission()` up front, so the setup cannot quietly decay into a duplicate of the plain-user case. Both Advanced tests now also assert the group's *visible children*, so they pin the derived rule rather than today's answer.

One naming note, yours to call: `testAdvancedGroupIsHiddenWhenItWouldBeEmpty` is arguably a misnomer now — since the move, the tab is hidden from every non-user-1 whether or not it would be empty. I left your just-approved test's name alone rather than renaming it inside this merge, but say the word and I will.

## One thing you should know about this push

**Two passes worked this branch at the same time, and both pushed.** A second pass merged `develop` and updated the same test about twenty minutes before I got to my push, so `origin/1560-platform-admin-settings-audit` had moved when I re-synced. I merged its commits into mine rather than force-pushing over them — nothing was discarded, and both sets of commits are in the history. Where the two differed I kept the better half of each:

- **Its account setup for the platform-admin test, over mine.** It caught something I had wrong: `setUpCurrentUser()`'s permissions argument creates a role with a *random* machine name. That is fine for a permission check — and my precondition assertion did prove the permission resolved — but it would never satisfy a *role*-based gate, so a test written that way silently stops discriminating the moment the code under test checks a role. Creating the role under its real name is strictly stronger, so that is what is in the tree.
- **My derived gate, over its hardcoded `'#access' => $is_user_1`.** Same behaviour, better altitude, for the reasons in item 4.
- **Its assertion that the environment indicator is absent from the group**, which I had dropped as redundant. Its reasoning is better than my removal: a merely *hidden* element still submits its default, which would overwrite what a platform admin set. That is worth asserting explicitly.

The history is uglier than a single clean push, but it is honest and nothing is lost. Flagging it because it means the deduplication in our own tooling has a gap, not because anything here is wrong.

## Evidence

Test-first throughout, and each new test was confirmed failing for the right reason before the fix rather than after:

- **Item 5 first.** Before touching the tests, the merged branch failed `testEverySettingBelongsToItsIntendedGroup` on the missing `environment_indicator_show` — the fallout confirmed rather than assumed.
- **Item 4.** With the group left on the wider gate, the run was `Tests: 10, Assertions: 190, Failures: 1`, and the one failure was the new test: *"A platform admin who is not user 1 was offered an empty Advanced tab."* Every other test in the file passed, which is also what confirmed the item-5 edits independently of the item-4 fix.

Then verified on a running site rather than only in tests, driving the real form out of the container as real users with `account_switcher` — no mocks:

| account | roles | holds the permission | `advanced` `#access` | visible children |
|---|---|---|---|---|
| uid 1 | — (user 1) | yes, by super-user | `true` | `cas_app_name` |
| uid 2 | `platform_admin` | yes | `false` | **none** |

And in a browser, logged in as each:
- As the platform admin: tabs are Site basics / Key pages / Look and feel / Search and analytics / Content and tagging. **No Advanced tab**, no CAS field, no environment indicator field.
- As user 1: the same five **plus Advanced**, with the CAS field in it and no environment indicator field.
- Nothing was lost in the move: as the platform admin, Platform Admin Settings still renders **Site Branding, Environment Indicator, Beacon (AI Chat), Dashboard Announcements Feed, Announcements Source** in the advertised order, with the environment indicator checkbox present.
- #1500's own Platform Admin toolbar section renders for both accounts, so the merge composed correctly rather than merely compiling.

**Gates:** `composer code-sniff` exit 0 on the final merged tree (both standards, whole custom tree), and `composer lint:php` exit 0. `ys_core` Unit **258 tests, 333 assertions, 0 failures, 1 skipped**. `SiteSettingsFormGroupingTest` **10/10, 193 assertions** on the merged result. `npm run lint:styles` fails, but only inside `atomic/_yale-packages/tokens/build/scss/tokens.scss` - a generated file in my local atomic clone; this branch touches no SCSS at all, so that is pre-existing local noise rather than anything from here. Full `ys_core` kernel suite still not run end to end — it takes well over an hour on this machine and I would rather say that than imply a gate I did not run; CI is the authority.

## Two things my review passes turned up that you did not ask for

Naming them because neither came from you, and one of them removed code I had added an hour earlier.

1. **A guard test I added, then deleted.** Reviewing my own change caught that the "this form must not write the moved key" assertion I first added *could not fail* — the shipped default and the value the deleted write would have stored are bit-identical. I rewrote it to discriminate (seed the opposite, submit a conflict), proved it now fails when the deleted write is pasted back, and then deleted it anyway: this branch's own `SiteSettingsFormTest::testSubmitDoesNotWriteMovedEnvironmentIndicatorKey` already pins that regression, against a mock that records writes, which is the better instrument. Two tests for one invariant is paying twice.
2. **Two stale doc statements this PR itself caused.** `ys_core/README.md` and the grouping test's `strictConfigSchema` docblock both said `SiteSettingsForm` saves `environment_indicator.show` as integer `1`. It does not any more — the plugin stores a bool. Both reworded to attribute the integer to sites that saved *before* the move, which is the fact that still blocks the #579 schema work, so the point they were making survives. Documentation only.

## Two pre-existing security notes for your list, neither from this branch

A security pass over the diff came back clean — no widening, and a crafted POST to the three original routes is inert (no `#config_target` anywhere, so `ConfigFormBase` writes nothing, and each `submitForm()` enumerates its writes). It did surface two things worth tickets on their own merits, both unchanged by this work:

1. `editor`, `contributor` and `site_admin` all hold core's `administer users`, which in Drupal lets the holder set another account's password — including user 1's. That is a path to super-user, and therefore to every platform-admin permission. It was equally a path to the *old* `uid == 1 || platform_admin` check, so it is not a regression here, but it is worth knowing while we are tightening this area.
2. `SiteBrandingPlatformAdminSetting` validates SVG uploads with the legacy array validator form, and the SVG is later inlined into the page. Verbatim-unchanged from `develop` and platform-admin-gated, so not urgent, but it deserves its own look.

Nothing in this push touches the companion repos — the change is yalesites-project only.

miketullo95 (approved):
@dblanken-yale Approving.

Everything I asked for on the last pass is here. The announcements source section moved to the Platform Admin Settings page next to the feed section, the `#access` gate and the submit guards came out rather than getting relocated, and `DashboardSettingsForm::isPlatformAdmin()` is gone. Decisions 1 and 2 are signed off, and the consolidation is a real service now rather than a permission check open-coded at each call site, which also picked up the two `ys_core.module` role checks the earlier passes missed.

I checked out the multidev and the settings pages look right. The regression that matters holds: a site admin saving Header Settings and Site Settings with nothing changed leaves branding, the environment indicator and CAS exactly as they were. The Platform Admin Settings page reads top to bottom in the advertised order, and there is no Advanced tab on Site Settings for a platform admin who is not user 1.

A few things, all non-blocking and none of them gating this merge:

- The user-1 route gap (a hardened-off `security.enable_super_user` would 403 user 1 on the Platform Admin Settings route): leaving it as is. The docblock documents the limit and we do not turn that parameter off. No follow-up needed.
- The Localist save bugs, the `file_validate_extensions` deprecation, the `ys_beacon` reflection helper and the `ys_core.module:1204` `id() != 1` check: I am bundling those into one cleanup ticket and will file it.
- The two security notes (three roles hold core's `administer users` and can therefore set user 1's password, and the legacy SVG validator) are pre-existing and not from this branch. I will track them separately.
- `site_wide_branding_link` URL validation is mine to file, per my last review.
- Documentation is descoped on the ticket, so nothing is owed here.

Thanks for the depth on this one.

PR #1482, opened 2026-08-19
1564: Bug: Quick Links "between 3 and 9 links" error is reported on the Component Title field

dblanken-yale:
### Description of work

The Quick Links link-count rule was enforced in the browser by `ys_core/js/block-form.js`, which called `setCustomValidity()` on whichever input a hand-maintained selector list matched first. For Quick Links that list led with `edit-field-heading`, so "Number of links must be between 3 and 9. Number of links added: 1." landed on the **Component Title** while the Links field carried nothing — and because it was a native browser bubble rather than a Drupal form error, there was no visible error text, no `aria-invalid`, and nothing announced to a screen reader (WCAG 2.1 AA 3.3.1, Error Identification). Media Grid and Gallery led with the same selector and misplaced it the same way; only Tabs happened to point at the right field.

- Added a reusable `YsItemCountRange` validation constraint in `ys_core` that limits how many items a multi-value field may hold, counting only deltas the editor actually filled in.
- Attached it through a new `ys_core_entity_bundle_field_info_alter()` to the counted field of each limited block type — `quick_links.field_links` (3-9), `tabs.field_tabs` (2-5), `media_grid.field_media_grid_items` (2+), `gallery.field_gallery_items` (2+) — so the violation names the field the message is about and Drupal flags that field's widget.
- Deleted `js/block-form.js`, its `block_form` library entry, and both `#attached` sites.

Because the rule is now an entity constraint it runs on the standalone block form (`ContentEntityForm::validateForm()`) and in the Layout Builder path (core's `InlineBlock::blockValidate()`) through the same code path, rather than needing a second implementation for each.

Where field storage already caps the count — Tabs, at `cardinality: 5` — the cap is read off the field storage definition rather than restated in the rules table, so the message an editor reads cannot drift from what the widget actually allows. Only Quick Links' maximum of 9 is stated explicitly, because its storage cardinality is unlimited.

Core's `Count` constraint was considered first and rejected: its messages run through Symfony's plural machinery and cannot produce the single "must be between 3 and 9. Number of links added: 1." sentence editors already know, and it counts the spare empty row the link widget leaves for "Add another item".

### Worth @miketullo95's eye: editors leave the modal on a validation failure

Submitting a Quick Links block with too few links from Layout Builder moves the editor out of the off-canvas dialog and onto the full-page block configuration form, where the error is shown against the Links field. The JavaScript this PR removes kept them in the dialog.

**This is not new, and it is not specific to this constraint.** The Layout Builder configure-block submit carries no `#ajax`, so *any* server-side validation failure on that form leaves the dialog and re-renders as a standalone page. That is already documented in `ys_core_form_alter()` and was already handled once, in yalesites-org/YaleSites-Internal#1497, by adding the `status_messages` element so the rejection is visible when it happens.

What changes here is **how often editors meet that path** for these four block types: the count rule now fails server-side rather than in the browser.

| | Old (JavaScript) | New (entity constraint) |
|---|---|---|
| Where the editor ends up | Stays in the modal | Moved to full-page block form |
| Which field is blamed | Component Title (wrong) | Links (correct) |
| Screen reader announces it | No | Yes |
| Survives a non-JS / failed-JS save | No | Yes |

The error-placement bug in yalesites-org/YaleSites-Internal#1564 is fixed either way. If the modal exit is judged too disruptive at this frequency, the options are re-adding a client-side pre-check on top of the constraint, or putting `#ajax` on the configure-block submit — both are follow-on work, and both are larger than this fix. Flagging it rather than deciding it.

**Three things worth a reviewer's eye:**

1. This platform does not enable core's Inline Form Errors module, so no validation error anywhere on YaleSites renders an inline `aria-describedby` error node. This error now behaves exactly like every other one — visible text in the messages region plus `aria-invalid="true"` on the field — rather than getting bespoke treatment. Turning Inline Form Errors on is a platform-wide decision and out of scope here.
2. Removing the client-side check costs a server round trip, but not live feedback: the old JS ran on the submit button's click handler, not as you typed, so an editor already had to press Save to learn anything.
3. `field.field.block_content.media_grid.field_media_grid_items.yml` still says `required: false`, and Media Grid is the one bundle where a zero-item block used to save (the old JS only attached when at least one item was already on the page). It is now rejected with "Number of media grid items must be 2 or more." — which matches that field's own description, but means the config flag no longer describes reality. Left alone deliberately: whether Media Grid should be a required field is a product call, not part of an error-placement fix.

Editors with pre-existing out-of-range content cannot save that block until they fix the count. Worth auditing per environment before release — on a current pull of the platform site it is one Tabs block:

```sql
SELECT 'field_links' f, COUNT(*) n FROM (SELECT entity_id FROM block_content__field_links GROUP BY entity_id HAVING COUNT(*) < 3 OR COUNT(*) > 9) a
UNION ALL SELECT 'field_tabs', COUNT(*) FROM (SELECT entity_id FROM block_content__field_tabs GROUP BY entity_id HAVING COUNT(*) < 2 OR COUNT(*) > 5) b
UNION ALL SELECT 'field_media_grid_items', COUNT(*) FROM (SELECT entity_id FROM block_content__field_media_grid_items GROUP BY entity_id HAVING COUNT(*) < 2) c
UNION ALL SELECT 'field_gallery_items', COUNT(*) FROM (SELECT entity_id FROM block_content__field_gallery_items GROUP BY entity_id HAVING COUNT(*) < 2) d;
```

This is not new behaviour — the old JS attached whenever its item selector matched, so a one-tab block already refused to save — but it is now reported as a form error against the Tabs field instead of a browser bubble.

Covered by `BlockItemCountRangeConstraintTest` (7 kernel tests): below minimum and above maximum flagged on `field_links` with the exact message, in-range validates cleanly, the no-maximum message branch used by Media Grid and Gallery, empty deltas excluded from the count, every limited block type carrying the rule on its counted field with the right bounds, and the rule scoped to the bundle rather than the field name alone.

### Functional testing steps:
- [ ] Go to `/block/add/quick_links`, fill in the Component Title and exactly one link, and save. The error "Number of links must be between 3 and 9. Number of links added: 1." appears in the messages area at the top of the page, the Links field is outlined and marked invalid, and the Component Title field is **not** flagged.
- [ ] Add two more links and save again — the block saves.
- [ ] **Modal behaviour:** from a page layout, add a Quick Links block with a single link and click "Add block". Note where you end up — in the off-canvas dialog, or on the full block configuration page. This is the behaviour the decision above is about.
- [ ] Repeat on an existing Quick Links block via the Layout Builder pencil, and confirm removing links down to two produces the same error on the Links field.
- [ ] Add a Tabs block with a single tab and save — the error names Tabs, not the block's title, and reads "between 2 and 5".
- [ ] Add a Media Grid block and a Gallery block with a single item each and save — the errors name Media Grid Items / Gallery Items and read "2 or more".
- [ ] Delete the single item from a new Media Grid block and save — it is now rejected rather than saving empty. Confirm that is the wanted behaviour.
- [ ] Confirm no JavaScript errors in the browser console on any of the above (the old client-side validator has been removed).

References yalesites-org/YaleSites-Internal#1564


dblanken-yale:
Pushed two follow-up commits, plus a question for @miketullo95 in the description (hence the `question` label).

**`chore`** — removed `ys_core.libraries.yml.orig`, a merge-conflict artifact that had been committed. Inert, but it shipped a stale library definition still listing the deleted `js/block-form.js`. Added `*.orig` / `*.rej` to `.gitignore`, which is why it slipped past `git add` to begin with.

**`refactor`** — the four limited block types were an array literal in the middle of `ys_core_entity_bundle_field_info_alter()`. They now live in `BlockItemCountRules`, keyed by bundle, each returned as an `ItemCountRule` value object. Adding a limited block type is a one-line edit in an obvious file instead of surgery inside a hook.

Kept deliberately boring: the validating is still the `YsItemCountRange` constraint and the attaching is still `hook_entity_bundle_field_info_alter()` — the documented Drupal places for each. No custom plugin type, no tagged services; either would make someone new to this code learn bespoke machinery to express four rows of data.

Behaviour is unchanged: `BlockItemCountRangeConstraintTest` (7 kernel tests, 49 assertions) and the new `BlockItemCountRulesTest` (5 unit tests, 21 assertions) are green, and `composer code-sniff` exits 0.

Also corrected a claim in the original description and in the hook docblock: both said the error renders inside the off-canvas dialog. It does not — the configure-block submit carries no `#ajax`, so a server-side failure leaves the dialog and re-renders the form as a standalone page. That is pre-existing Layout Builder behaviour, already documented in `ys_core_form_alter()` from the yalesites-org/YaleSites-Internal#1497 work, not something introduced here. The question for Mike is whether meeting that path more often for these four block types is an acceptable trade for correct, announced, server-enforced validation.

miketullo95 (commented):
@dblanken-yale Not approving yet, and not because of anything in the diff. The multidev cannot currently exercise this PR's code, so the question you raised for me is unanswerable as things stand.

## The multidev is serving code this branch deletes

I tried to answer your modal-exit question and got the old browser bubble instead of a Drupal form error. That turned out not to be a testing mistake:

```
$ curl -sI 'https://pr-1482-yalesites-platform.pantheonsite.io/profiles/custom/yalesites_profile/modules/custom/ys_core/js/block-form.js?cb=…'
200   x-cache: MISS   age: 0
```

`git ls-tree -r --name-only` on this branch returns no `block-form.js` at any path, and the API reports it as `removed` in this PR. So the artifact is carrying a file the branch does not have, and it is not an edge-cache artifact.

It is also running, not merely present: the bubble reproduces on the multidev with the wrong-field placement this PR fixes. That means the library attach is happening too, which points at the custom module source on that environment being stale generally rather than just the one deleted file surviving. If that is right, `ItemCountRangeConstraint` is not there to test either.

I have established the symptom, not the cause. Diagnosing the Pantheon build step is yours rather than mine.

## It is not specific to this branch

Files deleted in already-merged work are still live on dev:

| File | Deleted from develop | Still served on dev |
| --- | --- | --- |
| `ys_embed/js/LibCal.js` | 2026-07-06 | 200 |
| `ys_embed/css/LibCal.css` | 2026-07-06 | 200 |
| `ys_core/js/layout-builder-fix-modals.js` | 2023-09-11 | 404 |
| `ys_core/js/node-edit-form.js` | 2023-04-13 | 404 |
| `ys_views_basic/js/ys_views_basic.js` | 2022-12-15 | 404 |

The 2023 deletions propagated and the July 2026 ones did not, so this regressed at some point rather than never having worked. LibCal is the pointed case: that PR existed to remove an unsupported embed source, and its JS and CSS have been sitting in the docroot for six weeks. Raised separately as yalesites-org/YaleSites-Internal#1583, since it affects every PR that removes a file and is not yours to carry on this ticket.

## On the code itself

Read it closely and it looks right. Things I checked rather than assumed:

The cardinalities the rules table depends on line up: `field_links` is unlimited so 9 is stated explicitly, `field_tabs` is 5 so the cap is read from storage, `field_media_grid_items` and `field_gallery_items` are unlimited with no maximum. Reading the cap off storage is doing real work on Tabs rather than being decorative, and `field_media_grid_items` is indeed the only one of the four with `required: false`.

On your `aria-describedby` note: the AC over-specified the mechanism, and I would not treat it as unmet. WCAG 3.3.1 asks that the erroring item be identified and the error described in text. Message-region text naming the field plus `aria-invalid` delivers both. Inline Form Errors would be a nicer association, not a compliance prerequisite.

Also worth saying: deleting the committed `ys_core.libraries.yml.orig`, conflict markers and all, and gitignoring `*.orig` / `*.rej` is a good catch that had nothing to do with your ticket.

Mike is happy with the Media Grid minimum, so no change wanted there. The `required: false` inconsistency you flagged is going into a separate cleanup rather than this PR.

## What I need before approving

A retest of the modal-exit behaviour against an environment actually running this branch. That is the one open decision, and I would rather answer it from observation than from either of our reasoning about it. Happy to look again as soon as there is somewhere to look.

PR #1483, opened 2026-08-19
1569: Bug: Collection top-page delete shows contrib "child pages will be relocated" above the YaleSites copy

dblanken-yale:
### Description of work

Deleting a content collection's top page from Manage Content showed two contradictory sentences about the same irreversible action. Contrib Book's warning came first -- "...the child pages will be relocated automatically" -- followed by the accurate YaleSites copy added in #1511: "The other pages in the collection won't be deleted; they'll stay published as standalone content." Only the second is true, because `ys_book_node_predelete()` takes the collection apart and leaves its pages as standalone content.

- `_ys_book_add_collection_delete_notice()` now unsets contrib's `book_warning` element as well as adding our own copy, so a collection's top page shows only the accurate message. Both happen in one function under one condition, so the two messages cannot get out of step with each other.
- Contrib's warning is left alone everywhere else. Deleting a page from the middle of a collection really does move its children up under its parent, so the contrib wording is correct there, and a new test pins that it stays.
- Extracted the "is this the collection's top page" test (`book['bid'] == nid`) into `_ys_book_is_collection_top()` and pointed `ys_book_node_predelete()` at it too -- that is the code that actually dismantles the collection, so the message and the behaviour it describes now share one definition instead of restating the rule separately.

The fix stays inside `ys_book` with no patch to contrib Book, matching the constraint #1511 worked under.

Two new Kernel tests in `BookCollectionDeleteTest`: one asserts the top page's confirm form has no `book_warning` but still has the YaleSites notice, the other asserts a mid-collection page still has `book_warning`. Full `ys_book` Kernel suite green (37 tests, 473 assertions); `composer code-sniff` exit 0.

### Functional testing steps:

- [ ] Create a content collection with a top page and at least two sub-pages, one of them nested two levels deep.
- [ ] From Manage Content, delete the collection's **top** page. Confirm the confirmation page shows only "This page is the top of a content collection... they'll stay published as standalone content", and no longer says the child pages will be relocated automatically.
- [ ] Cancel, then delete a page from the **middle** of the collection (one that has a child of its own). Confirm it still shows "...is part of a book outline, and has associated child pages. If you proceed with deletion, the child pages will be relocated automatically" -- that message is accurate here and must not have disappeared.
- [ ] Delete a leaf sub-page and confirm nothing changed: the plain "This action cannot be undone" confirmation only.
- [ ] Confirm the top-page delete still behaves as #1511 established -- the sub-pages survive as standalone published content and are not promoted into new collections.
- [ ] Accessibility: on the top-page delete confirmation, confirm the heading order still runs down to the `<h1>` "Are you sure you want to delete the content item..." with no level skipped, and that tabbing still reaches Delete then Cancel.

References yalesites-org/YaleSites-Internal#1569

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_book/tests/src/Kernel/BookCollectionDeleteTest.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_book/tests/src/Kernel/BookCollectionDeleteTest.php
@@ -407,6 +407,54 @@ class BookCollectionDeleteTest extends KernelTestBase {
   }
 
   /**
+   * The collection top page drops contrib book's relocation warning.
+   *
+   * Contrib book_form_node_confirm_form_alter() adds "the child pages will be
+   * relocated automatically" to every book page that has children. On a
+   * collection's top page that is no longer what happens: ys_book takes the
+   * collection apart and the sub-pages stay put as standalone content. Showing
+   * both sentences tells the editor two contradictory things about the same
+   * irreversible action.
+   */
+  public function testDeleteConfirmFormDropsContribWarningOnCollectionTop(): void {
+    [$root] = $this->createCollection();
+
+    $form = $this->container->get('entity.form_builder')
+      ->getForm($this->reload($root), 'delete');
+
+    $this->assertArrayNotHasKey(
+      'book_warning',
+      $form,
+      "Contrib's relocation warning is wrong for a collection top page."
+    );
+    $this->assertArrayHasKey(
+      'ys_book_collection_notice',
+      $form,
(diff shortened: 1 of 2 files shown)


miketullo95 (approved):
@dblanken-yale Approving. Tested this on the multidev and it behaves.

Confirmed both halves: a collection's top page now shows only the YaleSites copy, and a mid-collection page with its own child still gets contrib's "relocated automatically" warning. That second case is the one I was watching, since it would be easy to over-remove.

Pulling `_ys_book_is_collection_top()` out so the delete message and `ys_book_node_predelete()` share one definition is the part I like most here. The rule was stated in two places before and only one of them was still right.

PR #1484, opened 2026-08-19
1566: Bug: ys_node_access_update_10001 runs an unbatched node_access_rebuild during drush deploy

dblanken-yale:
### Description of work

The ticket asked to measure the risk first and batch only if the timing was uncomfortable. Measuring turned up something worse than a slow deploy, so this batches both hooks.

**What the measurement found.** Core's `node_access_rebuild()` calls `deleteGrants()` — a truncate of the entire `node_access` table — *before* it rewrites anything, then rebuilds every node in one unbounded pass. So the exposure is not only "it might time out": for the whole duration of the rebuild, successful or not, no node has grant rows. Reproduced locally by replaying exactly what `ys_node_access_update_10001()` does and stopping it part way, standing in for a PHP time or memory limit:

```
rows_before=77  rows_after_deleteGrants=0  rows_after_5_of_71=9
/empty-testing-page      HTTP 200 -> 403
/content-types/resource  HTTP 200 -> 403
/content-types/profile   HTTP 200 -> 403
```

Every page a visitor can reach 403s, and stays that way until someone re-runs `updatedb`. Update hooks run in the `updatedb` phase of `drush deploy`, so this lands mid-deploy on a live site.

On timing, against 2,071 nodes generated locally: about **2.5 ms and 20 KB of unreleased memory per node**, both linear. Time alone would clear core's own `Environment::setTimeLimit(240)` up to roughly 36k nodes, so **memory is the binding constraint** and it is the one that produces the half-rebuilt table. I could not measure the largest real site — nothing in this ticket touched a remote environment — so treat the absolute node counts as indicative; the shape is what matters.

**The change.** Both `ys_node_access_update_10001()` and `ys_node_access_update_10002()` now delegate to a shared `_ys_node_access_rebuild_grants(array &$sandbox)`:

- Chunked over passes via `$sandbox`, 50 nodes at a time, with per-pass progress in the deploy log.
- **No table-wide delete.** `NodeGrantDatabaseStorage::write()` already replaces the rows of the node handed to it, so a per-node rewrite is self-cleaning. A node therefore always holds either its old grants or its new ones — never none — which is what removes the blackout entirely.
- **The `nid = 0` wildcard grant is deleted explicitly on the first pass.** This one is worth a close look: `NodeGrantDatabaseStorage::access()` ORs `nid = 0` into the query for published nodes and `node_access_grants()` always contributes realm `all` / gid 0, so a surviving nid-0 row grants view on every published node to everyone, CAS-protected pages included. `node_access_rebuild()` removed it as a side effect of truncating (that is what the comment on `ys_node_access_install()` is about); a per-node rewrite never reaches it, so it has to be removed on purpose. Covered by `testRebuildRemovesTheNidZeroWildcardGrant`.
- Paged on a **keyset cursor** (`nid > current_nid`, sorted ASC) rather than an offset. `$sandbox` is serialised to `key_value` between passes, and holding a snapshot of every nid on a large site is exactly the memory this change exists to avoid. The cursor advances past every nid the page asked for, not just the ones that loaded, so a node deleted mid-deploy cannot stall it.
- The entity static cache is released after each pass. Without that, drush running every pass in one PHP process means the batching bounds time but not memory — which would have left the real failure mode in place.
- Terminates on an **empty** page rather than a short one. A short page is the cheaper test, but the query pages over rows of `node_field_data`, so on a site with translations a full page can collapse to fewer than 50 nodes and stopping there would leave the rest of the site on stale grants. Not reachable today (no `language` / `content_translation` / `locale` in `core.extension.yml`), and the extra pass costs one query that returns nothing.

**One trade worth a reviewer's eye:** an aborted `updatedb` now fails *open* rather than *closed*. Previously an abort left the table empty and everything 403'd; now the nodes the rebuild never reached keep their pre-deploy rows, which for these two hooks means unpublished drafts keep the over-permissive grant until a successful re-run. That state is identical to not having run the update at all, the hook's schema version is not advanced so the next deploy re-runs it from scratch, and `drush deploy` aborts loudly — whereas the fail-closed version took the whole site down. I think that is the right trade, but it is a deliberate one rather than an accident.

`ys_node_access_install()` is deliberately left unbatched: `hook_install()` has no `$sandbox` to return to, and it runs at profile install time before any content exists. There is now a comment saying so.

**Release note:** no site should see a longer-than-usual deploy window from this — the rebuild is faster and bounded, not slower. Both hooks run back to back on any site upgrading past this release, so a large site pays the (now safe) rebuild twice; deduplicating them was considered and left alone, since the correct mechanism is a process-scoped static and the saving is seconds.

### Functional testing steps:

- [ ] Pull a real site database locally (not a fresh `site:install` — the whole point is that this runs against existing content).
- [ ] Roll `ys_node_access` back so the hooks re-run: `lando drush php:eval '\Drupal::keyValue("system.schema")->set("ys_node_access", 10000);'`
- [ ] Run `lando drush updatedb -y`. It should report per-pass progress (`Rebuilt node access grants for 50 of about N nodes.`) and finish with `Rebuilt node access grants for N nodes.`
- [ ] While it is running, load a published page as an anonymous visitor (a private/incognito window). It should return 200 throughout — no 403s at any point during the rebuild. This is the regression being fixed.
- [ ] After it finishes, confirm `ys_node_access` is at schema 10002 and that the "content access permissions need to be rebuilt" warning is **not** showing on `/admin/reports/status`.
- [ ] Confirm access still behaves: a published page is visible to anonymous; a page with "CAS login required" set redirects anonymous to CAS and renders for a logged-in user; an unpublished draft is visible to its owner and to a site admin in `/admin/content`, and not to an unrelated authenticated user.
- [ ] Interrupt it: roll the schema back to 10000 again, start `updatedb`, and kill it part way. Confirm the site still serves published pages (200, not 403), then re-run `updatedb` and confirm it completes cleanly from the start.

References yalesites-org/YaleSites-Internal#1566

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/tests/src/Kernel/NodeAccessRebuildUpdateTest.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_node_access/tests/src/Kernel/NodeAccessRebuildUpdateTest.php
@@ -1,0 +1,371 @@
+<?php
+
+namespace Drupal\Tests\ys_node_access\Kernel;
+
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\node\Entity\Node;
+use Drupal\node\Entity\NodeType;
+use Drupal\user\Entity\User;
+use Drupal\ys_node_access\NodeAccessManager;
+
+/**
+ * Tests the batched node access grant rebuild in ys_node_access.install.
+ *
+ * See _ys_node_access_rebuild_grants() for why the rebuild the two update
+ * hooks share is batched rather than a call to core's node_access_rebuild().
+ * These tests pin the properties that make it safe to run during a deploy:
+ * it is chunked over several passes via $sandbox, it never empties
+ * node_access, and it still clears the nid 0 wildcard grant that the rebuild
+ * it replaces removed. Issue #1566.
+ *
+ * @group yalesites
+ * @group ys_node_access
+ */
+class NodeAccessRebuildUpdateTest extends KernelTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'system',
+    'node',
+    'field',
+    'text',
+    'user',
+    'ys_node_access',
+  ];
+
+  /**
+   * Number of nodes the multi-pass tests build.
+   *
+   * Deliberately larger than the rebuild's batch size (50) so a complete
+   * rebuild has to span more than one pass -- that is the behaviour under
+   * test. Raising the batch size past this would make those tests fail.
+   */
+  const NODE_COUNT = 60;
+
+  /**
+   * Guard against a rebuild that never reports itself finished.
+   */
+  const MAX_PASSES = 20;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    $this->installEntitySchema('node');
+    $this->installEntitySchema('user');
+    $this->installSchema('node', ['node_access']);
+    $this->installConfig(['node', 'user']);
+
+    NodeType::create(['type' => 'page', 'name' => 'Page'])->save();
+
+    // The update hooks live in ys_node_access.install, which is not loaded
+    // for a kernel test the way it is during updatedb.
+    \Drupal::moduleHandler()->loadInclude('ys_node_access', 'install');
+  }
+
+  /**
+   * Tests that a full rebuild is chunked over more than one pass.
+   *
+   * An unbatched rebuild does all of its work in a single call and never
+   * reports progress, so a large site's rebuild is one unbounded unit of work
+   * that a PHP limit can kill outright.
+   */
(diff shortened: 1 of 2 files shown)


miketullo95 (approved):
@dblanken-yale Approving.

The measurement writeup is what sold me on this. Truncating the whole table before rewriting anything is a much worse failure mode than a slow deploy, and the per-node rewrite in `_ys_node_access_rebuild_grants()` closes it cleanly. Explicitly deleting the nid 0 row on the first pass is the detail I would have missed, and it is good to see it pinned by `testRebuildRemovesTheNidZeroWildcardGrant`.

One note for whoever picks this up in testing: the PR multidev is built from dev, which is already at schema 10002, so both hooks get skipped there and the new path never actually runs. Verification has to be the local run you describe in the testing steps, not a pass through the multidev.

The fail-open trade reads right to me. Identical to not having run the update, schema does not advance, deploy aborts loudly. Better than taking the site down.

PR #1487, opened 2026-08-19
1575: Bug: Pre-built Form (Contact Us): restore access to submission results via the Layout Builder edit icon

dblanken-yale:
### Description of work

- Adds a **View form submissions** item to the Layout Builder contextual pencil of any block that references a webform, so editors have an in-context route to their Pre-Built Form submissions again. yalesites-org/YaleSites-Internal#929 correctly hid the Webform module's own edit icon from rendered pages, but that removed the only path editors had to their submission data.
- Implemented the way this module already adds an action to that pencil: a new contextual-links group (`ys_layouts_webform_results` in `ys_layouts.links.contextual.yml`) attached by a `#pre_render` on the `layout_builder` element (`WebformResultsLinkBuilder`), registered in `ys_layouts_element_info_alter()` alongside `CloneBlockLinkBuilder`. Groups attached to the same element merge into one pencil and sort by weight, so nothing about core's group is altered — which matters, because core's route parameters describe a layout component and cannot address a webform route.
- The webform is read off the block content entity **Layout Builder has already put in the component's render array** (`content['#block_content']`, the same path `_ys_layouts_add_layout_section()` uses) rather than reloaded from section storage. That is deliberate on two counts. It costs no entity load at all — resolving through the section storage instead meant a `loadRevision()` per placed block, which the entity cache does not serve, so a 20-block page paid ~19 avoidable queries per render. And it is more correct: the rendered entity is the *pending* one, so a form block placed (or repointed at a different form) in an unsaved editing session links to the form the editor is actually looking at, where `block_revision_id` would have been `NULL` or stale.
- **Permission gating is inherited, not re-implemented.** `ContextualLinkManager` access-checks every link's route, and `entity.webform.results_submissions` requires the `webform.submission_view_any` operation — i.e. the `view any webform submission` permission. So the link appears exactly for the users allowed to follow it, driven off the real permission rather than a role list, as the scoping comment on the issue asks.
- **Only the Results link is surfaced.** Webform's own group also carries Test / Build / Settings; those are redundant with the Webform admin tabs and would take the pencil past eight items, so they are deliberately left out (the issue's "does not become cluttered" criterion). The label is worded for editors — "View form submissions" rather than Webform's bare "Results" — and it sorts after the in-place layout actions and before the destructive Remove. It has no `use-ajax`/dialog options, unlike its two siblings, because it navigates to a full admin page; pending layout edits survive in the tempstore.
- **`ys_core` now strips the Webform group on every route**, dropping the old `layout_builder.*` exception and `WebformContextualLinksSuppressor::shouldSuppress()`. That exception surfaced nothing: the group is attached deep inside the submission form rather than to the block, so it never reached the block's pencil. Verified in the browser before changing anything — the server did emit `data-contextual-id="webform:webform=contact"` inside the `<form>` on `/node/3/layout`, and `/contextual/render` returned Test/Results/Build/Settings for it, but none of the contextual widgets actually rendered on the page were that group. Keeping it would have meant two mechanisms for one affordance, the dead one of which exposes the three links we just decided not to show.
- Unit tests: new `WebformResultsLinkBuilderTest` (7 tests) covering a form block, a block with no webform field, an empty webform reference, a component that renders no block content at all, arbitrary nesting depth, the no-built-layout no-op, and the trusted-callback registration. `BlockContextualLinksTest` now asserts the full pencil order including the new link.

**Verified locally on Lando** (`lando composer code-sniff` and `lando composer lint:php` both exit 0):

- The pencil on a form block reads Configure / Move / Clone / **View form submissions** / Remove, resolving to that form's results page; non-form blocks' pencils are unchanged.
- On a page with two different Pre-Built Form blocks, each pencil resolves to its own form (`.../manage/contact/...` and `.../manage/contact_two/...`), not a shared or first-found one.
- A form block added through the real add-block form and **not yet saved** still gets the link, pointing at the right form.
- As a role with layout-edit access but without `view any webform submission`, the pencil shows no submissions link, and the rendered page shows no contextual icon at all — so #929 is not regressed.
- Light accessibility pass on the open menu: the new item's accessible name is the plain text "View form submissions", the pencil opens with Enter and Tab reaches the link, and the link text is ~4.98:1 against its background.

**Two notes for the reviewer**

- `BlockContextualLinksTest` was **already failing on `develop`** before this branch — its expected menu omitted `ys_layouts_block_detach` ("Make non-reusable"), which was added to the same YAML after the test was written. Since this change adds another link to that file the expectation had to be updated anyway, so that drift is fixed here; it is not new breakage.
- `attachResultsLinks()` walks the built layout the same way `CloneBlockLinkBuilder::attachCloneLinks()` does, so that recursion shape now exists twice (three times counting `DetachContextualOperations`). Extracting a shared "walk the layout and attach a group" helper is the obvious next step, but it would mean refactoring existing code that has no unit coverage on its walk, on a Hotfix-priority bug fix — so it is left as a follow-up. Happy to fold it in here if you would rather it land together.
- The issue's documentation criterion (updating user-guide content about how editors find their submissions) is **not** covered by this PR — that content lives outside this repo.

### Functional testing steps:

- [ ] As an administrator or editor, visit a page containing a Pre-Built Form (Contact Us) block in normal view (**not** Edit Layout and Content). Confirm there is still no edit icon on the form and no visible link to submission data, and that the form still renders and can be submitted.
- [ ] Enter **Edit Layout and Content** on that page and open the pencil on the Pre-Built Form block. Confirm it reads Configure / Move / Clone / **View form submissions** / Remove, and that the new link lands on that form's own results page.
- [ ] On a page with **two different** Pre-Built Form blocks, confirm each pencil links to its own form's results, not a shared or first-found one.
- [ ] Add a new Pre-Built Form block and, **without saving the layout**, confirm its pencil already offers the link and points at the form you just chose.
- [ ] Confirm the pencils on non-form blocks on the same page are unchanged (no submissions link).
- [ ] As a role with layout-edit access but **without** `view any webform submission`, confirm the pencil on the form block shows no submissions link, and the rendered page still shows no webform edit icon.
- [ ] With the menu open, confirm it is still keyboard-operable: Tab to the pencil, open it with Enter, and Tab to reach "View form submissions".

References yalesites-org/YaleSites-Internal#1575

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/WebformContextualLinksSuppressor.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/WebformContextualLinksSuppressor.php
@@ -5,38 +5,27 @@ namespace Drupal\ys_core;
 use Drupal\Core\Security\TrustedCallbackInterface;
 
 /**
- * Decides when to strip the Webform module's contextual "edit" links.
+ * Strips the Webform module's contextual "edit" links.
  *
  * The Webform module unconditionally attaches a "webform" contextual-links
  * group (Test / Results / Build / Settings) to every rendered submission form.
  * On YaleSites that surfaces an edit icon — and a direct link to submission
  * data — to anyone with "access contextual links" who merely views a page
- * containing a Pre-Built Form block, outside of Layout Builder's "Edit Layout
- * and Content" mode. The rule for when to remove that group is centralised here
- * so it is unit-testable in isolation from the render pipeline.
+ * containing a Pre-Built Form block (issue #929).
+ *
+ * The group is removed everywhere. It used to be kept on Layout Builder routes,
+ * but that surfaced nothing: the group is attached deep inside the submission
+ * form rather than to the block, so it never reaches the block's contextual
+ * pencil. Editors reach their submissions through the pencil's
+ * "View form submissions" link instead, which offers only that one link and is
+ * access-checked on the webform's own results route
+ * (yalesites-org/YaleSites-Internal#1575).
  *
  * @see \_webform_form_webform_submission_form_after_build()
  * @see ys_core_form_alter()
+ * @see \Drupal\ys_layouts\WebformResultsLinkBuilder
  */
 class WebformContextualLinksSuppressor implements TrustedCallbackInterface {
-
-  /**
-   * Determines whether the webform contextual links should be removed.
-   *
-   * They are kept only while editing in Layout Builder (its routes are prefixed
-   * with "layout_builder."); on every view route — and defensively when no
-   * route matched — they are removed so the edit icon cannot leak submission
-   * data on the live page.
-   *
-   * @param string|null $route_name
-   *   The current route name, or NULL when no route matched.
-   *
-   * @return bool
-   *   TRUE if the webform contextual links should be removed.
-   */
-  public static function shouldSuppress(?string $route_name): bool {
-    return $route_name === NULL || !str_starts_with($route_name, 'layout_builder.');
-  }
 
   /**
    * Pre-render callback that strips the webform contextual-links group.
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/tests/src/Unit/WebformContextualLinksSuppressorTest.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/tests/src/Unit/WebformContextualLinksSuppressorTest.php
@@ -11,9 +11,12 @@ use Drupal\ys_core\WebformContextualLinksSuppressor;
  * Regression coverage for the bug where the Webform module's contextual links
  * group (Test / Results / Build / Settings) — including access to submission
  * data — appeared as an edit icon on Pre-Built Form blocks while a page was
- * merely being viewed, outside of Layout Builder's "Edit Layout and Content".
- * The links must stay only on Layout Builder editing routes and be removed on
- * every view route.
+ * merely being viewed (issue #929). The group is now removed on every route,
+ * including Layout Builder's: keeping it there surfaced nothing, because it is
+ * attached inside the submission form rather than to the block, so it never
+ * reached the block's contextual pencil. The submissions link editors need is
+ * added to that pencil instead, and is covered by
+ * \Drupal\Tests\ys_layouts\Unit\WebformResultsLinkBuilderTest.
  *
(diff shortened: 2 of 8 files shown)


miketullo95 (approved):
@dblanken-yale Approving. I confirmed the "View form submissions" item on the multidev, and the mechanism holds up everywhere I checked it.

A few things I verified rather than took on trust:

**The pencil order is checkable from config alone.** Core's Configure and Move are unweighted, `ys_layouts_block_detach` and `layout_builder_block_clone` sit at 5, the new `ys_layouts_webform_results` at 6, and `ys_layouts_contextual_links_plugins_alter()` pins Remove at 10. The new link lands between Clone and Remove exactly as described.

**The permission gating is genuinely inherited.** The link declares `entity.webform.results_submissions` and `ContextualLinkManager` access-checks each link's route, so `view any webform submission` governs it without a role list anywhere in this diff. That is what the AC asked for and it is the right shape.

**#929 is not at risk, and the rendered page is actually stricter now.** The new link rides a `#pre_render` on the `layout_builder` element, which only exists inside Edit Layout and Content, and dropping the `layout_builder.*` exemption means the Webform group is now stripped on every route rather than most of them.

**Reading the webform off `content['#block_content']` covers three ACs structurally, not just empirically** - two different forms on one page, an unsaved newly placed block, and a form in a nested section - because each component carries its own entity and the walk is depth-agnostic. Nice.

## On the documentation criterion

You flagged this as not covered because the content lives outside the repo, which is right, but the AC overstates the problem. I checked the live page: https://yalesites.yale.edu/pre-built-form-block never documented the on-page edit icon. On submissions it says only that they are sent to the address in Settings > Site Settings > Site Email. So there is nothing stale to correct, and no training material referencing an icon that the guide never mentioned.

What is true is that the guide has never told editors how to reach *stored* submissions in the UI, and this PR gives us a defensible answer for the first time. I am opening that as a small follow-up doc ticket on the yalesites.yale.edu side rather than as anything owed by this PR.

## Accessibility

Treating your self pass as sufficient here: one link added to an existing core menu pattern, keyboard reachable, accessible name is the plain link text, and 4.98:1 clears AA's 4.5:1. Not routing it for a separate engineer review.

Not taking you up on folding the shared layout-walk refactor into this one. Three copies of that recursion is worth fixing, but not on a Hotfix bug fix against code with no coverage on its walk. Better as its own change.

Nothing blocking.

PR #1488, opened 2026-08-20
1572: Feature Request: Give the orphaned inline block sweep a per-site route beyond a manual drush command

dblanken-yale:
### Description of work

The orphaned inline block sweep from yalesites-org/YaleSites-Internal#1503 only existed as `drush ys-layouts:orphaned-blocks`, so it only ever ran on the sites someone with terminal access remembered to run it on. This gives it a route onto an individual site without trading away any of the safety posture that made it worth shipping.

**The decision AC#1 asks for: an on-demand UI action for platform admins, report-first.** Not a scheduled job — silent platform-wide deletion is exactly what #1503's guardrails exist to prevent, and a scheduled report nobody reads is the same toil in a different place.

- Adds a report screen at `/admin/yalesites/orphaned-inline-blocks` listing every unreferenced inline block by ID, type and label, plus a separate table for blocks held only by an older page revision (which the sweep never deletes, because a rollback would make them live again).
- Deletion is a separate confirm step at `/admin/yalesites/orphaned-inline-blocks/delete`, linked from the report **only when there is something to delete**. The report itself never mutates anything.
- Gated on a new `administer orphaned inline blocks` permission (`restrict access: true`), granted to `platform_admin` alone. Deliberately its own permission rather than reusing `administer platform admin settings`, so a destructive capability can be re-scoped on its own later.
- Lives under `/admin/yalesites` rather than beside the existing `ys_layouts` settings form, because that form requires `administer site configuration` — which **no role in `config/sync` actually holds**, making it effectively user-1-only. Putting the new screen there would have shipped a page platform admins get a 403 on.
- The confirm form deletes *all* orphans rather than a selection: `deleteOrphans()` accepts no IDs and re-derives the list at the moment you confirm, which is what makes deleting a still-referenced block structurally impossible rather than merely guarded against. A block put back on a page between the report and the confirmation is left alone.
- Every UI-triggered deletion is written to the `ys_layouts` log channel with the IDs removed, identically to a drush-triggered one.
- Developer docs added to `ys_layouts/README.md` (AC#6 — developer-facing only).

**On AC#5 (behaviour mid-deploy or with a stale layout tempstore)** — both checked before exposing the button, and documented in the README:

- *Unsaved layout work is invisible to the sweep.* Adding, cloning or detaching a block in Layout Builder carries the new block as `block_serialized` with `block_revision_id` set to `NULL`, so no `block_content` row exists until the layout is saved. Verified against a tempstore holding an unsaved cloned block: 0 rows created, 0 new orphans reported. An editor's in-progress layout cannot be swept out from under them.
- *A deploy does not change the sweep's answer.* Its only inputs are saved entity revisions and the Layout Builder defaults in `entity_view_display` config. No default layout in `config/sync` carries a `block_revision_id`, so a `config:import` cannot transiently strip a block's last reference, and `ys_layouts` ships no update hook that rewrites layout data.

**Still open for discussion (AC#4):** platform-wide orphan magnitude is not established here, because measuring it means running the sweep against every Pantheon site and this work was done entirely against local Lando. The one real data point: this checkout's DB, pulled from an ordinary production site, had **177 orphans**. This screen is what makes per-site measurement possible without terminal access.

**Accessibility (light gate):** both data tables use real `<caption>` elements and now carry explicit `scope="col"` on every `<th>` (Drupal's `template_preprocess_table()` hands a plain-string header cell an empty attribute bag, so a bare string renders a scopeless `<th>`). The danger delete link measures 4.71:1 and the confirm submit 4.99:1, both passing; no missing `alt`, no unlabelled controls, no empty accessible names. One finding is **not actionable**: the Gin admin toolbar's active nav item measures 1.53:1, which is contrib chrome present on every admin page in the platform and would need a contrib patch to fix.

**Heads-up unrelated to this branch:** `ys_layouts`' Unit suite has one pre-existing failure on `develop` — `BlockContextualLinksTest::testRemoveStaysLastAndLabelsDropTheWordBlock` expects the contextual link set *without* `ys_layouts_block_detach`, while the code now adds it. Neither that test nor its subject is touched here (suite goes 106→110 tests, same single failure), but it will show up if you run the suite.

### Functional testing steps:

- [ ] As a **platform administrator**, go to `/admin/yalesites/orphaned-inline-blocks` (also reachable from the YaleSites admin menu as "Orphaned inline blocks"). Confirm the page loads and lists any unreferenced inline blocks with their ID, type and label.
- [ ] Confirm that when blocks are held only by an older page revision, they appear in the second table ("Referenced only by an older revision — never deleted") and **not** in the deletable list.
- [ ] On a site with no orphans, confirm the page reads "No orphaned inline blocks found." and that **no delete button is shown**.
- [ ] As an **editor** or **site administrator**, confirm both `/admin/yalesites/orphaned-inline-blocks` and `/admin/yalesites/orphaned-inline-blocks/delete` return **403**. Confirm anonymous also gets 403.
- [ ] As a platform administrator, click the "Delete N orphaned inline blocks" button, confirm you land on a confirmation screen that explains what will and will not be deleted, and that **Cancel** returns you to the report without deleting anything.
- [ ] Confirm the deletion, then confirm the report shows the empty state, the delete button is gone, and the blocks held by older revisions are **still listed and untouched**.
- [ ] Confirm `lando drush ys-layouts:orphaned-blocks` agrees with what the screen showed, and that `drush watchdog:show --type=ys_layouts` records the deletion with the IDs removed.
- [ ] Take a page that has a non-reusable block, remove the block from the layout, save, and confirm it then appears on the report.
- [ ] Open a page in Layout Builder, add or clone a block but **do not save the layout**, then load the report and confirm the unsaved block does **not** appear as an orphan.

**Pagination (added after functional review):** both tables page at 50 rows, each on its own pager element.

- [ ] On a site with more than 50 blocks held only by an older revision, confirm the "Referenced only by an older revision" table shows 50 rows with a pager beneath it, and that the caption states the **full** total rather than the number of rows shown.
- [ ] Page that table to page 2 and confirm the remaining rows appear, and that the **unreferenced** table above it did **not** move — the two tables page independently.
- [ ] Confirm the "Delete N orphaned inline blocks" button reads the **full** orphan count on every page, not the number of rows currently displayed.
- [ ] On a table with 50 or fewer rows, confirm **no** pager is rendered at all.
- [ ] Append a nonsense page to the URL (`?page=999`) and confirm you land on the last page of results, not an empty table.

**Note:** the heads-up above about a pre-existing `BlockContextualLinksTest` failure is now stale — develop fixed it, and the `ys_layouts` Unit suite is fully green (123 tests, 123 pass).

References yalesites-org/YaleSites-Internal#1572

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/user.role.platform_admin.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/user.role.platform_admin.yml
@@ -114,6 +114,7 @@ dependencies:
     - ys_campus_groups
     - ys_core
     - ys_integrations
+    - ys_layouts
     - ys_localist
 id: platform_admin
 label: 'Platform administrator'
@@ -146,6 +147,7 @@ permissions:
   - 'administer coffee'
   - 'administer google_tag_container'
   - 'administer main menu items'
+  - 'administer orphaned inline blocks'
   - 'administer platform admin settings'
   - 'administer redirects'
   - 'administer users'
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/README.md
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/README.md
@@ -73,6 +73,54 @@ Safety guarantees, in order of importance:
 Reusable blocks from the Custom Block Library are never candidates; only `reusable = 0` blocks
 are considered.
 
+### From the admin UI
+
+The command needs terminal access, which means the sweep only ever ran on the sites someone
+remembered to run it on. Platform admins can reach the same sweep per site at:
+
+```
+/admin/yalesites/orphaned-inline-blocks
+```
+
+The screen reports and nothing more; deleting is a separate confirm step at
+`/admin/yalesites/orphaned-inline-blocks/delete`, linked from the report only when there is
+something to delete. Both routes are gated on `administer orphaned inline blocks`, a
+`restrict access: true` permission granted to `platform_admin` alone — deliberately not
+`yalesites manage settings` (site admins hold that) and not `administer site configuration`
+(no role on the platform holds it, so `ys_layouts.settings_form` is effectively user-1-only).
+
+The confirm form deletes *all* orphans rather than a selection, because `deleteOrphans()`
+derives its own list — see guarantee 5 above. The count shown on the report is therefore
+advisory: the list is worked out again when you confirm, so a block that was put back on a
+page in between is left alone. Every deletion is written to the `ys_layouts` log channel with
+the IDs removed, whichever route triggered it.
+
+The sweep runs synchronously in the request. It walks every revision of every layout-bearing
+entity, so cost scales with revision count rather than orphan count; on a site with ~180
+orphans and a normal revision history it completes in about two seconds, comfortably inside a
+web request. The drush command remains the escape hatch if a site ever grows large enough to
+risk a PHP timeout, and is still the right tool for scripting across many sites.
+
+#### Timing: deploys and unsaved layout work
+
+A button can be pressed at any moment, whereas a drush command is run deliberately by someone
+who knows what else is happening on the site. Both timing questions were checked before the
+screen was exposed:
+
+- **Unsaved layout work is invisible to the sweep, so it cannot be collected.** Adding, cloning
+  or detaching a block in Layout Builder carries the new block on the component as
+  `block_serialized` with `block_revision_id` set to `NULL`; no `block_content` row exists until
+  the layout is saved. Confirmed against a tempstore holding an unsaved cloned block: no rows
+  created and no new orphans reported, so an editor's in-progress layout cannot be swept out
+  from under them.
+- **A deploy does not change the sweep's answer.** Its only inputs are saved entity revisions
+  and the Layout Builder defaults in `entity_view_display` config. No default layout in
+  `config/sync` carries a `block_revision_id`, so a `config:import` cannot transiently strip the
+  last reference to a block, and this module ships no update hook that rewrites layout data.
+  Deploys do not enable maintenance mode, so the screen stays reachable while one runs — which
+  is safe today. An update hook that rewrote layouts would reintroduce this risk and should
+  finish before anything sweeps.
+
 ## Running tests
 
 This module has PHPUnit tests under `tests/src/` (`Unit/` and `Kernel/`). Run them from the project root on the local Lando environment, passing the module's `tests` path so PHPUnit only discovers this module's tests (not Drupal core/contrib):
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/README.md
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/README.md
@@ -101,6 +101,14 @@ orphans and a normal revision history it completes in about two seconds, comfort
 web request. The drush command remains the escape hatch if a site ever grows large enough to
 risk a PHP timeout, and is still the right tool for scripting across many sites.
 
+Both tables page at 50 rows. Paging bounds the work that scales with how many blocks were
+found — each row loads a `block_content` entity, so an unpaged report on a site holding
+thousands of orphans would load and render thousands of entities — but it cannot make the
+sweep itself cheaper, because `analyze()` has to walk every revision before it knows what is
+orphaned. Each table pages on its own pager element, so paging one leaves the other where it
+was, and each caption states the full total rather than the number of rows on screen. The
+delete button always counts every orphan, never just the page in front of you.
+
 #### Timing: deploys and unsaved layout work
 
 A button can be pressed at any moment, whereas a drush command is run deliberately by someone
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Controller/OrphanedInlineBlockReportController.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_layouts/src/Controller/OrphanedInlineBlockReportController.php
@@ -4,6 +4,7 @@ namespace Drupal\ys_layouts\Controller;
 
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Pager\PagerManagerInterface;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
 use Drupal\Core\Url;
 use Drupal\ys_layouts\Service\OrphanedInlineBlockCleanerInterface;
@@ -27,16 +28,37 @@ class OrphanedInlineBlockReportController implements ContainerInjectionInterface
   use StringTranslationTrait;
 
   /**
+   * How many blocks each table lists per page.
+   */
+  const ITEMS_PER_PAGE = 50;
+
+  /**
+   * Pager element index for the deletable orphans table.
+   */
+  const ORPHANS_PAGER = 0;
+
+  /**
+   * Pager element index for the revision-only table.
+   *
+   * Distinct from ORPHANS_PAGER because both tables render on one page: sharing
+   * an element would make paging either table move both.
+   */
+  const REVISION_ONLY_PAGER = 1;
+
+  /**
    * Constructs an OrphanedInlineBlockReportController.
    *
    * @param \Drupal\ys_layouts\Service\OrphanedInlineBlockCleanerInterface $cleaner
    *   The orphaned inline block cleaner.
    * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entityTypeManager
    *   The entity type manager.
+   * @param \Drupal\Core\Pager\PagerManagerInterface $pagerManager
+   *   The pager manager.
    */
   public function __construct(
     protected OrphanedInlineBlockCleanerInterface $cleaner,
     protected EntityTypeManagerInterface $entityTypeManager,
+    protected PagerManagerInterface $pagerManager,
   ) {}
 
   /**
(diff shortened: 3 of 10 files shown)


miketullo95 (approved):
@dblanken-yale Approving. The safety posture survives the move from CLI to UI intact, which was the thing worth getting right here.

A few claims I checked rather than took on trust:

**UI deletions really do log identically to drush ones**, and structurally rather than by duplication: the `notice` lives in `OrphanedInlineBlockCleaner::deleteOrphans()`, not in the command, so both callers get it for free and neither can drift from the other.

**No circular dependency** from adding `ys_core` to `ys_layouts.info.yml`. `ys_core` depends on multivalue_form_element, media_library_form_element, role_delegation and cas, so nothing points back. The `ys_core.admin_yalesites` parent link exists as referenced.

**The permission is genuinely narrow.** `restrict access: true`, granted to `platform_admin` alone in `config/sync`, and site admins do not pick it up through `yalesites manage settings`. Right call: orphaned inline blocks are invisible database cruft that a site owner can neither perceive nor usefully act on, so read-only visibility for site admins would be noise rather than transparency.

**The performance question is already answered, which I went looking for specifically.** Moving this onto a web route matters because `analyze()` walks every revision of every layout-bearing entity and Pantheon caps web requests at 120 seconds. The README meets that head-on with a measurement (~2 seconds at ~180 orphans on a normal revision history) and names the drush command as the escape hatch. A 60x margin is comfortable, and calling out that cost scales with revision count rather than orphan count is the detail that makes it useful later.

The AC#5 work is the part I would have most expected to be hand-waved and it is the most thoroughly evidenced. Confirming against a real tempstore that an unsaved cloned block creates no `block_content` row is the difference between believing an editor cannot be swept out from under and knowing it.

**On AC#4**, treating it as satisfied. The ticket asked for platform-wide magnitude to size the priority, and you inverted it by shipping the thing that makes measurement possible per site without terminal access. That is the better order given the alternative was running drush against every Pantheon site to justify building the tool that removes the need to do that. Not opening a follow-up to chase the number.

One note for whoever closes #1572: the ticket's own framing of AC#4 no longer applies, so it should be closed on this reasoning rather than left open waiting on a number nobody is going to collect separately.

Nothing blocking.

dblanken-yale:
This is working as intended.  There is one question I have that's up for debate, and that is this feels like a Platform Admin only form; should we nest this inside of Platform Admin Settings in the menu?  This could clean up the length of menu items we get over time.

dblanken-yale:
@dblanken-yale Two things: an answer to the menu question (recommending **no**, with the measurements behind it), and the pagination, which is **done and pushed** in 743a21826.

## 1. Nesting under Platform Admin Settings — recommending against it

Measured it both ways on Lando as a real `platform_admin` (not user 1), with `user.role.platform_admin` imported first — before that import the link is invisible for the ordinary reason that the permission is not granted yet, which would have made any earlier reading of the menu wrong.

| | current (`parent: ys_core.admin_yalesites`) | nested (`parent: ys_core.platform_admin_settings`) |
|---|---|---|
| Listed on `/admin/yalesites` | **yes** (11 items) | **no** (10 items — the report is the one that disappears) |
| Toolbar item | present, `<li>` depth 2, inside a collapsed `.toolbar-menu` | present, `<li>` depth 3, inside a collapsed `.toolbar-menu` |
| Links to the report in the parent page's own body | n/a | **0** |

**The toolbar is not the differentiator**, which is the part I expected to go the other way. Both variants sit inside a collapsed `.toolbar-menu` measuring zero by zero, so both need the tray expanded either way. Nesting does not bury it any deeper in the toolbar than it already is.

**The `/admin/yalesites` listing is the real loss.** That route is `SystemController::systemAdminMenuBlockPage`, which lists **direct children only** — it structurally cannot reach a grandchild. The section index drops from 11 entries to 10, and the report is the entry that goes.

On best practice specifically: three-level nesting is fine and we already do it — `Content > Manage Menus > Manage Main Menu` and `Content > Manage Taxonomy > Event Category` both render at depth 3. What makes those work is that **the parent page's own content leads to the child**: `/admin/structure/menu` carries 2 links to `/admin/structure/menu/manage/main`, so the collapsed toolbar item is a shortcut rather than the only way in.

`ys_core.platform_admin_settings` is a plain form route (`PlatformAdminSettingsForm`) — a leaf. It has **0** links to the report in its body, and a form has no mechanism for listing child menu links. So nesting there produces a menu item whose only path in is expanding a collapsed toolbar subtree, which is strictly worse than today.

Second, smaller reason: it couples menu *visibility* to `administer platform admin settings`. We deliberately made `administer orphaned inline blocks` its own re-scopable permission rather than reusing that one; nesting means that if the new permission is ever granted to a role without the platform-admin-settings one, Drupal prunes the parent and the whole subtree with it, and the item silently vanishes from the menu while the route still works by URL.

I also looked at contributing a `PlatformAdminSetting` plugin from `ys_layouts` to render a link into that page, and rejected it: the contract is `buildSettings`/`validateSettings`/`submitSettings`, so a section with nothing to save sitting above a "Save configuration" button abuses the extension point.

**If menu length does become the pressing problem**, the structurally right fix is a platform-admin **section landing page** — a `systemAdminMenuBlockPage` route, the same pattern `/admin/yalesites` itself uses — with both "Platform Admin Settings" and "Orphaned inline blocks" hanging under it. A menu-block page lists its children, so the tidiness costs nothing in discoverability. That is an `ys_core` IA change affecting every platform-admin screen, so it wants its own ticket rather than riding along here. Happy to open one if you want it.

Worth noting the current weight already does some of this work: the report is weight 40, directly after Platform Admin Settings at 35, so the platform-admin-only entries group together at the bottom of the section.

**One-line change if you'd rather have it nested anyway** — `parent: ys_core.platform_admin_settings` in `ys_layouts.links.menu.yml`. Your call; I just didn't want to trade the section index away silently.

## 2. Pagination — good catch, and worse than it looked

You were right, and the reason is sharper than page weight: `describe()` loads a `block_content` **entity per row**, so an unpaged report on a site carrying thousands of orphans loaded and rendered thousands of entities to fill a table nobody scrolls. And you picked the right table — the revision-only one is the one most likely to be large, because nothing ever deletes from it.

Both tables now page at 50 rows, each on **its own pager element**, so paging one leaves the other where it was.

Two things had to stay true now that the rows on screen are only a window:

- **The delete button still counts every orphan, not the page in front of you.** The confirm step re-derives the full set and removes all of it, so a page-sized label would understate what confirming does. Pinned by a test.
- **Each caption now states the full total.** The revision-only table has no delete button to carry a count, so without it an operator on page two cannot tell whether the site holds sixty of these or six thousand.

An out-of-range `?page=` lands on the last page rather than an empty table, which falls out of `Pager` clamping its own current page into range.

**What pagination does not fix, to be straight about it:** `analyze()` still walks every layout revision before it can know what is orphaned, so the sweep's own cost is unchanged — it is bounded by revision count, not by orphan count, exactly as the README already said. Paging bounds the per-ID work downstream of that walk. The drush command remains the escape hatch for a site where the walk itself is the problem. Documented in the README rather than left implicit.

### Verification

Local Lando, real DB (2 orphans, 66 revision-only):

- Revision-only table renders 50 of 66 with the total in the caption; `?page=0,1` renders the remaining 16 — so the two pagers genuinely move independently.
- Delete button reads "Delete 2 orphaned inline blocks" throughout, on every page.
- The orphans table has one page of results, so its pager correctly renders nothing at all.
- `ys_layouts` Unit suite: **123 tests, 123 pass** (6 new). `lando composer code-sniff` and `composer lint:php` both exit 0.
- Accessibility: the pager is core markup and checks out — `<nav aria-labelledby="pagination-heading">`, `aria-current="page"` on the active item, active pill at 4.80:1. All six `<th scope="col">` from the earlier a11y pass are intact.

**Heads-up on a stale note in the PR description:** it warns about a pre-existing `BlockContextualLinksTest::testRemoveStaysLastAndLabelsDropTheWordBlock` failure. That is gone — develop fixed it, and the suite is fully green now.

Moving this back to **needs review** rather than leaving it at ready to merge, since the pagination is new code that has not been through functional review. The nesting question is still yours to call.

miketullo95 (approved):
@dblanken-yale Tested and approved.

Checked the diff against [#1572](https://github.com/yalesites-org/YaleSites-Internal/issues/1572):

- **AC#1** (decide how the sweep runs): answered as an on-demand UI action, report-first. Good call over a scheduled job.
- **AC#2** (report-only default, deliberate opt-in to delete): the report route never mutates, and deletion is its own confirm step at `/admin/yalesites/orphaned-inline-blocks/delete`, linked only when there is something to delete.
- **AC#3** (gate on platform admin, report before deletion): `administer orphaned inline blocks` is `restrict access: true` and granted to `platform_admin` alone in `config/sync/user.role.platform_admin.yml`. Agreed on making it its own permission rather than folding it into `administer platform admin settings`, since it authorizes deleting editor-created content.
- **AC#5** (behavior mid-deploy or with a stale tempstore): checked and written up.
- **AC#6** (developer docs): `ys_layouts/README.md` covers the screen, the paging behavior, and the timing questions.

Two things I want on the record, neither blocking:

1. **AC#4 is still open.** Platform-wide orphan magnitude is not established here, and that is fine given this screen is what makes per-site measurement possible in the first place. But our PR body format is not a closing keyword, so [#1572](https://github.com/yalesites-org/YaleSites-Internal/issues/1572) will not auto-close on merge. Please leave it open. I want us to run this screen across sites after release and record real numbers before we close it out.
2. **The 177 orphans on one ordinary production site** is worth repeating to the team once we have more data points. That number is the argument for why this needed a route beyond drush.

Nice work on keeping the delete button above the revision-only table rather than below it. That is the kind of detail that prevents someone clicking the wrong thing at the wrong moment.

PR #1489, opened 2026-08-20
1573: Review whether platform/site admin and editor still need "administer book outlines"

dblanken-yale:
### Description of work

**What the permission actually grants (the review half of the ticket).** The live book module is
contrib `drupal/book` 2.0.4, not core's deprecated copy, and `administer book outlines` gates
exactly four things here:

| Surface | Gate before | Affected |
| --- | --- | --- |
| `/admin/structure/book` — the "Content Collections" overview | `administer book outlines` | lost without it |
| `/admin/structure/book/{node}` — reorder pages and titles | `_ys_book_outline_access` (added by #1145) | already safe |
| `/node/{node}/child-ordering` — the "Child order" node tab | `administer book outlines` **AND** contrib's own check, which requires `reorder book pages` | lost without it (needed *both*) |
| `/admin/structure/book/{node}/delete` — Delete collection | `administer book outlines` | lost without it |

Everything else book-related runs off other permissions the roles keep: the Content Collection
widget on the node form, the Outline tab and Remove-from-outline (`add content to books`), creating
a collection (`create new books`), printer-friendly export (`access printer-friendly version`).
`/admin/structure/book/settings` needs `administer site configuration`, which no role has.

One cosmetic behaviour does change: contrib's `BookOutlineConstraintValidator` uses the broad
permission only to decide whether its "you can only change the book outline for the published
version" violation message includes a link to the collections page, so these roles now see the
link-less variant. Reachable (a draft of a published collection page hits it), cosmetic, and in
contrib — noted rather than patched.

**So the original worry does not hold: nothing was exposed.** `/admin/structure/book` is not a
leaked contrib admin page — `ys_book` overrides that route with its own controller and re-parents
its menu link under Content as "Manage Content Collections". It is a first-class editor surface.
What *was* wrong is the shape of the grant: our own screens demanded contrib's
administer-everything permission, so every editing role had to hold it.

**What changed.**
- `book.admin` and `ys_book.collection_delete` now use the existing `_ys_book_outline_access` check
  (narrow `reorder book pages` OR the broad permission) instead of requiring the broad permission.
- Reinstated a `ys_book` override of contrib's `book.node_child_ordering` with the same swap, so the
  "Child order" tab keeps working. That override was deliberately removed in `0f8334bd4` as
  unneeded — it was, while every role held the broad permission, and it is needed again precisely
  because that is what this ticket removes. Contrib's own `_custom_access` still applies, so the
  narrow permission is still required and the tab still only appears when a collection has more
  than one sibling to order.
- `BookOutlineAccessCheck` now returns `AccessResult::allowedIfHasPermissions(..., 'OR')`. The old
  `AccessResult::allowedIf($a || $b)` carried no `user.permissions` cache context — a latent
  caching bug on one route that would have become a real one on four.
- Removed `administer book outlines` from `editor`, `site_admin` and `platform_admin` in
  `config/sync`. `reorder book pages` and the other book permissions are untouched.
- Documented the gate and the reasoning in `ys_book/README.md`.

**No update hook, deliberately.** Roles ship only from `config/sync` (nothing in any module
`config/install`/`config/optional`, and `config_ignore` has no `user.role*` pattern), so
`drush deploy`'s `config:import` phase applies the revocation on its own. A revoking hook would be
redundant with config import and could fight it. A site that granted the broad permission to a
custom role of its own is untouched by config import and keeps working, because every re-gated
check still accepts that permission — there is a test for exactly that.

**Two decisions worth a reviewer's attention.**
1. **Collection delete stays at the same level as reorder.** `ys_book.collection_delete` was gated
   on the broad permission, so removing it from a role would also have removed that role's ability
   to delete a collection. Rather than let that fall out by accident, delete is re-gated on the same
   check as reorder, which preserves exactly the access editors have today. Making delete a higher
   bar than reordering is defensible, but it is a behaviour change for editors and a separate product
   decision — say the word and it is a small follow-up. The sharpest way to state the consequence:
   `reorder book pages` now also permits dismantling a collection, so a site-specific role given
   that permission alone would get delete along with reorder. No such role exists today.
2. **`platform_admin` loses it too, not just editor and site_admin.** After the re-gate nothing on
   the platform requires the broad permission, so leaving it on platform_admin would grant only
   contrib breadth nobody uses — including any future contrib book route that lands gated on it.

**Tests.** New `BookCollectionPermissionTest` (4 tests, 72 assertions) asserts all four screens are
reachable with `reorder book pages` alone, that the broad permission still reaches the three screens
we gate, and that neither permission means denied. It also records the asymmetry the work turned up:
contrib gates child ordering on the narrow permission and nothing else, so the broad permission alone
has never been enough there. `BookCollectionDeleteTest` stays green at 14/14 (unchanged from
baseline); one comment in it had gone stale and was corrected.

**Known follow-up, not done here:** five `ys_book` kernel tests now hand-copy the same `setUp()`
fixture recipe and two have already drifted from it. Extracting a shared trait is the right call but
means editing four test files outside this ticket, so it is left as a follow-up rather than folded
into a permissions change.

### Functional testing steps:

- [x] As an **editor**, go to Content → Manage Content Collections. The overview loads and lists the site's collections.
- [x] Open a collection's re-order screen, change the order of its pages, and save. The new order sticks after a reload.
- [x] On a collection page with two or more sub-pages, open the **Child order** tab, reorder, and save. The new order sticks.
- [x] From the Collections overview, use **Delete collection** on a test collection. The confirmation appears, and after confirming the collection is gone while every page survives as standalone content.
- [x] Repeat all four steps as a **site admin** and as a **platform admin**.
- [x] Check `/admin/people/permissions` — "Administer book outlines" is unchecked for Editor, Site Admin and Platform Admin, while "Reorder child page of a book" stays checked.
- [x] As a **contributor** (who has neither book permission), confirm Manage Content Collections is still not reachable — this should be unchanged.
- [x] On a node edit form, confirm the Content Collection section still appears and a page can still be added to, or removed from, a collection.

References yalesites-org/YaleSites-Internal#1573

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/user.role.editor.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/user.role.editor.yml
@@ -114,7 +114,6 @@ permissions:
   - 'access webform overview'
   - 'add any content to books'
   - 'add content to books'
-  - 'administer book outlines'
   - 'administer main menu items'
   - 'administer redirects'
   - 'administer users'
--- a/web/profiles/custom/yalesites_profile/config/sync/user.role.platform_admin.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/user.role.platform_admin.yml
@@ -142,7 +142,6 @@ permissions:
   - 'administer block content'
   - 'administer block types'
   - 'administer blocks'
-  - 'administer book outlines'
   - 'administer coffee'
   - 'administer google_tag_container'
   - 'administer main menu items'
--- a/web/profiles/custom/yalesites_profile/config/sync/user.role.site_admin.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/user.role.site_admin.yml
@@ -122,7 +122,6 @@ permissions:
   - 'access webform overview'
   - 'add any content to books'
   - 'add content to books'
-  - 'administer book outlines'
   - 'administer google_tag_container'
   - 'administer main menu items'
   - 'administer redirects'
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_book/README.md
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_book/README.md
@@ -6,6 +6,46 @@ CAS-protected (and other access-restricted) published pages remain visible in th
 book navigation — flagged for a lock icon — rather than being filtered out, and
 invalidates collection-navigation caches on book node changes.
 
+## Who can manage content collections
+
+Collection management is gated on contrib book's narrow **`reorder book pages`**
+permission, via the `_ys_book_outline_access` access check
+(`src/Access/BookOutlineAccessCheck.php`). The broad **`administer book
+outlines`** permission is also accepted, so a site that granted it to a role of
+its own keeps working, but no platform role holds it any more.
+
+That check gates the three screens this module owns:
+
+| Screen | Route |
+| --- | --- |
+| Manage Content Collections overview | `book.admin` |
+| Re-order collection pages and titles | `book.admin_edit` |
+| Delete collection | `ys_book.collection_delete` |
+
+It also gates contrib's "Child order" node tab (`book.node_child_ordering`),
+which `ys_book.routing.yml` overrides for exactly that reason. Contrib's own
+access check on that route additionally requires `reorder book pages`, so the
+broad permission alone has never been enough there.
+
+Everything else book-related runs off other permissions and is unaffected: the
+Content Collection widget on the node form, the Outline tab and
+Remove-from-outline (`add content to books`), creating a new collection
+(`create new books`), and the printer-friendly export
+(`access printer-friendly version`).
+
+One cosmetic exception: contrib's `BookOutlineConstraintValidator` uses the broad
+permission only to decide whether its "you can only change the book outline for
+the published version" violation message includes a link to the collections page.
+These roles now see the link-less variant of that message.
+
+`administer book outlines` is contrib book's administer-everything grant, and
+`editor`, `site_admin` and `platform_admin` held it only because these screens
+asked for it. Deleting a collection stays gated at the same level as reordering
+one, which is the access those roles already had; raising that bar would be a
+separate product decision.
+
+See `BookOutlineAccessCheck` and yalesites-org/YaleSites-Internal#1573.
+
 ## Running tests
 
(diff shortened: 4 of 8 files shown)


miketullo95 (approved):
@dblanken-yale Approving. I walked all the functional testing steps as the three roles and saw no degradation anywhere, and I audited the permission surface independently rather than taking the table in the description on trust.

I pulled contrib `book` 2.0.4 and enumerated every non-test reference to `administer book outlines`:

| Reference | After this PR |
| --- | --- |
| `book.routing.yml:15` (`book.admin`) | overridden by ys_book, re-gated |
| `book.routing.yml:52` (`book.admin_edit`) | already re-gated by #1145 |
| `book.routing.yml:80` (`book.node_child_ordering`) | now overridden by ys_book |
| `book.install:152,177` | one-time update hooks, not runtime access |
| `BookNodeOutlineAccessCheck.php:43,48` | both branches read `add content to books` OR the broad one, and all three roles keep the former |
| `BookOutlineConstraintValidator.php:76,92,108` | the cosmetic message-link variant you disclosed |

That is the complete list and it matches your table. The `BookNodeOutlineAccessCheck` one is the interesting entry, since it is the surface where a silent loss would have hidden, and it holds: the Outline tab and Remove-from-outline are unaffected for all three roles.

The load-bearing precondition checks out too. `editor`, `site_admin` and `platform_admin` all already carry `reorder book pages` (lines 383, 400 and 438 of their respective `config/sync` role files), so every re-gated screen stays reachable after the broad permission comes off. `contributor` has only `add content to books`, so it was already excluded, which is what I saw testing it.

Two things your description undersells:

The reinstated `book.node_child_ordering` override is not a straight revert of `0f8334bd4`. The version that was removed omitted contrib's `_custom_access` check; the one you have added includes it. The override is tighter than the one it restores, not merely equivalent.

`book.settings` at `/admin/structure/book/settings` is gated on `administer site configuration`, which no platform role holds. Worth stating plainly, because it confirms dropping the broad permission neither opens nor closes anything there.

On your first reviewer question: keep collection delete gated at the same level as reordering. It preserves exactly the access editors have today, which is the right default for a permissions cleanup, and no role currently exists that holds `reorder book pages` alone. If we ever create one, that is the moment to revisit the bar, not now. Not opening a follow-up for it.

On the second: dropping the permission from `platform_admin` as well is the right call for the same reason you gave.

Nothing blocking. Good call surfacing the delete coupling as a decision instead of letting it fall out of the change.

PR #1490, opened 2026-08-20
1563: Add config schema for ys_core.site, ys_core.header_settings, ys_core.footer_settings, and ys_core.social_links

dblanken-yale:
### Description of work

- Adds `config_object` schema for `ys_core.site`, `ys_core.header_settings`, `ys_core.footer_settings`, and `ys_core.social_links` in `ys_core/config/schema/ys_core.schema.yml`, following the existing `ys_core.dashboard_settings` entry as the pattern. Until now those four sitewide settings objects were unvalidated, so a malformed value could be written and nothing caught it.
- The schema describes **what the forms actually write**, not the tidier shapes the install YAML declared. That means `custom_favicon` / `site_name_image` are sequences of file IDs (a `managed_file` value), the `search.*` flags are integers, `content.text` is a `{value, format}` mapping, and the footer's logo/link collections are sequences of mappings that may be sparsely indexed. Confirmed by reading every `submitForm()` and by inspecting a real pulled production database.
- Includes the nine keys that existed only in live active config and were declared nowhere: `taxonomy.custom_vocab_name`, `font_pairing`, `cas_app_name` (site); `site_wide_branding_name`, `site_wide_branding_link`, `cta_content`, `cta_url` (header); `bluesky` (social links). Also covers `ys_core.site:search.enable_search_form`, which `ys_core_update_9002` superseded but which is still exported in `config/sync`, so it has to validate on import.
- Corrects the two install defaults that declared a shape their field never produces: `custom_favicon: '' -> []` and `site_name_image: '' -> []`. Both consumers guard on truthiness (`YaleSitesMediaManager`, `CoreTwigExtension`) and `[]` is falsy exactly as `''` was, so this is behavior-preserving.
- Casts `environment_indicator.show` to `(bool)` in `SiteSettingsForm`, matching the cast pattern `DashboardSettingsForm` already uses. The install default already shipped boolean `true` while the checkbox was storing integer `1`. A comment at the call site explains why this one checkbox is normalized while the sibling `search.*` flags stay integers.
- **Adds `ys_core_deploy_10008()`** (plus a `ys_core_normalize_settings_shapes()` helper) to bring already-stored values into the shape the new schema declares: scalar `custom_favicon` / `site_name_image` -> list, integer `environment_indicator.show` -> boolean. It is a **deploy hook, not a `hook_update_N`**, because it writes to active config and `drush deploy` runs `updatedb` *before* `config:import` — a config edit made in the update phase can be undone by the import that follows. `config_ignore` does currently list `ys_core*`, which would spare these objects, but depending on that is fragile; `deploy:hook` runs after the import and is correct either way. See "Why a normalisation hook" below.
- Un-blocks and writes the custom-favicon branch of the favicon coverage as `ys_core/tests/src/Kernel/CustomFaviconTest.php`, using the real shipped favicon image styles and a real saved `custom_favicon` so each size is proven to resolve to its **own distinct** URL — which the existing unit test could not show, because it stubs one style for all four sizes. Saving that config is exactly what the missing schema used to prevent.
- Adds `ys_core/tests/src/Kernel/CoreSettingsSchemaTest.php` pinning that all four objects have a schema and validate both as shipped and in the shapes the forms write, and `ys_core/tests/src/Kernel/SettingsShapeNormalizationTest.php` covering the normalisation helper.
- Updates the "Known test coverage gaps" section of `ys_core/README.md` to drop the two entries this closes.

### Why a normalisation hook (the AC's "flag either case" branch)

The ticket asks to fix the schema to describe reality *or* write an update hook if a stored value is genuinely wrong, and to flag which. Both apply here, and here is the evidence.

The schema was written to describe reality wherever the stored value was defensible (integers for the search flags, `NULL` for cleared media references, sparse sequence keys). But two shapes are genuinely wrong rather than merely different: `custom_favicon` and `site_name_image` shipped `''` as their install default, while the `managed_file` element that owns them only ever writes a list of file IDs. So essentially every existing site stores a scalar where a list belongs, and `environment_indicator.show` stores integer `1` where its own install default ships boolean `true`.

Those stale values will not wash out on their own: `config_ignore.settings` lists **`ys_core*`**, so `config:import` never rewrites these objects. Verified on a pulled database — active `ys_core.site` carries `taxonomy`, `custom_favicon`, `font_pairing`, `cas_app_name` and `environment_indicator`, none of which appear in the profile's `config/sync/ys_core.site.yml`. Hence the normalisation hook.

### No existing site breaks

This was checked rather than assumed, because it is the kind of claim that deserves evidence. Two things are true at once:

1. A legacy scalar `''` **does not validate** against the new `sequence` element, but it also does not throw. `StorableConfigBase::castValue()` only casts a scalar when the matched schema element is a `PrimitiveInterface`; a `Sequence` element falls through and the value is returned untouched, so `Config::save()` on a legacy site does not blow up. Core's `SchemaCheckTrait`, run against the real pulled database, reports `variable type is string but applied schema class is Drupal\Core\Config\Schema\Sequence` (and likewise integer vs `BooleanData` for the environment indicator). Controls confirm the checker really bites, so those are real.
2. That still breaks nothing, because **strict schema checking is a test-only facility**. On the running production-like container there is no `ConfigSchemaChecker` service and no `config.save` listener for it, and saving the legacy `''` then reading it back returns `''` unchanged with no exception. Nothing validates config on save or on import in production.

There is also a self-healing effect worth knowing about: once a config object has a schema, `Config::save()` casts the whole data array to the declared types, so these values correct themselves on any site that saves one of these forms again. The hook exists for the sites that never do.

So the schema is safe to land as-is, and the hook is a tidy-up that makes every existing site's stored data actually match it — not a fix for a live breakage.

**Correction (follow-up commit).** That held for the two shapes above, but review found a third one where it does not: a `focus_header_image` storing an **array** of media IDs — the shape `ys_core_preprocess_region()` has guarded against since a production hotfix. Unlike a scalar meeting a `sequence`, an array meeting a scalar `string` element does not fall through: `castValue()` recurses into it looking for `focus_header_image.0`, finds no such element, and throws an uncaught `InvalidArgumentException`. That aborts *any* save of `ys_core.header_settings` on such a site, `ys_core_deploy_10008()`'s own write included — a release blocked on one site. `ys_core_normalize_settings_shapes()` now repairs that value **first**, so the hook clears the bad shape instead of tripping over it. So for that one key the hook is not a tidy-up; it is what lets the deploy finish.

### Verification

- All three kernel tests pass, and each was confirmed to fail first for the right reason. `CoreSettingsSchemaTest` before the schema: 7 errors, every one `SchemaIncompleteException: No schema for ys_core.<object>` raised by `ConfigSchemaChecker`. `CustomFaviconTest` with the schema temporarily reverted: the two tests that save a `custom_favicon` error with `SchemaIncompleteException: No schema for ys_core.site`, while the fallback-only test still passes — which is precisely the workaround the README documented.
- Validated against a **real pulled production database**, not a fresh install, per the AC. The three objects absent from that database were created by saving the real settings forms in a browser on local Lando (`/admin/yalesites/header`, `/admin/yalesites/footer`, `/admin/yalesites/settings`); all saved cleanly, and all four objects then report `VALID` under `SchemaCheckTrait`. The `(bool)` cast was confirmed end to end there — `environment_indicator.show` is stored as boolean `true` after a real form save — and `ys_core.social_links` gained `bluesky: ''` from the real form, which is exactly why a key present in no install default still needed a schema entry.
- `ys_core/tests/src/Unit/YaleSitesMediaManagerTest.php`, the closest existing coverage of the touched code, is unchanged at 16 tests / 52 assertions before and after. No other test in any custom module references `SiteSettingsForm` or `environment_indicator`.
- The hook was verified against the **real pulled database**, not just in a kernel test. The pre-schema state was recreated by writing the legacy shapes straight to `config.storage` (which is what an already-populated database looks like), then: the dry run reported 3 pending changes and wrote nothing; the real run reported the same 3 and left both objects `VALID`; a second run reported 0 changes (idempotent); and the pulled state restored cleanly afterwards.
- `lando composer code-sniff` (Drupal + DrupalPractice over the whole custom tree): exit 0. Worth noting for future changes that this script runs two standards and stops at the first failure, so a clean Drupal run does not by itself mean DrupalPractice is clean.

### Notes for the reviewer

- The AC says the favicon test "now lives in `ys_media` after yalesites-org/yalesites-project#1475". That PR is **still open**, so `ys_media` does not exist on `develop` and `YaleSitesMediaManagerTest` is still in `ys_core`. The new test is therefore written in `ys_core`, where the code lives today, rather than basing this PR on an unmerged branch. It travels with the module when #1475 lands.
- Deliberately **not** adding `bluesky: ''` to `config/install/ys_core.social_links.yml`. The AC asks only for the schema entry; adding the install default would change shipped defaults for new sites, which is a separate documented inconsistency. Same reasoning for the other live-only keys.
- `CoreSettingsSchemaTest` saves the four settings objects directly instead of calling `installConfig(['ys_core'])`, because installing the whole `config/install` directory also installs the `grand_hero` `field_display_mode` field and so drags in `block_content`, `field` and `options`. That is unrelated to this ticket and is now recorded in the README so it is not rediscovered.
- `ys_core.views_settings` is a fifth ys_core settings object that still has no schema. It is outside this ticket's stated scope, so it is left alone here — worth a follow-up if we want the module fully covered. The README wording this PR adds ("All four settings objects now have a schema") is accurate as written.
- One forward-looking note: modern core idiom would add `requiredKey: false` to optional mapping keys, because core's `mapping` base type now carries `ValidKeys: '<infer>'`. Nothing fails today, because kernel tests only pass `validateConstraints` for core tests, not for custom or contrib. If that ever changes, `ys_core.social_links` (no `bluesky` in install defaults) and `ys_core.footer_settings` (`content.text: []` against a `value`/`format` mapping) would be the two to revisit.
- Honest bound on the live check: that site's `footer_variation` is `basic`, which disables the logos/links/text accordions, so the live form save left `content.logos` and `links.links_col_*` empty. Their sequence-of-mapping shapes are covered by `CoreSettingsSchemaTest` rather than by the live save.

### Functional testing steps:

- [ ] On a site with existing settings, run `drush deploy` and confirm `ys_core_deploy_10008` runs in the `deploy:hook` phase and reports how many values it normalised.
- [ ] `drush cget ys_core.site custom_favicon` — confirm it is a list (`{  }` when empty), not `''`, and `drush cget ys_core.site environment_indicator.show` is `true`/`false`, not `1`/`0`.
- [ ] Go to `/admin/yalesites/settings`, upload a custom favicon, and save. Confirm the save succeeds with no validation error and the favicon appears in the browser tab at all four sizes.
- [ ] Remove the custom favicon and save again. Confirm the default Yale favicon returns.
- [ ] As a platform admin, toggle the environment indicator at `/admin/yalesites/settings` off and on. Confirm the banner hides and shows accordingly.
- [ ] Save `/admin/yalesites/header` (try the focus header variation and a site name image) and `/admin/yalesites/footer` (try the mega variation with logos, two link columns and footer text). Confirm each saves cleanly and renders as before.
- [ ] Save `/admin/yalesites/social-links`, including a Bluesky URL. Confirm it saves and the icon renders.
- [ ] **Array `focus_header_image` (the deploy blocker).** On a site with header settings saved, put the bad shape in place with `drush php:eval "\Drupal::service('config.storage')->write('ys_core.header_settings', ['focus_header_image' => ['12']] + \Drupal::service('config.storage')->read('ys_core.header_settings'));"`, then `drush cget ys_core.header_settings focus_header_image` to confirm it reads back as a list. Run `drush deploy` (or `drush deploy:hook`) and confirm it **completes** rather than aborting with `The configuration property focus_header_image.0 doesn't exist`, and that the hook's summary counts the cleared value.
- [ ] `drush cget ys_core.header_settings focus_header_image` again — confirm the key is now gone, and that `/admin/yalesites/header` still loads with the focus variation selectable and a new image savable.
- [ ] Re-run `drush deploy:hook`. Confirm `ys_core_deploy_10008` reports 0 changes the second time (idempotent) and that a scalar `focus_header_image` set through the form is left untouched.
- [ ] **Legacy `site_name_image` as a `site_admin`.** With `ys_core.header_settings:site_name_image` still holding the legacy `''`, log in as a plain `site_admin` (not a platform admin, so the sitewide-branding fields are hidden), save `/admin/yalesites/header`, and confirm the save succeeds — the form leaves the legacy scalar in place, and it must pass through the new `sequence` element un-cast rather than erroring.

References yalesites-org/YaleSites-Internal#1563

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/README.md
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/README.md
@@ -39,24 +39,13 @@ someone runs them locally.
 Recorded here so they are visible to anyone working in this module, particularly
 during the staged cleanup tracked in yalesites-org/YaleSites-Internal#579.
 
-- **No config schema for this module's settings objects.** `config/schema/ys_core.schema.yml`
-  covers only `ys_core.dashboard_settings`. `ys_core.site`, `ys_core.header_settings`,
-  `ys_core.footer_settings`, and `ys_core.social_links` have none, so a kernel test
-  that installs this module's config fails with `SchemaIncompleteException` under
-  PHPUnit's default strict schema checking. That blocks kernel-level characterization
-  of every service that reads these objects, so the tests here work around it by
-  exercising only code paths that need no config save. Adding the schema is not
-  test-only work: `environment_indicator.show` ships as boolean `true` in
-  `config/install` but `SiteSettingsForm` saves integer `1`, and `custom_favicon` /
-  `site_name_image` are declared as `''` but hold arrays of file IDs, so no schema
-  type validates both the install defaults and real saved values without also
-  correcting those.
-- **`getFavicons()` custom-favicon branch is only partly pinned.** The test for it
-  stubs a single image style for all four sizes, so it cannot show that each size
-  resolves to its own distinct URL. Verifying that needs a real image style and a
-  saved `custom_favicon` value, so it is blocked on the missing schema above. The
-  fallback branch is covered against the real filesystem and the shipped image
-  style config in `YaleSitesMediaManagerTest`.
+- **`installConfig(['ys_core'])` needs more modules than a settings test wants.**
+  All four settings objects now have a schema, so saving them under strict schema
+  checking works and `CoreSettingsSchemaTest` pins that. Installing the *whole*
+  `config/install` directory additionally installs the `grand_hero`
+  `field_display_mode` field, which pulls in `block_content`, `field` and
+  `options` and needs the block type to exist. A kernel test that only needs the
+  settings objects should save them directly, as `CoreSettingsSchemaTest` does.
 - **`CoreTwigExtension::getAssetPath()`'s `_yale-packages` fallback is uncovered.**
   A normal checkout has an asset manifest under the theme's `node_modules`, so the
   first path always wins; the fallback only runs where the theme's npm assets are
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/config/install/ys_core.header_settings.yml
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/config/install/ys_core.header_settings.yml
@@ -1,7 +1,7 @@
 header_variation: 'basic'
 nav_position: 'left'
 focus_header_image: ''
-site_name_image: ''
+site_name_image: []
 search:
   enable_search_form: 1
   enable_cas_search: 0
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/config/install/ys_core.site.yml
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/config/install/ys_core.site.yml
@@ -6,6 +6,6 @@ seo:
   google_analytics_id: ''
 image_fallback:
   teaser: ''
-custom_favicon: ''
+custom_favicon: []
 environment_indicator:
   show: true
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/config/schema/ys_core.schema.yml
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/config/schema/ys_core.schema.yml
@@ -20,3 +20,219 @@ ys_core.dashboard_settings:
     announcements_source_term:
       type: string
       label: 'Taxonomy term name used to tag announcement posts'
+
+ys_core.site:
+  type: config_object
+  label: 'YaleSites site settings'
+  mapping:
+    page:
+      type: mapping
+      label: 'Landing pages'
+      mapping:
+        posts:
+          type: string
+          label: 'Relative path of the posts landing page, for example /post'
+        events:
+          type: string
+          label: 'Relative path of the events calendar page, for example /events'
+    search:
+      type: mapping
+      label: 'Search (superseded by ys_core.header_settings by ys_core_update_9002; still exported)'
+      mapping:
+        enable_search_form:
+          type: integer
+          label: 'Show the search form'
+    seo:
+      type: mapping
+      label: 'SEO'
+      mapping:
+        google_site_verification:
+          type: string
+          label: 'Google site verification token'
+        google_analytics_id:
+          type: string
+          label: 'Google Analytics ID (unused; Google Tag Manager replaced it)'
+    image_fallback:
+      type: mapping
+      label: 'Fallback images'
+      mapping:
+        teaser:
+          type: string
+          label: 'Media ID used when a teaser has no image'
+    taxonomy:
+      type: mapping
+      label: 'Taxonomy'
+      mapping:
+        custom_vocab_name:
+          type: string
+          label: 'Editor-facing label of the custom vocabulary'
+    custom_favicon:
+      type: sequence
+      label: 'Uploaded favicon (a managed_file value, so a list holding at most one file ID)'
+      sequence:
+        type: integer
+        label: 'File ID'
+    font_pairing:
+      type: string
+      label: 'Sitewide font pairing'
+    cas_app_name:
+      type: string
+      label: 'CAS application name used for Yale search'
+    environment_indicator:
+      type: mapping
+      label: 'Environment indicator'
+      mapping:
+        show:
+          type: boolean
+          label: 'Show the environment indicator banner'
+
+ys_core.header_settings:
+  type: config_object
+  label: 'YaleSites header settings'
+  mapping:
+    header_variation:
+      type: string
+      label: 'Header variation'
+    nav_position:
+      type: string
+      label: 'Navigation position'
+    focus_header_image:
+      type: string
+      label: 'Media ID of the focus header image'
+    site_name_image:
+      type: sequence
+      label: 'Site name image (a managed_file value, so a list holding at most one file ID)'
+      sequence:
+        type: integer
+        label: 'File ID'
+    site_wide_branding_name:
+      type: label
+      label: 'Sitewide branding name'
+    site_wide_branding_link:
+      type: string
+      label: 'Sitewide branding link'
+    cta_content:
+      type: label
+      label: 'Call to action text'
+    cta_url:
+      type: string
+      label: 'Call to action link'
+    dropdown_button_title:
+      type: label
+      label: 'Utility navigation dropdown button title'
+    search:
+      type: mapping
+      label: 'Search'
+      mapping:
+        enable_search_form:
+          type: integer
+          label: 'Show the search form'
+        enable_cas_search:
+          type: integer
+          label: 'Offer the CAS-restricted search scope'
+        enable_all_yale_search:
+          type: integer
+          label: 'Offer the all-of-Yale search scope'
+
+ys_core.footer_settings:
+  type: config_object
+  label: 'YaleSites footer settings'
+  mapping:
+    footer_variation:
+      type: string
+      label: 'Footer variation'
+    content:
+      type: mapping
+      label: 'Footer content'
+      mapping:
+        logos:
+          type: sequence
+          label: 'Footer logos'
+          sequence:
+            type: mapping
+            label: 'Footer logo'
+            mapping:
+              logo:
+                type: string
+                label: 'Media ID of the logo'
+              logo_url:
+                type: string
+                label: 'Link the logo points to'
+        school_logo:
+          type: string
+          label: 'Media ID of the school logo'
+        school_logo_url:
+          type: string
+          label: 'Link the school logo points to'
+        text:
+          type: mapping
+          label: 'Footer text'
+          mapping:
+            value:
+              type: text
+              label: 'Text'
+            format:
+              type: string
+              label: 'Text format'
+    links:
+      type: mapping
+      label: 'Footer links'
+      mapping:
+        links_col_1_heading:
+          type: label
+          label: 'Heading of the first links column'
+        links_col_2_heading:
+          type: label
+          label: 'Heading of the second links column'
+        links_col_1:
+          type: sequence
+          label: 'Links in the first column'
+          sequence:
+            type: mapping
+            label: 'Footer link'
+            mapping:
+              link_url:
+                type: string
+                label: 'Link URL'
+              link_title:
+                type: label
+                label: 'Link title'
+        links_col_2:
+          type: sequence
+          label: 'Links in the second column'
+          sequence:
+            type: mapping
+            label: 'Footer link'
+            mapping:
+              link_url:
+                type: string
+                label: 'Link URL'
+              link_title:
+                type: label
+                label: 'Link title'
+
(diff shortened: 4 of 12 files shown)


dblanken-yale:
Addressed both correctness findings from review on this branch. Pushed as `13619bdaa`; the schema, form round-tripping, and the deploy-hook-over-`hook_update_N` choice are unchanged.

### Finding 1 (deploy blocker) — fixed

`focus_header_image` is declared `type: string`, but `ys_core.module` documents and defends against that value being an **array** ("There are cases where the id is really an array of images over one") — a guard added by production hotfix `baa95d114`, so the array shape is real on live sites.

Verified the mechanism in core rather than taking it on trust: `Config::save()` casts only when a schema exists, so pre-schema these arrays passed through `validateValue()` harmlessly. Post-schema, `castValue('focus_header_image', ['12'])` recurses with lookup key `focus_header_image.0`; `ArrayElement::get()` finds a `StringData`, sets `$element = NULL`, and throws an uncaught `InvalidArgumentException`. Reproduced it locally as the first (failing) test: the trace lands on `ys_core.deploy.php:397`, the `->set($key, $to)->save()` line — so `ys_core_deploy_10008()` would itself have aborted `drush deploy` on such a site, because that save serialises the whole `ys_core.header_settings` object including the bad key. The existing self-heal only fires on a header render, which never happens during a deploy.

`ys_core_normalize_settings_shapes()` now plans that repair **first**, so the bad shape is cleared before the `site_name_image` write that would otherwise trip over it. It reuses the render path's `_ys_core_clear_focus_header_image_config()` rather than duplicating the clear; that helper gained a `$notify` flag so the deploy path skips building an admin-route URL for a message no one is there to read. The plan-then-write structure, the dry-run mode, and idempotency all still hold.

### Finding 2 (untested BC claim) — fixed

The claim was correct — `castValue()` coerces only for a `PrimitiveInterface`, and `Sequence` is not one, so a legacy scalar `''` meeting the new sequence element passes through un-cast rather than throwing — but you were right that nothing tested it. `storeLegacy()` writes straight to `config.storage`, bypassing `Config::save()` entirely, and every other path replaced the legacy value before saving.

New `ys_core/tests/src/Kernel/LegacyConfigShapeSaveTest.php` covers it, along with the Finding 1 regression: 5 tests, 12 assertions. It needs `$strictConfigSchema = FALSE`, because `ConfigSchemaChecker` validates the data *post*-cast and correctly rejects the scalar — so the assertion cannot be made in the existing test classes at all. DrupalPractice has a sniff against that property; it is suppressed for this one class with the rationale on the property, since the sniff's remedy ("declare your schema properly") is exactly what this PR does and no declaration can make a pre-schema value valid.

The most useful test of the pair is `testArrayFocusHeaderImageIsClearedByNormalization`, whose fixture is the combination a real site holds — the array *and* the legacy scalar together. That is what pins the ordering: with the repairs in the other order the hook throws.

### Optional polish

- Took the first one. `CoreSettingsSchemaTest::testDeclaredTypesAreWhatIsStored()` now feeds in the shapes a Form API element hands over (`'1'`, `1`, `['9']`) and asserts they read back as `1`, `TRUE`, `[9]` — so the integer, boolean and sequence declarations are shown to be the types actually stored, not merely types the schema will accept.
- Left the second one. `ys_core.site` in the profile's `config/sync` does validate today, and pinning it would mean a test that reads outside the module it covers; noting it here instead.
- Also folded the two normalisation test classes' duplicated legacy-fixture helpers into `tests/src/Traits/LegacyConfigFixtureTrait.php`, which resolves the deploy-file include through the extension list rather than a `../../../` hop.

### Verification

Ran only the affected files, sequentially. Baseline first: `LegacyConfigShapeSaveTest` failed 3/5 before the fix, every failure the exact `The configuration property focus_header_image.0 doesn't exist` at `ys_core.deploy.php:397` — and the Finding 2 test passed on the baseline, which is what confirms that half of the claim was already true. After the fix:

| File | Result |
|---|---|
| `LegacyConfigShapeSaveTest` (new) | 5 tests, 12 assertions — OK |
| `CoreSettingsSchemaTest` | 8 tests, 17 assertions — OK |
| `SettingsShapeNormalizationTest` | 7 tests, 19 assertions — OK |
| `CustomFaviconTest` | 3 tests, 29 assertions — OK |

`lando composer code-sniff` (Drupal + DrupalPractice over the whole custom tree): exit 0. `develop` merged in before and after the work; no conflicts. Testing steps for both new behaviours are appended to the description, and the "No existing site breaks" section now carries a correction, since that claim held for the two original shapes but not for this third one.

Not re-running `/security-review` — the added code writes config during a deploy and adds tests; it touches no auth, user input, secrets, external calls, uploads or SQL, and the review was clean on the previous diff.

References yalesites-org/YaleSites-Internal#1563

miketullo95 (approved):
@dblanken-yale Approving. Every AC on yalesites-org/YaleSites-Internal#1563 is covered, and the deploy hook does not need to wait for manual testing to be believed. It is already verified in CI.

The Pantheon build runs `drush deploy -v -y`, so `ys_core_deploy_10008` ran on this branch's multidev against the dev-seeded database. From the first run's log:

```
[notice] Deploy hook started: ys_core_deploy_10008
[notice] Normalised stored ys_core settings value: ys_core.site:environment_indicator.show 1 => true
[notice] Normalised 1 stored ys_core settings value(s) to the shape the config schema now declares.
[notice] Performed: ys_core_deploy_10008
[success] Finished performing deploy hooks.
```

The deploy completed, the count is reported as designed, and the second run on the follow-up commit reported `No pending deploy hooks` with the site still serving 200s. That covers the first functional testing step outright. Two caveats on what it does not cover: dev's data had no legacy scalar in either managed_file key and no array `focus_header_image`, so neither of those branches was exercised there. Those still want the manual pass.

## One thing to watch on merge, not blocking

`ys_core/tests/src/Kernel/CustomFaviconTest.php` does `use Drupal\ys_core\YaleSitesMediaManager;` and resolves its fixture paths through `extension.list.module')->getPath('ys_core')`. #1475 moves that class into `ys_media`. The two PRs touch different file paths, so git will merge whichever lands second without a conflict and quietly leave a ys_core kernel test pointing at a class that has moved. With `composer unit-test` still a stub, nothing in CI would catch it.

Your note says the test travels with the module when #1475 lands, which is the right plan. I am just flagging that nothing enforces it, so whoever merges second should move the file and its namespace deliberately rather than relying on a clean merge to mean anything. Worth knowing that #1563's AC was written expecting #1475 to have landed first, and it has not.

## Smaller note

`ys_core_deploy_10008()` clears an array `focus_header_image` with `$notify = FALSE`. The interactive path tells the editor the image is gone and links them to the header settings form; the deploy path is silent. The practical impact is close to nothing, since that shape never rendered an image anyway, but a site builder on an affected site will find an empty field and no explanation for it. Fine to leave as-is. If any real site turns out to be affected when this deploys, that is worth a line in the release notes rather than a code change.

Nothing blocking from me. The evidence in the description held up everywhere I checked it.

PR #1491, opened 2026-08-20
1576: Prune the unused MathJax payload from the Pantheon artifact

dblanken-yale:
### Description of work

- Adds `ScriptHandler::pruneMathJaxLibrary()` and registers it on the root `composer.json`
  `post-install-cmd` and `post-update-cmd`. After composer places the self-hosted MathJax
  package it removes `unpacked/`, `test/` and `docs/`, which nothing loads at runtime.
- **Measured effect: `web/libraries/MathJax` goes from 66 MB / 3,147 files to 44 MB / 1,863
  files.** `unpacked/` alone is 22 MB / 1,264 files — an unminified mirror of the packed tree
  that `MathJax.js` never loads. What a page actually requests is unchanged, because these
  directories were never requested in the first place.
- **To be clear about what is *not* being claimed:** the remaining 44 MB is mostly `fonts/`,
  `jax/`, `config/` and `localization/`, which MathJax fetches lazily per formula. Very little
  of it is requested by any single page, but it all has to be present — so this is not a
  "trim to what a page requests", and the comments say so explicitly to stop someone
  "finishing the job" later. `extensions/a11y/` in particular stays: turning on the
  accessibility explorer pulls several MB from it.
- `web/libraries` is gitignored but force-added into the Pantheon artifact by the deploy job
  (`git add -f pantheon.upstream.yml vendor web`), so all 66 MB was being committed on every
  deploy. Composer has no way to filter paths inside a package, which is why this is a script
  rather than a composer config change — as the ticket anticipated.
- **Verified from a clean state, which is what CI actually does.** Deleting
  `web/libraries/MathJax` entirely and re-running `lando composer install` made composer
  re-download and re-extract the full package, and the hook pruned it back to 44 MB / 1,863
  files in the same run (exit 0). So the fresh-runner path — no pre-existing `web/libraries` —
  is confirmed end to end, not just the already-installed path.
- **`docs/` does not exist in the pinned 2.7.9 npm distribution.** The ticket listed it, so it
  is still handled, but that means the "directory already absent" path is hit on the very
  first clean install rather than only on a re-run — the idempotency requirement is real, not
  hypothetical.
- Registered on **both** composer events on purpose. Composer fires exactly one of the two per
  command (`$eventName = $this->update ? POST_UPDATE_CMD : POST_INSTALL_CMD`), and the deploy
  path builds the artifact with `composer update` (`.ci/build/build_frontend`) while the CI
  test job and local installs use `composer install`. Both are needed for coverage.
- Fails soft: a failed removal is reported as a warning and the build continues, since the only
  thing at stake is artifact size. It is reported, not swallowed. A guard also makes the whole
  hook a no-op unless `web/libraries/MathJax/MathJax.js` is actually present.
- The `catch` is deliberately `\Throwable` rather than `IOException`, which review showed was
  too narrow to honour that intent. `Filesystem::remove()` renames the target to a `.!XXXX`
  sibling and then builds a `\FilesystemIterator` over it unguarded
  (`vendor/symfony/filesystem/Filesystem.php:192`); that constructor throws
  `\UnexpectedValueException`, which is not an `IOException`. With the narrow catch it would
  have escaped, and `set -eo pipefail` in `.ci/build/build_frontend` would have failed the
  deploy — over a size optimization. Worse, Symfony renames the directory back on the
  `IOException` path but not that one, so a mangled `.!XXXX` directory would have been
  force-added into the artifact on top of an unpruned tree. The warning now names the library
  path so a leftover is findable.
- Extends the existing `MathjaxLibraryInstallTest` with both halves of the contract: the three
  directories are gone, **and** the runtime tree (`config/`, `extensions/`, `fonts/`, `jax/`,
  `localization/`) survives. The second test is the one that would catch an over-reaching
  prune, whose only symptom would be LaTeX sitting unrendered on the page.
- Updates the `ys_mathjax` README developer note to say the install is pruned and what a fresh
  `composer install` should produce.

### What I could not verify, and want a second pair of eyes on

- **That the Pantheon artifact actually shrinks.** I cannot run a Pantheon build locally, so
  this is a traced argument rather than an observed artifact. What I did confirm: the hook
  fires on `composer install` locally (with output); the artifact's `web/libraries` comes from
  the root-level `composer update --prefer-dist` inside `.ci/build/build_frontend`, which
  fires `post-update-cmd`; that happens *before* the `git add -f ... web` step that commits it;
  and the later `composer remove` runs from the profile directory, whose `composer.json` has no
  `scripts` block, so it cannot un-prune. Worth a glance at the first multidev build to confirm.
- **These tests are not a CI gate.** Root `composer unit-test` is `echo 'No unit test step
  defined.'`, so `MathjaxLibraryInstallTest` — new assertions included — does not run in CI at
  all. It is a local guard. Pre-existing, out of scope here, but I would rather say it than let
  the new tests read as protection they do not provide.

### Independently checked during review

A reviewer re-derived the numbers from the pinned upstream tarball rather than trusting mine —
it fetched `mathjax-2.7.9.tgz`, confirmed its shasum matches the `dist.shasum` pinned in the
root `composer.json`, and measured: whole package 3,147 files, `unpacked/` 1,264 files (34% of
the payload, so "roughly a third" holds), remainder **1,863 files / ~44 MB**. Those match what
the prune actually produces here. It also confirmed nothing in custom code or contrib
references the pruned directories — contrib `mathjax.module:65` only ever wants
`/libraries/MathJax/MathJax.js` — and that both phpcs standards pass on the test file with
`--no-cache`, ruling out the stale-cache trap that can make a per-file run disagree with CI.

One correction that came out of it: `unpacked/` mirrors the whole packed tree, not just `jax/`,
`extensions/` and `config/` as I first wrote (it also carries `localization/`, `MathJax.js` and
`latest.js`). Comment and README both corrected.

### Notes for the reviewer

- A reasonable alternative would be an `rm -rf` line in the deploy workflow next to the
  existing `.git`-stripping step, right before `git add -f`. I went with the composer hook
  because any composer call added after such a step would silently re-fatten the artifact, it
  also slims local checkouts, and it is what lets the new assertions be checkable at all.
  Happy to move it if you would rather keep artifact concerns in the artifact-only place.
- `ScriptHandler.php` lives in `scripts/`, which `composer code-sniff` does not scan, so the
  new method deliberately follows that file's existing style (Allman braces, `//` comments
  rather than docblocks) rather than Drupal standard. Please don't "fix" it to match Drupal.
- Heads-up for anyone debugging math later: `MathJax.js` tells developers to load
  `unpacked/MathJax.js` when a rendering problem needs an unminified trace, and the prune
  removes it. `composer reinstall mathjax/mathjax --no-scripts` brings it back without
  re-running the prune; a plain `composer install` will **not**. The reason is worth knowing
  because it also explains why the README's other advice ("no `web/libraries/MathJax`? run
  `composer install`") is still correct: `LibraryInstaller::isInstalled()` treats a package as
  present when it is recorded in the lock **and** its install directory is readable, and never
  looks inside. So a missing package directory is reinstalled, while a missing `unpacked/`
  inside it goes unnoticed. Verified both directions locally (reinstall -> 66 MB / 3,147 files
  with `unpacked/` back; `composer install` -> pruned to 44 MB / 1,863).
- Commits two and three on this branch are self-corrections from review: I first documented that
  recovery recipe wrongly, then explained the mechanism behind it wrongly in a way that
  contradicted a neighbouring bullet. Both are fixed; flagging them so the commit history reads
  as intended rather than as churn.
- Side benefit: `test/` was 20 upstream sample HTML/JS files that were force-added into the
  artifact and therefore publicly reachable at `/libraries/MathJax/test/` on every deployed
  site. Pruning it is a small reduction in public attack surface, as is dropping `unpacked/`
  (an unminified mirror of the runtime code) out of the docroot.

### Functional testing steps:

- [ ] Run `lando composer install` on a checkout that has `web/libraries/MathJax`. Confirm the
      output includes `Pruned unused MathJax directory: unpacked` and `... test`, and that
      `du -sh web/libraries/MathJax` reports ~44 MB rather than ~66 MB.
- [ ] Run `lando composer install` a second time. Confirm it exits 0, prints no prune lines
      (nothing left to remove), and leaves the tree unchanged — the step must be idempotent.
- [ ] Visit `/admin/reports/status` and confirm there is **no** MathJax "local library files
      could not be found" error. (Sanity check that this is a real assertion: temporarily
      renaming `web/libraries/MathJax` makes that error appear.)
- [ ] Add a Text block containing inline `\(a^2 + b^2 = c^2\)`, display `$$\sum_{i=1}^{n} i$$`,
      a `$$\begin{pmatrix} a & b \\ c & d \end{pmatrix}$$` matrix, and
      `$$\int_0^1 x^2 \, dx = \tfrac{1}{3}$$`. Confirm all four typeset correctly and none are
      left as raw LaTeX.
- [ ] With that page open, confirm each rendered expression contains an `.MJX_Assistive_MathML`
      element (inspect the DOM), and that the browser network tab shows no 404s under
      `/libraries/MathJax/`.
- [ ] Confirm `https://<site>/libraries/MathJax/extensions/a11y/accessibility-menu.js` returns
      200 — the a11y extensions are what the 2.7.1+ pin exists for. (MathJax loads it lazily on
      first context-menu use, so it will not show in the initial network log.)
- [ ] Confirm `/libraries/MathJax/unpacked/MathJax.js` and `/libraries/MathJax/test/index.html`
      now return 404.
- [ ] Run `lando ssh -c "php /app/vendor/bin/phpunit -c /app/phpunit.xml /app/web/profiles/custom/yalesites_profile/modules/custom/ys_mathjax/tests/src/Unit/"`
      and confirm it is green, including `testUnusedUpstreamDirectoriesArePruned` and
      `testThePruneKeepsTheRuntimeTree`.
- [ ] On the multidev build for this branch, spot-check that the deployed
      `web/libraries/MathJax` has no `unpacked/` or `test/` directory.

References yalesites-org/YaleSites-Internal#1576

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -171,10 +171,12 @@
     ],
     "lint:php": "./.ci/test/static/lint_php",
     "post-install-cmd": [
-      "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
+      "DrupalProject\\composer\\ScriptHandler::createRequiredFiles",
+      "DrupalProject\\composer\\ScriptHandler::pruneMathJaxLibrary"
     ],
     "post-update-cmd": [
-      "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
+      "DrupalProject\\composer\\ScriptHandler::createRequiredFiles",
+      "DrupalProject\\composer\\ScriptHandler::pruneMathJaxLibrary"
     ],
     "post-create-project-cmd": [
       "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
--- a/scripts/composer/ScriptHandler.php
+++ b/scripts/composer/ScriptHandler.php
@@ -47,6 +47,65 @@ class ScriptHandler
     }
   }
 
+  // Trims the self-hosted MathJax package down to what pages actually request.
+  // web/libraries is gitignored, but CI force-adds it into the Pantheon
+  // artifact, so whatever composer leaves here gets committed - the whole
+  // 66 MB, when a math-heavy page only ever pulls about 448 KB. unpacked/ is
+  // an unminified mirror of the packed tree that MathJax.js never loads, and
+  // on its own is roughly a third of the payload;
+  // test/ is the upstream sample suite, otherwise publicly reachable at
+  // /libraries/MathJax/test/. Composer cannot filter paths inside a package,
+  // so this has to run after it places them.
+  //
+  // Registered on both post-install-cmd and post-update-cmd because composer
+  // fires exactly one of the two per command: the deploy build uses
+  // `composer update` (.ci/build/build_frontend) while the test job and local
+  // installs use `composer install`. Re-running against an already-pruned
+  // tree is the normal local case, so it has to be a no-op.
+  //
+  // The package is required by the profile's composer.json; only its
+  // repository definition and this hook sit at root, because composer runs
+  // root-package scripts only.
+  public static function pruneMathJaxLibrary(Event $event)
+  {
+    $fs = new Filesystem();
+    $library = static::getDrupalRoot(getcwd()) . '/libraries/MathJax';
+
+    // Nothing to do when the library was never installed in this checkout,
+    // and a safe stop if the pin ever moved to MathJax 3.x, which ships no
+    // root MathJax.js and none of the directories below.
+    if (!$fs->exists($library . '/MathJax.js')) {
+      return;
+    }
+
+    // docs/ only exists in some upstream builds, not in the pinned 2.7.9 dist.
+    // Filesystem::remove() is already a no-op on a missing path, so the
+    // exists() check is here only to keep the log line to what was removed.
+    foreach (['unpacked', 'test', 'docs'] as $dir) {
+      $path = $library . '/' . $dir;
+      if (!$fs->exists($path)) {
+        continue;
+      }
+      try {
+        $fs->remove($path);
+        $event->getIO()->write("Pruned unused MathJax directory: $dir");
+      }
+      catch (\Throwable $e) {
+        // Deliberately broad. Failing to prune only costs artifact size, so
+        // this must never fail a build - and IOException alone is not a wide
+        // enough net: Filesystem::remove() renames the target aside and then
+        // builds a \FilesystemIterator over it, which throws
+        // \UnexpectedValueException (not an IOException) if a directory in
+        // the tree cannot be opened. Note that path leaves the renamed
+        // `.!XXXX` sibling behind, hence naming the directory here.
+        $event->getIO()->writeError(
+          "<warning>Could not prune MathJax $dir; check $library for a "
+          . "leftover .! directory: " . $e->getMessage() . "</warning>"
+        );
+      }
+    }
+  }
+
   // This is called by the QuickSilver deploy hook to convert from
   // a 'lean' repository to a 'fat' repository. This should only be
   // called when using this repository as a custom upstream, and
--- a/scripts/composer/ScriptHandler.php
+++ b/scripts/composer/ScriptHandler.php
@@ -47,23 +47,38 @@ class ScriptHandler
     }
   }
 
-  // Trims the self-hosted MathJax package down to what pages actually request.
-  // web/libraries is gitignored, but CI force-adds it into the Pantheon
-  // artifact, so whatever composer leaves here gets committed - the whole
-  // 66 MB, when a math-heavy page only ever pulls about 448 KB. unpacked/ is
-  // an unminified mirror of the packed tree that MathJax.js never loads, and
-  // on its own is roughly a third of the payload;
-  // test/ is the upstream sample suite, otherwise publicly reachable at
-  // /libraries/MathJax/test/. Composer cannot filter paths inside a package,
-  // so this has to run after it places them.
+  // Removes the parts of the self-hosted MathJax package that the runtime
+  // never loads. web/libraries is gitignored, but the deploy job force-adds
+  // all of web into the Pantheon artifact
+  // (.github/workflows/build_deploy_and_test.yml), so whatever composer
+  // leaves on disk gets committed - all 66 MB of MathJax 2.7.9, across 3,147
+  // files.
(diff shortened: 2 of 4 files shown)


miketullo95 (approved):
@dblanken-yale Approving from a functional standpoint.

I checked the one thing you flagged as unverifiable, and the deployed artifact on `pr-1491-yalesites-platform.pantheonsite.io` is pruned:

- 200: `/libraries/MathJax/MathJax.js`, `/libraries/MathJax/latest.js`, `/libraries/MathJax/config/TeX-MML-AM_CHTML.js`, `/libraries/MathJax/jax/output/CommonHTML/jax.js`, `/libraries/MathJax/extensions/a11y/accessibility-menu.js`
- 404: `/libraries/MathJax/unpacked/MathJax.js`, `/libraries/MathJax/unpacked/latest.js`, `/libraries/MathJax/unpacked/extensions/tex2jax.js`, `/libraries/MathJax/test/index.html`

`latest.js` and `unpacked/latest.js` are siblings served the same way, so one at 200 and the other at 404 means the directory is genuinely absent from the artifact rather than a routing quirk. Both halves of the contract hold: the dead weight is gone and the runtime tree survived. That closes the "confirm the prune runs in CI and the Pantheon artifact actually shrinks" criterion on yalesites-org/YaleSites-Internal#1576, which was the last AC without evidence behind it.

Separate observation from the same check: `dev-yalesites-platform.pantheonsite.io` currently 404s on `/libraries/MathJax/MathJax.js` entirely, so the self-hosted library from #1481 does not look like it has landed on dev yet. Nothing to do with this PR, I just went looking for an unpruned baseline to compare against and did not find one. Worth a glance in case something went sideways with that deploy.

One open item, and it is not blocking the approval: your "Notes for the reviewer" asks whether to keep the composer hook or move the prune into the deploy workflow as an `rm -rf` next to the `.git`-stripping step. That is an implementation call rather than a product one, so I am not going to answer it. Since this PR came through your Claude agent, please give that question your own human pass before merging. My approval covers the outcome either way, not the specific placement.

PR #1492, opened 2026-08-20
1584: Bug: Beacon index strips all formatting from chunk content — convert rendered HTML to Markdown instead

dblanken-yale:
### Description of work

**What was wrong.** Everything Beacon hands the language model arrived as one undifferentiated block of plain text. Search API's `html_filter` processor was enabled on `rendered_item` in `search_api.index.ys_beacon.yml` and stripped every tag before the chunk was embedded, so the model could not tell a heading from a sentence. It was worse than plain text: adjacent list items and table cells were concatenated into single words (`TranscriptTwo letters of recommendation`, `DeadlineTermJan 15Fall`), and a page's `<script>` body was indexed verbatim as content.

**The fix turned out to be smaller than the issue assumed.** `ai_search` *already* converts every indexed field to Markdown at index time — `EmbeddingStrategyPluginBase` builds an `HtmlConverter`, switches on `strip_tags` and `strip_placeholder_links`, and registers a `TableConverter`, and `EmbeddingBase::getValue()` runs it over each value. That conversion was a no-op only because `html_filter` had already flattened its input. So this does not add a conversion; it stops destroying the markup and lets the existing one work.

- Removed `html_filter` from the Beacon index (`rendered_item` was its only configured field). Its per-tag boosts are not carried over — the index is served by a vector database, so relevance comes from the embedding and a boost on `<h1>` never affected retrieval.
- Added a `ys_beacon_indexable_html` processor at the same `preprocess_index` weight, backed by a new `IndexableHtmlFilter` service. It only **deletes**, in three passes:
  - **Non-content elements**, because each one otherwise survives conversion: a `<script>`/`<style>` body becomes indexed text, `<nav>` becomes a Markdown list of every menu link, `` becomes an image tag, and `<figure>` runs its caption into the next paragraph.
  - **Textless elements** (all but `<br>`, `<hr>` and table structure). Real output needs this: a `<h1 class="page-title__heading">` containing only whitespace still emitted its setext underline, putting a bare `=` line into the chunk, and empty field wrappers put a run of ~19 blank lines between two paragraphs — all charged against the 1024-token chunk budget.
  - **Dangerous link targets**, replaced by their own text. The converter stores a target verbatim and chat answers render as Markdown, so `[Click](javascript:alert(1))` would otherwise be a live link.
- Added the config schema entry for the new processor. The generic `plugin.plugin_configuration.search_api_processor.*` fallback only validates `weights`, so `fields`/`all_fields` would have been schema-invalid without it; it reuses search_api's own `search_api.fields_processor_configuration` type.

**Deliberately not done, and worth knowing why:** the filter does not emit Markdown itself, which is what the issue originally proposed. The value passes through that converter afterwards regardless, and its `TextConverter` escapes Markdown punctuation — an earlier revision produced `\## Admissions`, `\*\*full packet\*\*` and `\[portal\](...)`, all on one line. It also no longer unwraps layout wrappers: `strip_tags` already does that with correct block spacing, and doing it here meant inserting a separator space that split words abutting inline markup (`un<span>believable</span>` came out as `un believable`). Both failure modes now have regression tests.

**Measured effect** (all 18 published nodes on a local build, through the real processor chain and the real production-configured converter): characters +15.3%, estimated chunks 109 → 123 (+12.8%). Zero script/style leakage, zero stray `=` lines, zero blank-line runs.

---

**Second, unrelated change in this PR — Beacon model context window, Haiku 200k → Claude Sonnet 5 1M.** Bundled here at Dave's request; it is a separate commit and can be split out if it would hold up the fix above.

- `model_context_window` and `ChatApiController::DEFAULT_CONTEXT_WINDOW` raised to `1000000`, with the admin-form description and three comments updated off Haiku.
- Added `ys_beacon_deploy_10003()`. This part is easy to miss: `ys_beacon.settings` is config-ignored (`ys_beacon*`) **and** absent from `config/sync`, so the raised default in `config/install` reaches new installs only and `config:import` would never move an existing site off 200k. The hook raises the value only when it is still exactly the old default, so a deliberate per-site override survives.

Two caveats on that change, neither blocking:

- **`1000000` is operator-assumed, not measured.** Dave confirms the Portkey config selects Sonnet 5 (and `chat_models: ['gpt-4o']` is an intentional placeholder — the admin form and `PortkeyProvider` both document that Portkey picks the real model server-side). What is *not* verified is that the route's upstream serves Sonnet 5 at the full 1M window; long-context can be gated differently via Bedrock or Vertex, and the gateway's own body limit is independent of the model. A probe exists that would settle it empirically, but it spends real Portkey/Anthropic budget (~$0.60 for a single 300k rung), so it is left for Dave to approve rather than run speculatively.
- **The value is not load-bearing either way.** `MAX_PAYLOAD_BYTES` (1 MiB, ~262k tokens), `MAX_TRANSCRIPT_MESSAGES` (20) and `top_k: 5` cap the reachable worst case near 270k tokens, so at 1M the transcript-trim path effectively goes dormant and the 413 "conversation too long" branch becomes unreachable in practice. Not a regression, but that safety net is no longer the binding constraint. The only risky direction would be setting the window *below* the reachable ceiling.

### Functional testing steps:

- [ ] On a site with Beacon enabled, go to the Beacon settings (or Platform Admin Settings) and click **Re-index all content**, then **Index now**. No new code is needed to trigger this — `reindexAll()` re-queues every item and the provider deletes an entity's prior chunks before inserting the new ones.
- [ ] Ask Beacon a question whose answer lives on a page with headings, a bulleted list, and a table. Confirm the answer reflects the page's structure — that list items are treated as distinct items and tabular values are not run together.
- [ ] Ask a question about a page containing a table of dates or fees. Previously the cells arrived as `DeadlineTermJan 15Fall`; confirm the answer now reads the table correctly.
- [ ] Confirm citations still resolve to the right page and title, and that answers link using the cited source URL rather than an arbitrary in-body link.
- [ ] Check Beacon still answers normally on a page with embeds, images, or a video — those are dropped from the index by design, so the surrounding prose should still be found.
- [ ] Verify existing sites are unaffected until re-indexed: before clicking re-index, answers should be unchanged.
- [ ] For the context-window change: run `drush deploy` on a site still holding `200000` and confirm the log reports it raised to `1000000`; on a site with a deliberately different value, confirm it reports the value was left alone. Then confirm the Beacon administration form shows the new default and a long conversation still answers without error.

### Notes for the reviewer

- **This needs a re-index per site to take effect.** The change only alters what gets *written* to the index, so every existing Beacon site keeps serving plain-text chunks until someone clicks **Re-index all content**. Worth a release-note line.
- The end-to-end "do the answers actually get better" comparison through `ys_ai_tester` could not be run locally — it needs live Azure AI Search and Portkey credentials, and local secrets are placeholders. The mechanism is verified locally end-to-end (real processor chain → real converter); the answer-quality evidence belongs on a multidev.
- Tables now become real Markdown pipe tables, because `ai_search` registers the library's `TableConverter` even though the default environment omits it. The test suite builds the converter exactly as `EmbeddingStrategyPluginBase` does, and says why in a comment — a bare `HtmlConverter` would silently pin behaviour production never runs.
- Unrelated observation, no action needed: nodes 181-184 in the local content are migrated policy pages with Markdown **pasted as plain text** into text blocks (node 184 has 134 literal `**` and zero `<strong>` tags). `TextConverter` correctly backslash-escapes that literal punctuation so it is not misread as bold. Real `<strong>` still yields unescaped `**`.
- Beacon unit tests: 345 passing (baseline 328, plus 14 for the filter and 3 for the deploy hook). `composer code-sniff` and `composer lint:php` both exit 0.

References yalesites-org/YaleSites-Internal#1584

### DIFF
--- a/web/profiles/custom/yalesites_profile/config/sync/search_api.index.ys_beacon.yml
+++ b/web/profiles/custom/yalesites_profile/config/sync/search_api.index.ys_beacon.yml
@@ -100,27 +100,16 @@ datasource_settings:
 processor_settings:
   add_url: {  }
   entity_status: {  }
-  html_filter:
-    weights:
-      preprocess_index: -49
-      preprocess_query: -48
-    all_fields: false
-    fields:
-      - rendered_item
-    title: true
-    alt: true
-    tags:
-      b: 2.0
-      em: 1.0
-      h1: 5.0
-      h2: 3.0
-      h3: 2.0
-      strong: 2.0
-      u: 1.0
   rendered_item: {  }
   ys_beacon_ai_metadata: {  }
   ys_beacon_citation_fields: {  }
   ys_beacon_exclude_ai_disabled: {  }
+  ys_beacon_indexable_html:
+    weights:
+      preprocess_index: -49
+    all_fields: false
+    fields:
+      - rendered_item
 tracker_settings:
   default:
     indexing_order: fifo
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/config/install/ys_beacon.settings.yml
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/config/install/ys_beacon.settings.yml
@@ -18,12 +18,14 @@ top_k: 5
 score_threshold: 0.0
 streaming: true
 # Context window (in tokens) of the model Portkey currently routes to. Used to
-# window long conversations so they degrade gracefully. Defaults to Claude
-# Sonnet 5's 1,000,000-token window; update it in the Beacon administration form
-# when the routed model changes. Keep in sync with
-# ChatApiController::DEFAULT_CONTEXT_WINDOW, the runtime fallback used when this
-# value is absent.
-model_context_window: 1000000
+# window long conversations so they degrade gracefully. Set from a measurement
+# rather than the model's nominal window: the Portkey route (Claude Sonnet 5)
+# accepted 433437 input tokens in one request, so 400000 stays under the proven
+# figure while still exceeding the largest request Beacon can build. Update it
+# in the Beacon administration form when the routed model changes. Keep in sync
+# with ChatApiController::DEFAULT_CONTEXT_WINDOW, the runtime fallback used when
+# this value is absent.
+model_context_window: 400000
 instructions_max_length: 4000
 instructions_warning_threshold: 3500
 # Fallback used only when a site has no active system instructions. Keep in
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/config/install/ys_beacon.settings.yml
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/config/install/ys_beacon.settings.yml
@@ -19,11 +19,11 @@ score_threshold: 0.0
 streaming: true
 # Context window (in tokens) of the model Portkey currently routes to. Used to
 # window long conversations so they degrade gracefully. Defaults to Claude
-# Haiku's 200k window; update it in the Beacon administration form when the
-# routed model changes. Keep in sync with
+# Sonnet 5's 1,000,000-token window; update it in the Beacon administration form
+# when the routed model changes. Keep in sync with
 # ChatApiController::DEFAULT_CONTEXT_WINDOW, the runtime fallback used when this
 # value is absent.
-model_context_window: 200000
+model_context_window: 1000000
 instructions_max_length: 4000
 instructions_warning_threshold: 3500
 # Fallback used only when a site has no active system instructions. Keep in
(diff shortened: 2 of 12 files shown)


miketullo95 (approved):
@dblanken-yale Approving. Reviewed and the Markdown looks good.

One non-blocking cleanup before this merges:

**The PR description is out of date on the context window.** The body says `1000000` and describes the value as operator-assumed rather than measured, but the code sets `400000` from the measured 433,437-token request:

- `web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/config/install/ys_beacon.settings.yml` (`model_context_window: 400000`)
- `web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/src/Controller/ChatApiController.php:66` (`DEFAULT_CONTEXT_WINDOW = 400000`)
- `web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/ys_beacon.deploy.php` (`ys_beacon_deploy_10003()` raises 200000 to 400000)

The last functional testing step still tells the tester to confirm the log reports it raised to `1000000`, which will never match what the deploy hook actually logs. Please update the description and that step to `400000` so QA is not chasing a number that does not exist. The two caveats in that section about 1M being unverified need the same pass, since they no longer apply to a measured value.

Keeping the context-window change bundled here rather than splitting it. It is a separate commit with its own deploy hook and tests, so splitting costs a PR cycle for no testing benefit.

For my own tracking, no action needed from you: this takes effect only after a per-site re-index, so I will carry the release note and Release Testing Step at release prep.

PR #1498, opened 2026-08-21
1587: Bug: AI Tester marks long runs as failed when a single question hits a transient upstream error

dblanken-yale:
### Description of work

- **A run where some questions failed is now recorded `partial`.** `failed` is reserved for a run where nothing was answered. Previously one transient error out of a hundred questions recorded the same `failed` as a total collapse, so a 99-percent-successful run read as broken. No schema change or update hook: the status column is already `varchar(32)`.
- **Transient failures are retried, deterministic ones are not.** Up to two retries with jittered exponential backoff, honouring `Retry-After`. Retryable means 5xx, 429, 408, a connect or read timeout, or `AiRateLimitException`. A filtered completion, a malformed request, a bad key, an exhausted quota, or "the assistant is not available on this site" fails on the first attempt rather than the third. Classification reads HTTP status codes and the AI module's typed exceptions, never message wording, for the reason already documented in `PortkeyProvider::handleApiException()`: the wording varies by upstream provider and by library version while the codes stay stable.
- **Classification walks the `getPrevious()` chain**, which is required rather than defensive. `LegacyConversationClient::ask()` rethrows every Guzzle failure wrapped in a `RuntimeException`, so reading only the outermost exception would classify every legacy failure as permanent and retry none of them.
- **Every wait is bounded**, because these sleeps happen inside a batch request that the platform terminates at 120 seconds. Per-attempt caps plus a total per-question budget, so a gateway asking to be retried in five minutes cannot get the whole request killed and take the queued questions with it.
- **A configurable pause between questions**, on the Beacon administration form. The tester is the only path on the platform that bursts these APIs; the chat widget answers one question per visitor, already rate limits per IP, and its provider documents a deliberate no-sleep contract so a request never holds a PHP-FPM worker. The widget is untouched, and the field is contributed by the tester's own `hook_form_FORM_ID_alter()` so it disappears with the submodule.
- **Failures now log enough to diagnose.** openai-php's `ServerException` message is only `Server error (HTTP 500) occurred.` and discards the response body it is holding; Guzzle truncates response bodies to 120 characters when building an exception message, which is how the 400 in the issue was recorded cut off mid-word with its cause lost. Failures now log the status code, the response body bounded to 2000 bytes, and Portkey's `x-portkey-trace-id` / `x-portkey-request-id`, without which a failure cannot be matched against the gateway's own logs at all.
- **The log names which upstream call failed.** Chat and embeddings authenticate with different Portkey keys and can sit in different workspaces, so naming the stage is the difference between checking the right log and concluding the gateway saw nothing.
- **Closes a gap in the same mechanism for the legacy assistant.** It streams its reply, so it can report a failure inside an HTTP 200 body; those arrived as a bare `RuntimeException` with no status, meaning retry would have worked for Beacon and silently not for legacy even when the body said 429. `UpstreamStatusInterface` lets a backend report a status it parsed itself. Symbolic codes such as `content_filter` are deliberately not coerced into a status.

### Two things for the reviewer

**1. A deliberate exception to config ownership, flagged for you to overrule.** `ai_tester_question_delay_ms` is read and written only by `ys_ai_tester`, so it would normally belong in a `ys_ai_tester.settings` object. It is in `ys_beacon.settings` on purpose: that object is covered by the `ys_beacon*` pattern in `config_ignore.settings.yml` and so is absent from `config/sync`, which is what lets a value tuned during an incident survive the next `drush deploy` config import. A `ys_ai_tester.settings` object would be exported and re-imported, resetting the knob on every deploy unless a second ignore pattern were added to chase it. Being in that file also means `_ys_beacon_seed_settings()` backfills the new default onto existing sites for free. Reasoning is recorded at both the key and the submit handler.

**2. A separate blind spot this uncovered, not fixed here.** `RagRetriever::retrieve()` catches `Throwable` and degrades to an empty citation list, logging separately as "Beacon retrieval failed". So an embeddings or Azure AI Search outage during a tester run produces answers with **no sources** rather than errors — a run conducted during such an outage reads as "Beacon gives poor, uncited answers" instead of "retrieval was down", which is actively misleading for the evaluation work the tester exists to support. Not changed here because `RagRetriever` is shared with the live chat widget, where quiet degradation is deliberate and documented. Worth its own ticket to surface citation-free answers on the run view.

This also narrows the original diagnosis: because retrieval failures are swallowed, the `Server error (HTTP 500) occurred.` recorded against run 9 question 89 cannot have come from the embeddings call. It came from the chat call or the tool follow-up, both on `portkey_llm_api_key`.

### Functional testing steps:

- [ ] As user 1, visit `/admin/config/yalesites/ys-beacon/admin`. Confirm an **AI Tester** section with a **Pause between questions (milliseconds)** field, defaulting to 500, with a description explaining what it does.
- [ ] Change the pause to 1000 and save. Confirm it persists on reload, and confirm the other Beacon settings on that form (sources per answer, relevance score, streaming) still save correctly.
- [ ] Try to save the pause field empty. Confirm it is rejected rather than silently reverting to 500.
- [ ] Run the AI Tester with a short question list against Beacon. Confirm it still completes normally and the run history shows status `complete`.
- [ ] Run a list that includes one question you expect the assistant to refuse or filter, alongside questions that answer normally. Confirm the run is recorded **`partial`** rather than `failed`, that the successful answers are all saved and viewable, and that the completion message names which question numbers failed.
- [ ] Open the failed question's entry at `/admin/reports/dblog`. Confirm the log line now carries an HTTP status, a response body (not cut off at 120 characters), and, for a Beacon failure, which stage failed. Confirm no API key or credential appears in it.
- [ ] Confirm a `partial` run can still be re-run from the run history.
- [ ] Regression: confirm the Beacon chat widget on a page still answers normally and streams as before. It should be entirely unaffected by this change.

### Notes on verification

- Unit tests: `ys_ai_tester` 114 -> 190, `ys_ai_tester_legacy` 29 -> 37, `ys_beacon` 328 unchanged. All pass, no pre-existing failures in the baseline and none introduced.
- `composer lint:php` and `composer code-sniff` both exit 0 over the whole custom tree.
- The transient upstream failure in the issue cannot be induced locally, so the retry and classification paths are covered by unit tests over real `openai-php`, Guzzle, and AI-module exception objects rather than by a live reproduction. The form alter, its field bounds, its label association, and the preservation of the config form's own submit handler were each verified against the rendered form and the built form, not only in tests.

References yalesites-org/YaleSites-Internal#1587

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/config/install/ys_beacon.settings.yml
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/config/install/ys_beacon.settings.yml
@@ -80,3 +80,20 @@ fallback_system_prompt: |-
   - Support accessibility through clear communication
   - Prioritize helping users accomplish their goals efficiently
 guardrail_supplement: ''
+# Milliseconds the AI Tester pauses between questions. The tester is the only
+# path that issues a long burst of upstream requests - the chat widget answers
+# one question per visitor - so pacing lives here rather than in the shared
+# provider. Small by default because each question already costs seconds of
+# model latency; raise it if an upstream per-minute ceiling is suspected. Clamped
+# to 0-10000 by AiTesterRetry::questionDelayMs(); 0 disables pacing.
+#
+# Read and written only by the ys_ai_tester submodule, which is a deliberate
+# exception to keeping a module's settings in its own config object. It lives
+# here because this object is covered by the 'ys_beacon*' config_ignore pattern
+# and is therefore absent from the profile's config/sync: a value an operator
+# tunes during an incident survives the next `drush deploy` config import. A
+# ys_ai_tester.settings object would be exported and re-imported instead,
+# resetting the knob on every deploy unless a second ignore pattern were added
+# to chase it. Being here also means _ys_beacon_seed_settings() backfills this
+# default on existing sites for free, since it unions the whole shipped file.
+ai_tester_question_delay_ms: 500
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/config/schema/ys_beacon.schema.yml
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/config/schema/ys_beacon.schema.yml
@@ -77,3 +77,6 @@ ys_beacon.settings:
     guardrail_supplement:
       type: text
       label: 'Site-specific additions to the platform guardrail'
+    ai_tester_question_delay_ms:
+      type: integer
+      label: 'Milliseconds the AI Tester pauses between questions'
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/modules/ys_ai_tester/src/AiTesterBatch.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_beacon/modules/ys_ai_tester/src/AiTesterBatch.php
@@ -36,13 +36,15 @@ class AiTesterBatch {
     $citations = [];
     $error = '';
 
+    self::pace($delta);
+
     try {
       $backend = \Drupal::service('ys_ai_tester.answer_backend_registry')
         ->getAvailable($backend_id);
       if ($backend === NULL) {
         throw new \RuntimeException(sprintf('The "%s" assistant is not available on this site.', $backend_id));
       }
-      $result = $backend->answer($question);
+      $result = self::answerWithRetry($backend, $question, $run_id, $delta, $backend_id);
       $answer = $result['answer'];
       // Normalizing here rather than in each backend is what makes the stored
       // shape uniform: the formatter owns marker parsing, de-duplication, and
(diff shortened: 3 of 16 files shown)


miketullo95 (approved):
@dblanken-yale Approving.

Dave and I both ran the AI Tester against this and it worked great. Runs that would have come back `failed` before now finish and report accurately, and the answers we expected to be saved were all there.

I went through the diff against #1587 and everything in the acceptance criteria is covered:

- Retry is scoped to transient failures only, capped at two retries, with the deterministic cases (content filter, bad request, quota, missing assistant) failing on the first attempt. `AiTesterBatch::answerWithRetry()`.
- Waits are bounded on all three axes: per-attempt backoff, a capped `Retry-After`, and a total per-question budget. Given the 120 second request ceiling, giving up on one question rather than risking the whole batch is the right call.
- `partial` is derived the way the ticket asked, and it reserves `failed` for a run where nothing succeeded. `AiTesterBatch::runStatus()`. Keying failures by delta so a question that fails twice does not count twice is a good catch.
- The completion message now names the failed question numbers, and it numbers them from 1 to match what the user saw on screen.
- Logging carries status code, bounded response body, Portkey trace/request ids, and which upstream stage failed.
- Run status is interpolated as plain text in the run view and comparison view, so `partial` reads correctly with no template change and is not conveyed by color alone.

On your two flagged items:

**1. `ai_tester_question_delay_ms` living in `ys_beacon.settings`.** Not overruling it. Surviving `drush deploy` is the behavior we actually want for a knob we would reach for mid-incident, and chasing it with a second `config_ignore` pattern is the worse trade. The reasoning is documented at both the key and the submit handler, so the next person will find it.

**2. `RagRetriever::retrieve()` swallowing retrieval failures.** Agreed this is a real gap and agreed it does not belong in this PR. A tester run during an embeddings or Azure AI Search outage reading as "Beacon gives uncited answers" instead of "retrieval was down" would send us chasing answer quality for something that is an outage. I will get a ticket written for surfacing citation-free answers on the run view.

Non-blocking, nothing gating merge.

PR #1499, opened 2026-08-22
1589: Bug: Site Settings page loads scrolled down to the Fallback teaser image field

dblanken-yale:
### Description of work

- Site Settings (`/admin/yalesites/settings`) opened scrolled **1888px** down, landing on the Fallback teaser image field. Fixed at the source; the page now loads at `scrollY: 0` with Site name as the first visible field.
- **Root cause.** Nothing scrolled the page — a focus did. The "Add media" button was rendered both `visually-hidden` and flagged `data-disabled-focus="true"`, the contrib JS (`Drupal.behaviors.MediaLibraryFormElementDisableButton`) focuses any `.js-media-library-open-button[data-disabled-focus="true"]` on behavior attach, and focusing an offscreen element makes the browser scroll it into view. Confirmed by instrumenting `focus`, `scrollIntoView` and `scrollTo` in the browser: exactly one `focus()` call, on that button, at `scrollY` 0, immediately followed by the scroll. Nothing ever called `scrollTo`/`scrollIntoView` — which is why the ticket is right that a scroll-to-top script would be fighting the browser.
- The field is not core's media widget: `SiteSettingsForm` uses `'#type' => 'media_library'` from contrib **`media_library_form_element`**, which forked core's code and **dropped core's guard**. Core only sets that attribute when the widget is rebuilt from a selection change (core issue [#3091297](https://www.drupal.org/project/drupal/issues/3091297)); contrib set it on *every* build once the element was at cardinality, i.e. whenever an image was already selected.
- **Fix:** a composer patch on the profile porting core's guard verbatim — flag for focus only on a `media_library_update_widget` rebuild, otherwise set `#disabled` server-side. One patch instead of three form_alters, which is what makes the "fix it consistently rather than only on this one page" criterion hold: it also covers **Header Settings** (`/admin/yalesites/header`) and both **Footer Settings** logo fields (`/admin/yalesites/footer`), which had the same latent bug.
- **The accessibility behaviour the attribute exists for is preserved.** After a real media selection the flag is still set, so keyboard and screen reader users still get focus back on the control they left from. Verified in the browser and pinned by a test.
- Adds a `ys_core` kernel test (7 tests) and ys_core's first test-only module, `ys_core_test`.

### Reviewer notes

- **Patch application confirmed by the build** (was the one item I could not verify locally). The `Build frontend components` step of run [32543527729](https://github.com/yalesites-org/yalesites-project/actions/runs/32543527729) logs both patches applied to a freshly downloaded 2.1.4:
  ```
  - Applying patches for drupal/media_library_form_element
      patches/contrib/media_library_form_element/3168027-8.patch (Order Media items ...)
      patches/contrib/media_library_form_element/core-3091297-port-focus-guard.patch (Only shift focus ...)
  ```
  Worth knowing why local differs: in my checkout `PATCHES.txt` lists only the pre-existing patch, because that `composer.lock` is stale and every partial update aborts on drift unrelated to this PR (root `composer.json` asks core `10.6.15` / atomic `1.81.0`; the lock has `10.6.13` / `1.80.0`). `composer.lock` is untracked, so the build resolves fresh — which is what the log above shows. Anyone else seeing the bug persist locally after pulling this branch needs a full `composer update`, not just `composer install`.
- **This patch should go upstream.** It is a faithful port of a core fix and there is no matching issue in the `media_library_form_element` queue (the only focus-related one, #3509212, is a different concern), so today the patch has no path to ever being dropped. Filing that is outside what I did here.
- One deliberate deviation from core: `$triggering_element['#array_parents'] ?? []` instead of core's inline assignment-in-condition, which warns under PHP 8 if a triggering element ever lacks `#array_parents`. Equivalent across all four reachable input states.
- Pre-existing and untouched: `ys_core.libraries.yml.orig` is tracked in git and looks like a leftover merge artifact.

### Verification already run

- Kernel tests: **`OK (7 tests, 120 assertions)`**. Regression net measured rather than assumed — with the patch reverted the same file gives **`Tests: 7, Assertions: 116, Failures: 3`**, so 3 tests genuinely catch this bug and 4 are guard rails against over-correcting.
- CI lint gate: `composer lint:php` and `composer code-sniff` (Drupal + DrupalPractice) both **exit 0**. No SCSS touched.
- Accessibility gate: **PASS**. On load `activeElement` is `body` and `scrollY` is 0; the first 25 tab stops follow DOM order and never jump mid-form (2.4.3); no keyboard trap (2.1.2) — the now-disabled "Add media" button is correctly skipped while the field's **Remove** button stays reachable in both directions; text contrast in the region 4.5:1–21:1.
- Full interaction pass on the real form: initial load `scrollY` 0 → Remove (button re-enables, value clears) → open modal → Insert selected (flag correctly returns, value updates) → Save ("The configuration options have been saved.") → reload `scrollY` 0, value persisted. No JS errors at any step.

### Functional testing steps:

- [ ] With a fallback teaser image already set, open **YaleSites > Site Settings** (`/admin/yalesites/settings`). The page loads at the **top**, with **Site name** as the first visible field — it no longer jumps to Fallback teaser image.
- [ ] Same check on `/admin/yalesites/header` (focus header image set) and `/admin/yalesites/footer` (footer logos / school logo set) — both should also load at the top.
- [ ] On Site Settings, click **Remove** on the fallback teaser image. The **Add media** button becomes available again.
- [ ] Click **Add media**, pick an image, and **Insert selected**. The image appears and the Add media button goes away (the field only allows one).
- [ ] Press **Save configuration**, then reload the page. The image persisted, and the page still loads at the top.
- [ ] Keyboard check: from a fresh load press Tab repeatedly. Focus starts at the top of the document and moves in visual order; it never lands on the hidden "Add media" button, and the image's **Remove** button is still reachable by Tab and Shift+Tab.
- [x] ~~Confirm the Pantheon build log shows **both** `media_library_form_element` patches applied.~~ Done — see Reviewer notes.

References yalesites-org/YaleSites-Internal#1589

### DIFF
--- a/patches/contrib/media_library_form_element/core-3091297-port-focus-guard.patch
+++ b/patches/contrib/media_library_form_element/core-3091297-port-focus-guard.patch
@@ -1,0 +1,32 @@
+diff --git a/src/Element/MediaLibrary.php b/src/Element/MediaLibrary.php
+index d122a42..8aa169d 100644
+--- a/src/Element/MediaLibrary.php
++++ b/src/Element/MediaLibrary.php
+@@ -301,10 +301,24 @@ class MediaLibrary extends FormElementBase {
+     // items, the button needs to be disabled. Since we can't shift the focus to
+     // disabled elements, the focus is set back to the open button via
+     // JavaScript by adding the 'data-disabled-focus' attribute.
+-    // @see Drupal.behaviors.MediaLibraryWidgetDisableButton
++    // @see Drupal.behaviors.MediaLibraryFormElementDisableButton
+     if (!$cardinality_unlimited && $remaining === 0) {
+-      $element['media_library_open_button']['#attributes']['data-disabled-focus'] = 'true';
+-      $element['media_library_open_button']['#attributes']['class'][] = 'visually-hidden';
++      $triggering_element = $form_state->getTriggeringElement();
++      $trigger_parents = $triggering_element['#array_parents'] ?? [];
++      if ($trigger_parents && end($trigger_parents) === 'media_library_update_widget') {
++        // The element is being rebuilt from a selection change, so shift the
++        // focus back to the button the user left from before disabling it.
++        $element['media_library_open_button']['#attributes']['data-disabled-focus'] = 'true';
++        $element['media_library_open_button']['#attributes']['class'][] = 'visually-hidden';
++      }
++      else {
++        // The element is being built without a selection change, so it can be
++        // disabled right away and there is no need to set the focus first.
++        // Flagging it for focus here would make the browser scroll the
++        // visually-hidden button into view on a plain page load.
++        $element['media_library_open_button']['#disabled'] = TRUE;
++        $element['media_library_open_button']['#attributes']['class'][] = 'visually-hidden';
++      }
+     }
+ 
+     // This hidden field and button are used to add new items to the widget.
--- a/web/profiles/custom/yalesites_profile/composer.json
+++ b/web/profiles/custom/yalesites_profile/composer.json
@@ -165,7 +165,8 @@
         "Decode file URLs (link field) https://www.drupal.org/project/linkit/issues/3436733": "patches/contrib/linkit/3436733-5.patch"
       },
       "drupal/media_library_form_element": {
-        "Order Media items https://www.drupal.org/project/media_library_form_element/issues/3168027": "patches/contrib/media_library_form_element/3168027-8.patch"
+        "Order Media items https://www.drupal.org/project/media_library_form_element/issues/3168027": "patches/contrib/media_library_form_element/3168027-8.patch",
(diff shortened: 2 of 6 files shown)


miketullo95 (approved):
Approving — verified Site Settings, Header Settings, and Footer Settings all load at the top now, and the fallback teaser/logo fields still work (add, remove, save). The automated a11y checks in the PR description (activeElement/scrollY/tab order/keyboard trap) look sufficient to me for the screen reader AC, no need for a separate manual pass.

One heads up, not blocking: the linked issue notes a companion ticket (yalesites-org/YaleSites-Internal#1590) for the larger Site Settings reorg — worth retesting that one against this fix once it lands, per the issue's own note.

@dblanken-yale

PR #1238, opened 2026-03-31
#1140 :: Support Portrait Images in Resource Category View Block

laura-johnson:
### Description of work
- Adds image styles and responsive image style for resource portrait grid view mode
- Modifies the Content Resource widget to add new Portrait Grid view mode
- Adds placeholder image for Portrait Grid
- Updates ViewsBasicDynamicStyle plugin to add grid modifiers

### Functional testing steps:
- [ ] Add a Resource View to a page, select Portrait Grid. Verify it appears as expected, verify filters work normally.
- [ ] Verify the Portrait Grid and Portrait Grid Card appear in Storybook (see CL PR)

Demo: https://pr-1238-yalesites-platform.pantheonsite.io/resource-test

CL PR: https://github.com/yalesites-org/component-library-twig/pull/628
Atomic PR: https://github.com/yalesites-org/atomic/pull/449

PR #1500, opened 2026-08-22
1590: Site Settings: group settings into task-based sections instead of one flat list

dblanken-yale:
### Description of work

Restructures `/admin/yalesites/settings` from one flat run of fifteen unrelated settings into six task-based vertical tabs. **No capability change:** every setting still exists, still saves to the same config key, the route is still `/admin/yalesites/settings`, and the permission is still `yalesites manage settings`.

- Groups, in the order a user is likely to need them:
  - **Site basics** — Site name, Site email
  - **Key pages** — Homepage, Post landing page, Events calendar page, Access denied page (403), Page not found (404)
  - **Look and feel** — Font Pairing (+ its live preview), Custom Favicon, Fallback teaser image
  - **Search and analytics** — Google Site Verification, Google Analytics/Tag Manager
  - **Content and tagging** — Custom Vocabulary Name
  - **Advanced** — CAS Application Name, Show environment indicator
- Follows the existing house pattern in `ViewsSettingsForm`: one `#type: vertical_tabs` container plus `details` children carrying `#group`.
- **Grouped with `#group`, not by nesting.** `#group` relocates an element at render time and leaves `#parents` alone, so all sixteen values stay flat in the form state. `submitForm()` is therefore untouched, and so are the two `setValueForElement($form['site_page_403'|'site_page_404'], ...)` calls in `validateForm()` that reach for those elements by their top-level key — nesting would have silently repointed every config write and broken those two.
- **Label improvements** from the ticket: `Front page` → `Homepage`, `403 page` → `Access denied page (403)`, `404 page` → `Page not found (404)`. `Fallback teaser image` keeps its label (the word is used consistently platform-wide, including three node field descriptions in `config/sync` that point back at this setting) and gets a clearer description instead.
- **The Advanced group is `#access`-gated** on (user 1 OR `platform_admin`), so it never renders as an empty tab for a user who cannot see either field inside it.
- **Validation errors reveal their own group.** New `openGroupsWithErrors()` sets `#open` on the group holding any errored field. Core's `vertical-tabs.js` already opens the tab containing a `.error` field, but only once JS attaches — doing it server-side too means the reveal holds with JS off, avoids a flash of the wrong tab, and makes the behaviour assertable in a Kernel test rather than only in a browser.
- **Deep links work natively.** `core/misc/vertical-tabs.js` resolves `window.location.hash` to a tab, so documentation can point at `/admin/yalesites/settings#edit-key-pages` instead of "scroll down". Anchors: `#edit-site-basics`, `#edit-key-pages`, `#edit-look-and-feel`, `#edit-search-and-analytics`, `#edit-content-and-tagging`, `#edit-advanced`.
- `getFormId()` is deliberately unchanged (`ys_admin_settings`) — FormBuilder turns it into the `.ys-admin-settings` class that `ys_admin_theme/scss/gin-custom.scss` targets to constrain input widths. Pinned by a test so it cannot drift. No new CSS was needed.
- **Accessibility fixes that came out of reviewing this change** (both in the categories the ticket's a11y AC names):
  - The three path validation messages now name their field: "Page not found (404): the path 'not-a-path' has to start with a slash." Four Key pages fields share those three messages, and `inline_form_errors` is not enabled anywhere on this platform, so the messages region is the *only* error text on the page — with tabs, an unnamed message could be about any of four fields in a pane you are not looking at (3.3.1). Done by threading a `$label` argument into `validateStartWithSlash()`, `validateIsNotRootPath()` and `validatePath()`; every call site is in this class.
  - The three font-preview *sample headings* are now `aria-hidden`. They are `<h2>` purely for size, so they were the only headings in the Look and feel pane's outline and read as "1234567890" with nothing to convey (1.3.1). Deliberately scoped to the headings only, not the whole preview: the paragraphs next to them carry information the radio labels do not ("some digits (3, 4, 5, 7, 9) descend below the text baseline"), so those stay exposed. Retagging the `<h2>` to a `div` was the other option and was rejected — `font-preview.css` sets font-family, font-size and margin-bottom on `.preview-heading` but not font-weight or margin-top, so it would have changed the visuals.
  - `openGroupsWithErrors()` runs after `parent::validateForm()` so it also sees errors the parent adds.
- **First test coverage this form has ever had:** `SiteSettingsFormGroupingTest` (Kernel, 9 tests, 178 assertions) covers the tab structure and group order; that every setting lands in its intended group (a drop guard, plus an orphan guard for whoever appends the next setting); that settings are nested rather than `#group`-tagged; the form id; the relabels; Advanced hidden/visible per role; and that a validation error both opens its own group and names its field.

  The important one is `testSubmittingStillWritesEveryValueToItsOwnConfigKey()`, which submits the form for real and asserts all eleven config keys landed where they should. An earlier version of this test just asserted `#tree` was absent — which passed *vacuously*, since the code never mentions `#tree`, and would have kept passing under the regressions that actually matter (a child element declaring `#tree` itself, an explicit `#parents`, or a wrapper whose `getInfo()` sets it). This is the test that would catch a repointed config key, and nothing covered that before.

- Uses `phpcs:ignore` + `$strictConfigSchema = FALSE` for the test, matching the six sibling settings-form Kernel tests in this profile. `ys_core.site` still has no config schema; per this module's README that is deliberately **not** test-only work and is tracked in yalesites-org/YaleSites-Internal#579, so this PR does not pre-empt it.

### One thing worth knowing about how this was built

The first attempt tagged each setting with `#group` and left them at the top level. The Kernel suite went green — and the page was wrong: Font Pairing, Custom Favicon, the fallback teaser image and the Tag Manager link all rendered *outside* the tab widget. `#group` is only honoured by element types that wire `preRenderGroup` (textfield, textarea, checkbox, container, details, fieldset), so `radios`, `managed_file`, `media_library` and `item` silently ignored it. The tests had asserted intent (`#group` is set) rather than outcome (the element is actually inside the group). Caught by driving the real page, then fixed by nesting. The tests were rewritten to assert placement, and `testSettingsAreNestedRatherThanGroupTagged()` exists specifically so nobody re-introduces the tagging approach.

### Decisions that need a human, please confirm on review

1. **The open decision on group 5 (Content and tagging) was made without the team, because this was implemented autonomously.** The ticket says to get a decision rather than choose silently, and offers (a) fold Custom Vocabulary Name into Site basics and ship five groups, or (b) keep Content and tagging as a real group on the grounds that sitewide content settings are coming. I took the ticket's own recommendation — **(b)**, since yalesites-org/YaleSites-Internal#1061 (sitewide default event meta layout setting) is open and is exactly the named future occupant. If the team prefers (a), it is a one-line change. **A tab with one field in it does look thin, so this is the thing to push back on if you're going to push back on anything.**
2. **`Custom Vocabulary Name` was NOT renamed**, per the ticket's instruction to flag rather than rename unilaterally — "vocabulary" is jargon, but the label appears in help documentation and training.

### Acceptance criteria this PR does NOT cover

These are on the ticket and genuinely need a person; none of them are code:

- [ ] The two-person untrained-user findability test ("find the page shown when a link is broken" / "the image used when a post has no picture" in under two groups), including noting what they tried first.
- [ ] Accessibility engineer's WCAG 2.1 AA sign-off on the tab pattern. This PR uses core's standard vertical tabs unmodified rather than a custom widget, so the keyboard/focus/ARIA behaviour is core's — but that is an argument for why it should pass, not evidence that it did.
- [ ] Screen reader confirmation that group names are announced.
- [ ] `yalesites.yale.edu` Site Settings help page: new screenshots and instructions rewritten to reference groups instead of scroll position, plus the other help pages that say "scroll down to". The relabels need to land in docs in the same release.
- [ ] Release notes callout — users who know this page well will notice it moved.
- [ ] Coordination check with yalesites-org/YaleSites-Internal#1238 and yalesites-org/YaleSites-Internal#1368, both of which change the font controls on this exact form.
- [ ] yalesites-org/yalesites-project#1499 (issue yalesites-org/YaleSites-Internal#1589, the scroll-jump bug on this same page) is open and in review. It does **not** touch `SiteSettingsForm` — it patches a contrib media-library focus guard — so there is no file conflict, but whichever merges second should have the other retested, as that ticket's AC asks.

### Verification already done locally

Driven with Playwright against the local Lando site, not just asserted in tests:

- Every field checked for **actual DOM containment** inside its own `<details>` — 11 of 11 testable fields `ok`, the fallback-teaser fieldset and the font-preview container included. Nothing outside the tabs except core's own `visually-hidden` active-tab input.
- Font preview live-updates on radio change (`yalenew` -> `yalenew-oldstyle`).
- Deep links verified on fresh page loads: `#edit-search-and-analytics`, `#edit-content-and-tagging`, `#edit-key-pages` each open the right tab.
- Bad 404 path -> error message shown, **Key pages** becomes the active tab, errored field visible.
- **`system.site` and `ys_core.site` are byte-identical after two full form saves** (diff clean) — nothing dropped or repointed.
- Keyboard: Tab walks all six tabs in order, Enter activates the focused one, focus ring visible (confirmed on a screenshot).
- 390px viewport: no horizontal overflow; tabs degrade to six stacked collapsed `details`, which is core's behaviour.
- No console errors. `composer code-sniff` exit 0 (both standards, whole custom tree); `composer lint:php` clean. Full `ys_core` Kernel suite green against a captured pre-change baseline of 52 tests / 580 assertions.

### Functional testing steps:

- [ ] Go to `/admin/yalesites/settings` as a site admin. The page shows vertical tabs — Site basics, Key pages, Look and feel, Search and analytics, Content and tagging — instead of one long list, and **no** Advanced tab.
- [ ] Every setting that was on the page before is still present, in the group listed above. Nothing is stranded above the tabs.
- [ ] Open **Look and feel** and change Font Pairing. The live preview underneath updates immediately.
- [ ] Change something in two different tabs, save, and confirm both saved correctly and the page still saves everything else unchanged.
- [ ] In **Key pages**, put `not-a-path` (no leading slash) in Page not found (404) and save. The error appears with the Key pages tab open and the invalid field visible — not a message with no field in sight.
- [ ] Visit `/admin/yalesites/settings#edit-look-and-feel` directly. The Look and feel tab is the one open on arrival.
- [ ] As user 1, the **Advanced** tab appears and contains CAS Application Name and Show environment indicator. As a `platform_admin` who is not user 1, Advanced appears with Show environment indicator only.
- [ ] Save as a non-user-1 admin, then check `ys_core.site:cas_app_name` is unchanged (not blanked).
- [ ] Keyboard only: Tab through the six tab links and press Enter to switch panes (core's vertical tabs are a list of links, not an ARIA tablist, so there is no arrow-key navigation — worth an opinion from the accessibility engineer). Confirm focus stays visible, and note that core moves focus into the newly revealed pane's first field on Enter.
- [ ] Narrow the viewport to phone width and confirm the tabs collapse to stacked sections sensibly rather than breaking.

References yalesites-org/YaleSites-Internal#1590

### DIFF
--- a/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Form/SiteSettingsForm.php
+++ b/web/profiles/custom/yalesites_profile/modules/custom/ys_core/src/Form/SiteSettingsForm.php
@@ -162,14 +162,37 @@ class SiteSettingsForm extends ConfigFormBase implements ContainerInjectionInter
     $siteConfig = $this->config('system.site');
     $yaleConfig = $this->config('ys_core.site');
 
-    $form['site_name'] = [
+    // Group the settings by the task a user came here to do, rather than
+    // leaving them in one flat run. Each setting is nested inside its details
+    // group, the same way ViewsSettingsForm already does it: '#tree' is never
+    // set, so nesting rearranges only the render tree and every value still
+    // arrives flat in the form state. submitForm() therefore keeps reading
+    // 'site_name' and friends exactly as before.
+    //
+    // Nesting rather than tagging each setting with '#group' is deliberate.
+    // '#group' is only honoured by element types that wire preRenderGroup —
+    // textfield, textarea, checkbox, container, details, fieldset — so radios
+    // (font pairing), managed_file (favicon), media_library (teaser fallback)
+    // and item (the Tag Manager link) would silently render outside the tabs.
+    $form['vertical_tabs'] = [
+      '#type' => 'vertical_tabs',
+    ];
+
+    $form['site_basics'] = [
+      '#type' => 'details',
+      '#title' => $this->t('Site basics'),
+      '#description' => $this->t('What this site is called and who automated email comes from.'),
+      '#group' => 'vertical_tabs',
+    ];
+
+    $form['site_basics']['site_name'] = [
       '#type' => 'textfield',
       '#title' => $this->t('Site name'),
       '#default_value' => $siteConfig->get('name'),
       '#required' => TRUE,
     ];
 
-    $form['site_mail'] = [
+    $form['site_basics']['site_mail'] = [
       '#type' => 'textfield',
       '#description' => $this->t("The From address in automated emails sent during registration and new password requests, and other notifications. (Use an address ending in your site's domain to help prevent this email being flagged as spam.)"),
       '#title' => $this->t('Site email'),
@@ -177,16 +200,23 @@ class SiteSettingsForm extends ConfigFormBase implements ContainerInjectionInter
       '#required' => TRUE,
     ];
 
-    $form['site_page_front'] = [
+    $form['key_pages'] = [
+      '#type' => 'details',
+      '#title' => $this->t('Key pages'),
+      '#description' => $this->t('Which page the site should use for each of these purposes.'),
+      '#group' => 'vertical_tabs',
+    ];
+
+    $form['key_pages']['site_page_front'] = [
       '#type' => 'entity_autocomplete',
-      '#title' => $this->t('Front page'),
-      '#description' => $this->t("Specify a relative URL to display as the front page. Typically this points to a page in Drupal and is referenced by a node id. Use this autocomplete field to select the correct node."),
+      '#title' => $this->t('Homepage'),
+      '#description' => $this->t("Start typing to find the page you want visitors to land on first, then pick it from the list."),
       '#default_value' => $this->pathToNode($siteConfig->get('page')['front']),
       '#required' => TRUE,
       '#target_type' => 'node',
     ];
 
-    $form['site_page_posts'] = [
+    $form['key_pages']['site_page_posts'] = [
       '#type' => 'textfield',
       '#description' => $this->t("Specify a relative URL to display as the post landing page. This can be set to an existing page URL or use the default value '/post'."),
       '#title' => $this->t('Post landing page'),
@@ -194,7 +224,7 @@ class SiteSettingsForm extends ConfigFormBase implements ContainerInjectionInter
       '#required' => FALSE,
     ];
 
-    $form['site_page_events'] = [
+    $form['key_pages']['site_page_events'] = [
       '#type' => 'textfield',
       '#description' => $this->t("Specify a relative URL to display as the events calendar page. This can be set to an existing page URL or use the default value '/events'."),
       '#title' => $this->t('Events calendar page'),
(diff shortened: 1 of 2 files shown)


miketullo95 (approved):
Approving. Went through this against #1590's acceptance criteria and the two open decisions flagged in the PR description:

- **Content and tagging tab:** keeping option (b) as built — a real tab now, anticipating #1061 landing there. Thin today, but that's fine given what's coming.
- **Custom Vocabulary Name:** agreed not to rename unilaterally, as flagged. No action needed.
- **Advanced tab (CAS Application Name / Show environment indicator):** this matches what #1590 itself scoped, so no change needed here. I raised the "should these actually live on the dedicated Platform Admin Settings page" question during review — turns out #1560 already owns that as a broader audit ticket, so I left context there (yalesites-org/YaleSites-Internal#1560#issuecomment-5396306818) rather than duplicating scope on this PR.
- **Keyboard nav on vertical tabs** (Tab+Enter only, no arrow-key nav since core's tabs aren't an ARIA tablist): not blocking, this is core's platform-wide behavior. Flagging for the accessibility engineer's opinion as a follow-up rather than holding this PR on it.
- **Everything else the PR description lists as not covered** — accessibility engineer's WCAG sign-off, screen reader confirmation of group announcements, the two-person untrained-user findability test, docs/screenshot updates, release notes callout, coordination checks with #1238/#1368 — approving with these tracked as follow-up/parallel work rather than merge blockers. The code itself is solid and well tested.

@dblanken-yale

PR #1501, opened 2026-08-22
1591: Add a platform-admin-only top-level toolbar section and move Platform Admin Settings and Reports under it

dblanken-yale:
### Description of work

Platform admins had a growing pile of screens only they could see, scattered with no home of
their own: Platform Admin Settings sat inside the general Settings section next to the pages
site admins use every day, Reports was its own top-level toolbar item nobody else could reach,
and the orphaned inline blocks report had nowhere clean to go.

This adds one top-level, platform-admin-only **Platform Admin** section to the admin toolbar and
gathers those screens under it.

- New route `ys_core.admin_platform_admin` at `/admin/platform-admin`, using core's
  `SystemController::systemAdminMenuBlockPage` — the same pattern as `ys_core.admin_yalesites`,
  so the landing page lists its own children. Deliberately not a form route, which cannot.
- New top-level menu link (`parent: system.admin`), in `ys_core` with the rest of the top-level
  YaleSites admin IA.
- `ys_core.platform_admin_settings` reparents under it and its menu title becomes just
  **Settings** — the parent already names the audience. Route path, form class and page title
  are unchanged, so bookmarks and existing links keep working.
- `/admin/reports` reparents under it and is no longer a top-level toolbar item.
- Swept for other platform-admin-only screens and found exactly one: the orphaned inline blocks
  report (`ys_layouts`), which moves too. Everything else under `/admin/yalesites` is reachable
  by `site_admin`, so nothing else qualified — the sweep stayed bounded, and nothing ambiguous
  was absorbed from #1560's audit.
- Toolbar icon: a server/rack glyph, added the way `ys_toolbar` already does it (Gin's `--icon`
  custom property), so it inherits Gin's sizing and works in both horizontal and vertical
  toolbar states. Distinct from the gear already on Settings.

### The permission decision — needs your sign-off

The ticket asked for a real answer here rather than a default, so this is the part to review
first. **The landing route declares no permission of its own** — only core's
`_access_admin_menu_block_page` check, which grants access when at least one child is reachable.

Route requirements are ANDed, and Drupal prunes a menu subtree whose parent fails access. So a
permission on the parent would hide a child from any role holding only that child's permission:
a role with just `access site reports` would lose Reports from the menu while `/admin/reports`
kept working. That is exactly the coupling trap called out in the #1488 comment. The menu-block
check inverts it — the section can never hide a screen its user is allowed to see — and it needs
no edit as children come and go. The section stays platform-admin-only because every child is.
Reasoning is recorded on the route and in the `ys_core` README.

Verified, not assumed: `administer platform admin settings`, `access site reports` and
`administer orphaned inline blocks` are each held by `platform_admin` alone; no role has
`is_admin: true` and none holds `administer permissions`, so no role can grant itself in. No leak
via Reports' ~15 grandchildren either — the access check tests a child's own route before
recursing, so `access site reports` is still required.

### Testing done

- `PlatformAdminSectionTest` (Kernel): menu shape, per-role access on the landing route
  (including the pruning guard — an account with only `access site reports` still reaches the
  section), that the exported `core.menu.static_menu_link_overrides` agrees with the reparenting
  hook, and that the parent ID `ys_layouts` hangs its link off is actually defined by `ys_core`.
- Verified as a real `platform_admin` user, not user 1: `/admin/platform-admin` returns 200 with
  all three children listed in the **page body**, not only in a collapsed toolbar subtree.
  `site_admin` and `editor` are denied.
- Direct URLs still work regardless of menu placement: `/admin/reports`,
  `/admin/reports/dblog`, `/admin/yalesites/platform-admin-settings`, `/admin/yalesites`.
- ys_core kernel suite and ys_layouts kernel suite green; `composer code-sniff` clean.

### Notes for the reviewer

- **`/admin/reports/status` returns 403 for `platform_admin`, and did before this change** —
  it needs `administer site configuration`, which no role holds. Menu placement does not affect
  route access. Called out because the ticket asks to check Reports' children individually.
- **Children keep their `/admin/yalesites/*` paths.** Drupal builds admin breadcrumbs from the
  path, not the menu, so those pages still breadcrumb through YaleSites. Moving the paths would
  be a URL break and `ys_beacon`'s docs reference them, so the tradeoff is documented in the
  README instead. Say the word if you'd rather have the URL move plus redirects.
- **No browser screenshot.** `playwright-cli` is not installed in this environment, so the icon
  was verified by markup, the rendered CSS aggregate and the asset serving 200 — collapsed /
  expanded / active states were not visually confirmed. Worth a glance.
- **Icon overlap, minor:** Gin uses its own `server` sprite icon for `toolbar-icon-environment`
  (environment_indicator, which is enabled). No environment item renders locally, but on a
  Pantheon env the glyphs could look similar. Easy swap if you'd prefer, e.g. a shield.
- **Out of scope, worth a ticket:** the sibling `ys_core.admin_yalesites` route ANDs
  `_permission: 'yalesites manage settings'` with the same menu-block check, so it has the
  pruning bug this section avoids — a user with `administer redirects` but not
  `yalesites manage settings` loses Redirects from the menu while the URL works. Two sections,
  two access models. Not changed here to keep this diff to the issue.
- The Reports reparent is declared in both `ys_core_menu_links_discovered_alter()` and the
  exported static override. The config wins on a built site; the hook is kept so ys_core states
  the intent in code and survives a menu-link reset. A test asserts they agree, so they cannot
  drift silently.

### Functional testing steps:

- [ ] Log in as a real `platform_admin` user (not user 1 — user 1 bypasses permission checks and will make a broken menu look fine).
- [ ] Confirm a new **Platform Admin** item appears in the admin toolbar, with a server/rack icon that is clearly not the gear on **Settings**.
- [ ] Confirm **Reports** is no longer its own top-level toolbar item.
- [ ] Open `/admin/platform-admin` and confirm the page body lists **Settings**, **Reports** and **Orphaned inline blocks** — you should not have to expand a collapsed toolbar subtree to reach them.
- [ ] Confirm the icon looks right in all toolbar states: horizontal, vertical/collapsed, and when the section is the active trail.
- [ ] Confirm **Settings** under Platform Admin opens the platform-admin settings form, and that `/admin/yalesites/platform-admin-settings` still works directly (the path did not change).
- [ ] Confirm `/admin/reports` and its children (e.g. Recent log messages) still work by direct URL.
- [ ] Confirm `/admin/yalesites` no longer lists Platform Admin Settings or Orphaned inline blocks, and still lists everything else it did before.
- [ ] Log in as a `site_admin` and confirm the Platform Admin item does not appear at all, and that `/admin/platform-admin` returns 403.
- [ ] As `site_admin`, confirm the `/admin/yalesites` Settings section is unchanged apart from the two removals above.

References yalesites-org/YaleSites-Internal#1591


miketullo95 (approved):
Approving. On the two things you flagged for review:

- **Permission decision** (no route-level permission, just core's `_access_admin_menu_block_page` menu-block check instead of requiring `administer platform admin settings` directly): approach and reasoning check out. The subtree-pruning coupling you called out is exactly the right thing to design around, and the verification that no role can leak in via this holds up.
- **Icon overlap** with Gin's built-in environment-indicator icon: let's swap to the shield option you mentioned. Doesn't need to block merge — go ahead and land that as a quick follow-up commit on this branch rather than a separate review cycle.

Everything else — menu shape, the Reports reparent, test coverage, the direct-URL verification — looks solid.

@dblanken-yale

dblanken-yale:
## Icon overlap addressed

Landed on this branch as a follow-up commit rather than a separate review cycle, per your note.

- **Icon overlap with Gin's environment-indicator glyph** (@miketullo95) — swapped the Platform
  Admin toolbar icon from Feather `server` to Feather `shield`. `bd2ce217`

Confirmed the concern first rather than taking it on faith: `gin/styles/helpers/_toolbar-icons.scss`
does map `.toolbar-icon-environment::before` to `icon('server')`, and `environment_indicator` is
enabled. I then enumerated all 33 icon names in Gin's toolbar set — none is a shield, so the new
glyph collides with nothing rather than trading one overlap for another.

**Files:** `ys_toolbar/images/shield.svg` added, `--icon` repointed in `ys_toolbar/css/ys_toolbar.theme.css`,
`images/server.svg` deleted (unreferenced). The CSS comment now records *why* it must not be a
server glyph, so it does not get helpfully swapped back later.

**Also fixed the gap that let this through in the first place.** The original PR shipped without a
screenshot, and nothing in Drupal fails loudly when one of these `--icon` overrides points at a
missing file — the toolbar item just renders a blank square. Added
`ys_toolbar/tests/src/Unit/ToolbarIconAssetsTest.php`, which resolves every icon `url()` in that
stylesheet the way a browser would (including the root-relative form the custom property requires,
since a relative `url()` inside it resolves against Gin's stylesheet and would 404), and fails on an
icon file left behind unreferenced after a swap.

**Verification**
- Browser, as a real `platform_admin` user (not user 1): shield renders in the active trail, in the
  non-active state, at a narrow viewport and in dark mode. Zero failed asset requests in every run.
- `shield.svg` serves `200 image/svg+xml`; the removed `server.svg` now `404`s.
- `ys_toolbar` suite: 38 tests, 57 assertions, OK (includes the 4 new ones).
- `composer code-sniff`: exit 0.
- At the narrowest widths Gin collapses the toolbar to a hamburger, so no icon shows — standard Gin
  behaviour, not changed here.

**One thing worth your eye on the multidev:** `environment_indicator` renders only as a banner
locally, never as a toolbar item, so the actual side-by-side of shield vs. environment glyph is the
one check I could not make in Lando. Description has a testing step for it.

Labels left as they are — this was not a `needs work` PR and you said it needn't block merge, so no
new review has been requested.

